PR #4, opened 2020-11-05
API alignment with zenoh-c and code cleanup

Mallets:


kydos (commented):
I think we should use key as opposed to uri. In any case this is a matter that should be uniformed across all of our examples thus not necessarily an issue here.

PR #12, opened 2021-02-07
Add initial support for ESP32

cguimaraes:
- [x]  Fixed compilation errors to use zenoh-pico as a library in ESP32 projects
- [x]  Tested publish operations from a ESP32 NodeMCU (AZ-Delivery development board)
- [x]  Included ESP32 zn_pub example

### DIFF
--- a/include/zenoh-pico/net/private/system.h
+++ b/include/zenoh-pico/net/private/system.h
@@ -15,7 +15,7 @@
 #ifndef _ZENOH_NET_PICO_SYSTEM_H
 #define _ZENOH_NET_PICO_SYSTEM_H
 
-#if (ZENOH_LINUX == 1) || (ZENOH_MACOS == 1)
+#if (ZENOH_LINUX == 1) || (ZENOH_MACOS == 1) || (ESP32 == 1) || (ARDUINO_ARCH_ESP32 == 1)
 #include "zenoh-pico/net/private/system/unix.h"
 #elif (ZENOH_CONTIKI == 1)
 #include "zenoh-pico/net/private/contiki/types.h"
@@ -23,7 +23,11 @@
 
 /*------------------ Network ------------------*/
 // @TODO: remove the platform-specific include and data types
+#if (ESP32 == 1) || (ARDUINO_ARCH_ESP32 == 1)
+#include <lwip/sockets.h>
+#else
 #include <netinet/in.h>
+#endif
 #include "zenoh-pico/net/private/result.h"
 
 _ZN_RESULT_DECLARE(_zn_socket_t, socket)
--- a/include/zenoh-pico/net/types.h
+++ b/include/zenoh-pico/net/types.h
@@ -21,7 +21,7 @@
 #include "zenoh-pico/private/collection.h"
 #include "zenoh-pico/private/iobuf.h"
 
-#if (ZENOH_LINUX == 1) || (ZENOH_MACOS == 1)
+#if (ZENOH_LINUX == 1) || (ZENOH_MACOS == 1) || (ESP32 == 1) || (ARDUINO_ARCH_ESP32 == 1)
 #include "zenoh-pico/net/private/system/unix.h"
 #elif (ZENOH_CONTIKI == 1)
 #include "zenoh-pico/net/private/contiki/types.h"
--- a/include/zenoh-pico/private/system.h
+++ b/include/zenoh-pico/private/system.h
@@ -15,7 +15,7 @@
 #ifndef _ZENOH_PICO_SYSTEM_H
 #define _ZENOH_PICO_SYSTEM_H
 
-#if (ZENOH_LINUX == 1) || (ZENOH_MACOS == 1)
+#if (ZENOH_LINUX == 1) || (ZENOH_MACOS == 1) || (ESP32 == 1) || (ARDUINO_ARCH_ESP32 == 1)
 #include "zenoh-pico/private/system/unix.h"
 #elif (ZENOH_CONTIKI == 1)
 #include "zenoh-pico/private/contiki/types.h"
--- a/src/net/unix/network.c
+++ b/src/net/unix/network.c
@@ -14,7 +14,11 @@
 
 #include <arpa/inet.h>
 #include <errno.h>
+#if (ESP32 == 1) || (ARDUINO_ARCH_ESP32 == 1)
+#include <lwip/sockets.h>
+#else
 #include <ifaddrs.h>
+#endif
 #include <netdb.h>
 #include <net/if.h>
 #include <stdint.h>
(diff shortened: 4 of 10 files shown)


Mallets:
Hi @cguimaraes ,

Thank you for this contribution and your interest in Eclipse zenoh-pico!
However, we can't yet merge your pull request because the legal terms of the Eclipse Foundation requires that each contributor signs the Eclipse Contributor Agreement (ECA). And as you can see in the automatic Github checks here, the "eclipsefdn/eca" one failed because it can't find your signature.

To comply, you just need to create an Eclipse account if you haven't already (use the same email than your Github account) and to sign the document on this page: [https://accounts.eclipse.org/legal/eca/validation/58559](https://accounts.eclipse.org/legal/eca/validation/58559).
Then you can re-start the checks of this PR.

cguimaraes:
Hi @Mallets ,
I had to push -f (not a good thing to do) in order to make the commit compliant with the ECA and I handled the ECA as well.
I am assuming that it takes some time for the information to be updated on the link you provided.

cguimaraes:
Tested zn_pull and zn_sub against the zn_pub using two different ESP32.

include/zenoh-pico/net/private/system.h
```diff
@@ -15,15 +15,19 @@
 #ifndef _ZENOH_NET_PICO_SYSTEM_H
 #define _ZENOH_NET_PICO_SYSTEM_H
 
-#if (ZENOH_LINUX == 1) || (ZENOH_MACOS == 1)
+#if (ZENOH_LINUX == 1) || (ZENOH_MACOS == 1) || (ESP32 == 1) || (ARDUINO_ARCH_ESP32 == 1)
```
Mallets, line 18: It would be preferable to use the same naming convention:
ESP32 -> ZENOH_ESP32
ARDUINO_ARCH_ESP32 -> ZENOH_ARDUINO_ESP32

include/zenoh-pico/net/private/system.h
```diff
@@ -15,15 +15,19 @@
...
 #include "zenoh-pico/net/private/system/unix.h"
 #elif (ZENOH_CONTIKI == 1)
 #include "zenoh-pico/net/private/contiki/types.h"
 #endif
 
 /*------------------ Network ------------------*/
 // @TODO: remove the platform-specific include and data types
+#if (ESP32 == 1) || (ARDUINO_ARCH_ESP32 == 1)
```
Mallets, line 26: It would be preferable to use the same naming convention:
ESP32 -> ZENOH_ESP32
ARDUINO_ARCH_ESP32 -> ZENOH_ARDUINO_ESP32

include/zenoh-pico/net/types.h
```diff
@@ -21,7 +21,7 @@
 #include "zenoh-pico/private/collection.h"
 #include "zenoh-pico/private/iobuf.h"
 
-#if (ZENOH_LINUX == 1) || (ZENOH_MACOS == 1)
+#if (ZENOH_LINUX == 1) || (ZENOH_MACOS == 1) || (ESP32 == 1) || (ARDUINO_ARCH_ESP32 == 1)
```
Mallets, line 24: It would be preferable to use the same naming convention:
ESP32 -> ZENOH_ESP32
ARDUINO_ARCH_ESP32 -> ZENOH_ARDUINO_ESP32

include/zenoh-pico/private/system.h
```diff
@@ -15,7 +15,7 @@
 #ifndef _ZENOH_PICO_SYSTEM_H
 #define _ZENOH_PICO_SYSTEM_H
 
-#if (ZENOH_LINUX == 1) || (ZENOH_MACOS == 1)
+#if (ZENOH_LINUX == 1) || (ZENOH_MACOS == 1) || (ESP32 == 1) || (ARDUINO_ARCH_ESP32 == 1)
```
Mallets, line 18: It would be preferable to use the same naming convention:
ESP32 -> ZENOH_ESP32
ARDUINO_ARCH_ESP32 -> ZENOH_ARDUINO_ESP32

src/net/unix/network.c
```diff
@@ -27,6 +31,8 @@
 /*------------------ Interfaces and sockets ------------------*/
 char *_zn_select_scout_iface()
 {
+    char *result;
```
Mallets, line 34: Pointer is not initialised but should be initialized to NULL.

cguimaraes: True. Thanks for pointing this out.

include/zenoh-pico/net/private/system.h
cguimaraes: Although this might be possible, I think it is better to keep it like this.
   - to use your naming conventions, this needs to be done either in the CMakefile.txt (this is a still pending point) or to add them **manually** in whatever tool for embedded development you use (I am using platformio for example).
   - in my experience, I have already seen several occurrences of people failing to correctly follow the installation instructions, especially if they require manually/additional configurations.

Let me know if you still want me to change them.

Mallets: Based on your reply, I believe we might consider to leave the naming as you proposed. However, it would be great that zenoh-pico could be compiled for ESP32 platform also via the shipped CMakefile, do you think is it feasible? What kind of changes would be required?

cguimaraes: I was taking a look in the file struct of zenoh-pico, and maybe we can solve this by having a mapping in the [zenoh-pico.h](zenoh-pico.h). Something like:
```
#if (ESP32 == 1) || (ARDUINO_ARCH_ESP32 == 1)
  #define ZENOH_ESP32
#endif
```
This would allow to use your naming conventions in the rest of the code and to avoid issues when compiling with the CMake or when using tools for embedded development like platformio.

Regarding CMake, it is something that I would need to look into, but for sure it is feasible to compile for ESP32 platform via the shipped CMakefile.

Mallets: I like the proposal. I believe the zenoh-pico.h is a good place where to put those kinds of definitions. Please let us know if you need any help with the CMakefile, unfortunately I don't have an ESP32 board at hand to help you with.

src/net/unix/network.c
```diff
@@ -79,8 +85,9 @@ char *_zn_select_scout_iface()
             current = current->ifa_next;
         } while ((iface == 0) && (current != 0));
     }
-    char *result = strdup((iface != 0) ? iface : loopback);
-    freeifaddrs(ifap);
```
Mallets, line 83: Is there any particular reason on removing 'freeifaddrs(ifap)'? 
This could lead to a memory leak where the memory allocated for 'ifap' is never freed.

cguimaraes: Fixed.
I forgot to put it back after debugging the zenoh-pico library for ESP32.
In any case, I still plan to go through that function to see how to support the scouting in ESP32.

src/net/unix/network.c
cguimaraes: Fixed.

cguimaraes:
Closed, since pull request #42 was submitted with a fully integrated supported for ESP32.

PR #24, opened 2021-03-31
Add support for zephyr

esteve:
@Mallets this PR has many changes, let me know if you'd like me to break it out:

- I changed the `#if`s to just check if the macros had been defined.
- I added an implementation of `strdup`, since Zephyr does not provide it
- Because some MCUs do not implement `_times()` when calling `clock()`, I replaced the `srand()` initialization to just use `time()`
- Added support for Zephyr. Zephyr does not support many of the socket flags that are used in zenoh-pico, so this last commit is the one I'm not sure it should be merged as is, let me know if the flags are really needed.


src/net/zephyr/network.c
```diff
@@ -0,0 +1,389 @@
...
+
+    if (r.value.socket < 0)
+    {
+        r.tag = _z_res_t_ERR;
+        r.value.error = r.value.socket;
+        r.value.socket = 0;
+        return r;
+    }
+
+    // NOTE(esteve): SO_KEEPALIVE and SO_LINGER not supported in Zephyr
```
Mallets, line 205: I would simply remove the commented code and leave the note saying that those flags are not supported in Zephyr.

src/net/zephyr/network.c
```diff
@@ -0,0 +1,389 @@
...
+                return -1;
+            }
+            n -= wb;
+            bs.val += bs.len - n;
+        } while (n > 0);
+    }
+
+    return 0;
+}
+
```
Mallets, line 338: I would simply remove the commented code and leave the note saying that iovec is not supported in Zephyr.

Mallets:
@esteve I went through your PR and it seems fine to me. Thanks for adding the support for Zephyr into zenoh-pico!

For what concerns the socket flags, they are not mandatory in zenoh-pico. It's only a matter of tuning that we found beneficial on UNIX systems. If those flags are not supported in Zephyr, let's simply ignore them.

PR #59, opened 2021-12-23
Merge from parallel development in atolab

cguimaraes:
This merge includes:
 - Refactor of collections
 - Refactor of the entity/primitive management
 - Support for multicast
 - Support for multicast transport
 - Peer mode capabilities without routing and only on multicast transport
    -- No resource declaration supported yet


cguimaraes:
- Additional tests were successfully run in Arduino (ESP32) and Zephyr (Nucleo-767ZI) frameworks. Namely, running the different examples in unicast or multicast transport.
- Check for memory leaks.

PR #66, opened 2022-02-28
Align keep alive and lease expiration with the Rust implementation

cguimaraes:


include/zenoh-pico/config.h
```diff
@@ -115,11 +115,10 @@
  * Default session lease in milliseconds: 10 seconds
  */
 #define ZN_TRANSPORT_LEASE 10000
-#define ZN_KEEP_ALIVE_INTERVAL 1000
-#define ZN_LEASE_EXPIRE_FACTOR 3.5
+#define ZN_TRANSPORT_LEASE_EXPIRE_FACTOR 3.5
 
 /**
- * Default multicast session join interval in milliseconds: 10 seconds
+ * Default multicast session join interval in milliseconds: 1 seconds
```
Mallets, line 121: This should be 2.5 seconds, isn't it?

cguimaraes: 1 second matches the default configured value in ZN_JOIN_INTERVAL.

cguimaraes: changed to 2.5s by default

PR #91, opened 2022-06-16
Fix #90 . Make use of random number generators available on each system

### LINKED ISSUE #90: [Bug] Improve random number generator
### Describe the bug

The current random number generator in Zenoh-Pico might lead to issues while connecting to a Zenoh router.

In the microcontrollers, as the time always start from 0, it is very likely that the seed well be the same between multiple boots / reset. As a consequence, the PEER_ID might also be the same, meaning that Zenoh router might close new connections if the maximum number of links is reached.

In Unix systems, running two clients exactly at the same second might run into similar problems.

### To reproduce

N/A

### System info

Verified in ESP32 and Unix systems.

cguimaraes:


tests/zn_msgcodec_test.c
```diff
@@ -313,7 +317,7 @@ void assert_eq_payload(_zn_payload_t *left, _zn_payload_t *right)
 void payload_field(void)
 {
     printf("\n>> Payload field\n");
-    _z_wbuf_t wbf = gen_wbuf(128);
+    _z_wbuf_t wbf = gen_wbuf(65535);
```
Mallets, line 320: What's the interest of changing `128` to `65535` in the codec test? There were some edge cases triggered by that value? 
Finally, if `65535` is to get the max batch size, shouldn't we use a const for that and make it explicit?

include/zenoh-pico/system/platform.h
```diff
@@ -30,6 +30,9 @@
 #error "Unknown platform"
 #endif
 
+/*------------------ Random ------------------*/
+uint8_t z_random(void);
```
Mallets, line 34: In standard C/C++ `rand()` is defined as `int rand(void)`. 
For the sake of compatibility, I believe it would be to expose the same function signature.
What do you think?

tests/zn_msgcodec_test.c
cguimaraes: `_z_wbuf_t wbf = gen_wbuf(65535);` represents the buffer that will have the serialized message. As such, it should be able to keep the maximum size of the message.

With the new random number generator, serialized message size were surpassing the size of buffer, thus failing the serialization.

Still, applying the suggestion on the const to make it explicit.

Mallets: It sounds good. Thanks for the clarification.

include/zenoh-pico/system/platform.h
cguimaraes: I would agree with you at first, but then an `int` does not have the same size for all the platforms. Especially for small embedded devices on which an `int` can be 16-bits instead of 32-bits.

Because of that, and in order to have the same expected output across platforms, I opt for making it return a random value between 0-255. Bigger random values can still be generated by using something like this:
```
z_zint_t gen_zint(void)
{
    z_zint_t ret = 0;
    for (unsigned int i = 0; i < sizeof(z_zint_t); i++)
        ret |= z_random() << (i * 8);

    return ret;
}
```

Counter-proposal: To extend the z_random methods to:
```
int z_random(void);
uint8_t z_random_u8(void);
uint16_t z_random_u16(void);
uint32_t z_random_u32(void);
uint64_t z_random_u64(void);
z_zint_t z_random_zint(void);
```

what do you think?

Mallets: I believe it's a better proposal since it allows to get the best out of each platform. 
E.g.: ESP32 could make use of `esp_random` and `esp_fill_random` (see [here](https://docs.espressif.com/projects/esp-idf/en/latest/esp32s2/api-reference/system/random.html)) to generate PRN instead of zenoh-pico building everything from a `uint8_t`.

cguimaraes: Addressed in the commit below.

cguimaraes:
Request for comments for the moment.
Testing in all supported platforms still needs to be done before the merge.

PR #96, opened 2022-07-13
Reduce heap size to avoid SRAM overflow

jmrua:
Reduce the heap size in the example configuration for the nRF52840 to solve a linking error resulting from an overflow in the SRAM section.


Mallets:
Bump @cguimaraes to review.

PR #105, opened 2022-08-19
Introduce encoding for query_reply and publisher

Mallets:


include/zenoh-pico/api/primitives.h
```diff
@@ -317,6 +317,14 @@ const char *zp_config_get(z_config_t *config, unsigned int key);
...
 
+/**
+ * Constructs a known :c:type:`z_encoding_t`.
+ *
+ * Returns:
+ *   Returns the constructed :c:type:`z_encoding_t`.
+ */
+z_encoding_t z_encoding(z_encoding_prefix_t prefix);
+
 /**
```
cguimaraes, line 328: If we are providing the helpers for the encodings, then I would provide one to set the suffix as well.
To be considered if we also need the getters for the `prefix` and `suffix` in  `z_encoding_t`.

And in this case, both `prefix` and `suffix` in  `z_encoding_t` can be privatized.

src/api/api.c
```diff
@@ -122,10 +122,12 @@ int8_t zp_config_insert(z_config_t *config, unsigned int key, z_string_t value)
     return _zp_config_insert(config, key, value);
 }
 
-z_encoding_t z_encoding_default(void) {
-    return (_z_encoding_t){.prefix = Z_ENCODING_PREFIX_EMPTY, .suffix = _z_bytes_make(0)};
+z_encoding_t z_encoding(z_encoding_prefix_t prefix) {
+    return (_z_encoding_t){.prefix = prefix, .suffix = _z_bytes_make(0)};
 }
```
cguimaraes, line 127: To be consistent with other data structures at the API-level, I would still keep the `z_encoding_t z_encoding_default(void)`.

cguimaraes (changes requested):
Small changes suggested.

src/api/api.c
p-avital: As long as `z_encoding_default` keeps existing (it does, l129, I almost missed it), I'm good with this

Mallets: Indeed, `z_econding_default` has not been remove: just reworked.

include/zenoh-pico/api/primitives.h
Mallets: I've extended `z_encoding` to accept both `prefix` and `suffix` in the constructor. I believe there is no need to add getters for `z_encoding_t`.

cguimaraes:
Regarding the last commit, completeness (considering it is the N parameter on the complete) should be an integer and not a bool. Unless, it has been discussed to be changed while I was off.

Mallets:
It is just misname, `complete` is a bool. `completeness` is still a `z_int_t`.

cguimaraes (approved):

PR #107, opened 2022-08-30
Fix #106 . Assign IP address of interface to imr_interface member of ip_mreq struct

### LINKED ISSUE #106: Assign IP address of interface to imr_interface member of ip_mreq struct
### Describe the feature

Hi,

following test setup does not work:
- zn_peer_pub.c running on microcontroller connected to en1 interface of Linux workstation
- zn_peer_sub.c running on Linux workstation

The problem is that the subscriber listens on en0 interface for multicast messages.
This is due to the INADDR_ANY assignment to mreq.imr_interface.s_addr in zenoh-pico/src/system/unix/network.c.
"imr_address is the address of the local interface with which the system should join the multicast group; if it is equal to        INADDR_ANY, an appropriate interface is chosen by the system." - https://man7.org/linux/man-pages/man7/ip.7.html
"If this member specifies an IPv4 address of 0.0.0.0, the default IPv4 multicast interface is used." - https://docs.microsoft.com/en-us/windows/win32/api/ws2ipdef/ns-ws2ipdef-ip_mreq

To be able to handle multiple network interfaces, it would be better to pass the address of the respective interface to _zn_listen_udp_multicast() and _zn_close_udp_multicast().

Thanks!

cguimaraes:


src/system/unix/network.c
```diff
@@ -377,10 +377,33 @@ void *_z_listen_udp_multicast(void *arg, uint32_t tout, const char *iface) {
...
+                    else {
+                        continue;
+                    }
+
+                    break;
+                }
+            }
+        }
+        freeifaddrs(l_ifaddr);
+
```
Mallets, line 402: I suggest to move the newly-added block of code to a dedicated function and not part of the logic.

Mallets (changes requested):

src/system/unix/network.c
cguimaraes: Converted into an utility function.

PR #110, opened 2022-09-12
Change default UDP multicast scouting port to 7446

Mallets:


tests/multicast.sh
```diff
@@ -47,7 +47,7 @@ for INTERFACE in $(echo $INTERFACES | xargs); do
     fi
 done
 
-LOCATORS="udp/[ff10::1234]:7447#iface=$INTERFACE"
+LOCATORS="udp/[ff10::1234]:7446#iface=$INTERFACE"
```
cguimaraes, line 50: This one should not be changed.
These are the tests for the multicast transport, so it should run on 7447 as defined for the unicast transport.

tests/multicast.sh
```diff
@@ -28,7 +28,7 @@ for INTERFACE in $(echo $INTERFACES | xargs); do
     fi
 done
 
-LOCATORS="udp/224.0.0.225:7447#iface=$INTERFACE"
+LOCATORS="udp/224.0.0.225:7446#iface=$INTERFACE"
```
cguimaraes, line 31: This one should not be changed.
These are the tests for the multicast transport, so it should run on 7447 as defined for the unicast transport.

cguimaraes (approved):

PR #112, opened 2022-09-14
Allow reply and hello in callbacks to be taken by the user

cguimaraes:
It includes moving the z_hello_array_t into a z_hello_list_t.

### DIFF
--- a/examples/c11/z_scout.c
+++ b/examples/c11/z_scout.c
@@ -64,8 +64,8 @@ void fprinthello(FILE *stream, const z_hello_t *hello) {
     fprintf(stream, " }");
 }
 
-void callback(z_owned_hello_t hello, void *context) {
-    fprinthello(stdout, hello._value);
+void callback(z_owned_hello_t *hello, void *context) {
+    fprinthello(stdout, hello->_value);
     fprintf(stdout, "\n");
     (*(int *)context)++;
 }
--- a/examples/c11/z_scout.c
+++ b/examples/c11/z_scout.c
@@ -65,7 +65,7 @@ void fprinthello(FILE *stream, const z_hello_t *hello) {
 }
 
 void callback(z_owned_hello_t *hello, void *context) {
-    fprinthello(stdout, hello->_value);
+    fprinthello(stdout, z_loan(*hello));
     fprintf(stdout, "\n");
     (*(int *)context)++;
 }
--- a/examples/c99/z_scout.c
+++ b/examples/c99/z_scout.c
@@ -62,8 +62,8 @@ void fprinthello(FILE *stream, const z_hello_t *hello) {
     fprintf(stream, " }");
 }
 
-void callback(z_owned_hello_t hello, void *context) {
-    fprinthello(stdout, hello._value);
+void callback(z_owned_hello_t *hello, void *context) {
+    fprinthello(stdout, hello->_value);
     fprintf(stdout, "\n");
     (*(int *)context)++;
 }
--- a/examples/c99/z_scout.c
+++ b/examples/c99/z_scout.c
@@ -63,7 +63,7 @@ void fprinthello(FILE *stream, const z_hello_t *hello) {
 }
 
 void callback(z_owned_hello_t *hello, void *context) {
-    fprinthello(stdout, hello->_value);
+    fprinthello(stdout, z_hello_loan(hello));
     fprintf(stdout, "\n");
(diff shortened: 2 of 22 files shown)


src/api/api.c
```diff
@@ -238,7 +242,7 @@ _MUTABLE_OWNED_FUNCTIONS_DEFINITION(z_bytes_t, z_owned_bytes_t, bytes, _z_bytes_
 _MUTABLE_OWNED_FUNCTIONS_DEFINITION(z_string_t, z_owned_string_t, string, _z_string_free, _z_owner_noop_copy)
 _IMMUTABLE_OWNED_FUNCTIONS_DEFINITION(z_keyexpr_t, z_owned_keyexpr_t, keyexpr, _z_keyexpr_free, _z_keyexpr_copy)
 _MUTABLE_OWNED_FUNCTIONS_DEFINITION(z_config_t, z_owned_config_t, config, _z_config_free, _z_owner_noop_copy)
-_MUTABLE_OWNED_FUNCTIONS_DEFINITION(z_scouting_config_t, z_owned_scouting_config_t, scouting_config, _z_owner_noop_free, _z_owner_noop_copy)
+_MUTABLE_OWNED_FUNCTIONS_DEFINITION(z_scouting_config_t, z_owned_scouting_config_t, scouting_config, _z_config_free, _z_owner_noop_copy)
```
Mallets, line 245: Shouldn't be `_z_scouting_config_free`?

Mallets (changes requested):

src/api/api.c
cguimaraes: Done.

src/session/query.c
```diff
@@ -213,19 +219,12 @@ int _z_trigger_query_reply_final(_z_session_t *zn, const _z_reply_context_t *rep
...
-        _z_pending_reply_list_t *pen_rps = pen_qry->_pending_replies;
-        _z_pending_reply_t *pen_rep = NULL;
-        while (pen_rps != NULL) {
-            pen_rep = _z_pending_reply_list_head(pen_rps);
+        while (pen_qry->_pending_replies != NULL) {
+            _z_pending_reply_t *pen_rep = _z_pending_reply_list_head(pen_qry->_pending_replies);
 
-            // Check if this is the same resource key
             // Trigger the query handler
-            if (_z_keyexpr_intersect(pen_qry->_key._suffix, strlen(pen_qry->_key._suffix),
```
Mallets, line 223: I believe this check shouldn't be removed to guarantee that all the replies are for the query being issued.

Mallets (changes requested):

examples/c11/z_scout.c
```diff
@@ -64,8 +64,8 @@ void fprinthello(FILE *stream, const z_hello_t *hello) {
     fprintf(stream, " }");
 }
 
-void callback(z_owned_hello_t hello, void *context) {
-    fprinthello(stdout, hello._value);
+void callback(z_owned_hello_t *hello, void *context) {
+    fprinthello(stdout, hello->_value);
```
p-avital, line 68: not necessary for this PR, but we need a loan function here (zenoh-c currently doesn't have the equivalent loan-type)

examples/c99/z_scout.c
```diff
@@ -62,8 +62,8 @@ void fprinthello(FILE *stream, const z_hello_t *hello) {
     fprintf(stream, " }");
 }
 
-void callback(z_owned_hello_t hello, void *context) {
-    fprinthello(stdout, hello._value);
+void callback(z_owned_hello_t *hello, void *context) {
+    fprinthello(stdout, hello->_value);
```
p-avital, line 66: same as c11 scouting example

src/session/query.c
p-avital: As discussed internally, agreed

cguimaraes: I agree. We must ensure that the user do not get a reply outside the requested set of keyexpr.

p-avital (approved):

examples/c99/z_scout.c
cguimaraes: In Pico we have it, so I am just using the z_loan now.

examples/c11/z_scout.c
cguimaraes: In Pico we have it, so I am just using the z_loan now.

Mallets (approved):

PR #130, opened 2022-11-16
Add support for query payload

cguimaraes:
As defined in https://github.com/eclipse-zenoh/roadmap/blob/3df8f147ae7c9586d452bc38e83497d52d71eade/rfcs/ALL/Query%20Payload.md

To be merged after:
- [x] https://github.com/eclipse-zenoh/zenoh/pull/389 is merged into master.
- [x] Point reference from zenoh query-body to zenoh master branch.

### DIFF
--- a/examples/unix/c11/z_queryable.c
+++ b/examples/unix/c11/z_queryable.c
@@ -26,6 +26,9 @@ void query_handler(const z_query_t *query, void *ctx) {
     char *keystr = z_keyexpr_to_string(z_query_keyexpr(query));
     z_bytes_t pred = z_query_parameters(query);
     printf(" >> [Queryable handler] Received Query '%s?%.*s'\n", keystr, (int)pred.len, pred.start);
+    if (query->_body.payload.len > 0) {
+        printf("     with value '%.*s'\n", (int)query->_body.payload.len, query->_body.payload.start);
+    }
     z_query_reply_options_t options = z_query_reply_options_default();
     options.encoding = z_encoding(Z_ENCODING_PREFIX_TEXT_PLAIN, NULL);
     z_query_reply(query, z_keyexpr(keyexpr), (const unsigned char *)value, strlen(value), &options);
--- a/examples/unix/c11/z_queryable.c
+++ b/examples/unix/c11/z_queryable.c
@@ -26,8 +26,8 @@ void query_handler(const z_query_t *query, void *ctx) {
     char *keystr = z_keyexpr_to_string(z_query_keyexpr(query));
     z_bytes_t pred = z_query_parameters(query);
     printf(" >> [Queryable handler] Received Query '%s?%.*s'\n", keystr, (int)pred.len, pred.start);
-    if (query->_body.payload.len > 0) {
-        printf("     with value '%.*s'\n", (int)query->_body.payload.len, query->_body.payload.start);
+    if (query->_with_value.payload.len > 0) {
+        printf("     with value '%.*s'\n", (int)query->_with_value.payload.len, query->_with_value.payload.start);
     }
     z_query_reply_options_t options = z_query_reply_options_default();
     options.encoding = z_encoding(Z_ENCODING_PREFIX_TEXT_PLAIN, NULL);
--- a/examples/unix/c11/z_queryable.c
+++ b/examples/unix/c11/z_queryable.c
@@ -25,7 +25,7 @@ void query_handler(const z_query_t *query, void *ctx) {
     (void)(ctx);
     char *keystr = z_keyexpr_to_string(z_query_keyexpr(query));
     z_bytes_t pred = z_query_parameters(query);
-    z_value_t payload_value = z_query_payload_value(query);
+    z_value_t payload_value = z_query_value(query);
     printf(" >> [Queryable handler] Received Query '%s?%.*s'\n", keystr, (int)pred.len, pred.start);
     if (payload_value.payload.len > 0) {
         printf("     with value '%.*s'\n", (int)payload_value.payload.len, payload_value.payload.start);
--- a/examples/unix/c11/z_queryable.c
+++ b/examples/unix/c11/z_queryable.c
@@ -25,9 +25,10 @@ void query_handler(const z_query_t *query, void *ctx) {
     (void)(ctx);
     char *keystr = z_keyexpr_to_string(z_query_keyexpr(query));
     z_bytes_t pred = z_query_parameters(query);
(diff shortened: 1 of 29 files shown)


src/api/api.c
```diff
@@ -227,6 +231,8 @@ z_bytes_t z_query_parameters(const z_query_t *query) {
     return parameters;
 }
 
+z_value_t z_query_payload_value(const z_query_t *query) { return query->_with_value; }
```
Mallets, line 234: What about just calling it `z_query_payload`? 
Or `z_query_body`?

src/api/api.c
```diff
@@ -477,7 +483,9 @@ int8_t z_delete(z_session_t zs, z_keyexpr_t keyexpr, const z_delete_options_t *o
 }
 
 z_get_options_t z_get_options_default(void) {
-    return (z_get_options_t){.target = z_query_target_default(), .consolidation = z_query_consolidation_default()};
+    return (z_get_options_t){.target = z_query_target_default(),
+                             .consolidation = z_query_consolidation_default(),
+                             .with_value = {.encoding = z_encoding_default(), .payload = _z_bytes_make(0)}};
```
Mallets, line 488: Should we use the term `body` instead of `payload`?

examples/unix/c11/z_queryable.c
```diff
@@ -25,7 +25,11 @@ void query_handler(const z_query_t *query, void *ctx) {
     (void)(ctx);
     char *keystr = z_keyexpr_to_string(z_query_keyexpr(query));
     z_bytes_t pred = z_query_parameters(query);
+    z_value_t payload_value = z_query_payload_value(query);
     printf(" >> [Queryable handler] Received Query '%s?%.*s'\n", keystr, (int)pred.len, pred.start);
+    if (payload_value.payload.len > 0) {
```
Mallets, line 30: Should we have some sort of `z_check` in place for `z_value_t`?

src/protocol/msgcodec.c
```diff
@@ -1019,6 +1024,19 @@ void _z_query_decode_na(_z_zbuf_t *zbf, uint8_t header, _z_query_result_t *r) {
     _z_query_consolidation_result_t r_con = _z_query_consolidation_decode(zbf);
     _ASSURE_P_RESULT(r_con, r, _Z_ERR_PARSE_CONSOLIDATION)
     r->_value._consolidation = r_con._value;
+
+    if (_Z_HAS_FLAG(header, _Z_FLAG_Z_B) == true) {
+        _z_data_info_result_t r_dti = _z_data_info_decode(zbf);
+        _ASSURE_P_RESULT(r_dti, r, _Z_ERR_PARSE_ZENOH_MESSAGE)
```
Mallets, line 1030: Is there any `_Z_ERR_PARSE_DATA_INFO`?

src/protocol/msgcodec.c
```diff
@@ -1019,6 +1024,19 @@ void _z_query_decode_na(_z_zbuf_t *zbf, uint8_t header, _z_query_result_t *r) {
...
+        r->_value._info = r_dti._value;
+
+        _z_payload_result_t r_pld = _z_payload_decode(zbf);
+        _ASSURE_P_RESULT(r_pld, r, _Z_ERR_PARSE_PAYLOAD)
+        r->_value._payload = r_pld._value;
+    } else {
+        (void)memset(&r->_value._info, 0, sizeof(_z_data_info_t));
+        r->_value._payload = _z_bytes_make(0);
```
Mallets, line 1038: As a suggestion, we might have a function `_z_bytes_empty()` to properly initialize empty bytes instead of using `_z_bytes_make(0)`.

tests/z_msgcodec_test.c
```diff
@@ -182,6 +182,19 @@ _z_locator_array_t gen_locator_array(size_t size) {
     return la;
 }
 
+_z_value_t gen_with_value(void) {
```
Mallets, line 185: Shouldn't be called `gen_value`?

tests/z_msgcodec_test.c
```diff
@@ -1339,7 +1352,16 @@ _z_zenoh_message_t gen_query_message(void) {
     z_consolidation_mode_t consolidation;
     consolidation = con[gen_uint8() % (sizeof(con) / sizeof(uint8_t))];
 
-    return _z_msg_make_query(key, parameters, qid, target, consolidation);
+    _z_value_t with_value;
+    if (gen_bool()) {
+        with_value = gen_with_value();
+    } else {
+        with_value.encoding.prefix = Z_ENCODING_PREFIX_EMPTY;
```
Mallets, line 1359: Should have a constructor for the empty value?

include/zenoh-pico/api/primitives.h
```diff
@@ -454,6 +454,17 @@ z_query_consolidation_t z_query_consolidation_none(void);
...
+/**
+ * Get a query's payload value by aliasing it.
+ *
+ * Parameters:
+ *   query: Pointer to the query to get the value selector from.
+ *
+ * Returns:
+ *   Returns the payload value wrapped as a :c:type:`z_value_t`, since payload value is a user-defined representation.
+ */
+z_value_t z_query_payload_value(const z_query_t *query);
```
Mallets, line 466: I tend to prefer the term `z_query_body`. TBD.

src/api/api.c
cguimaraes: `z_query_value` was agreed.

examples/unix/c11/z_queryable.c
cguimaraes: @p-avital @cguimaraes keep the examples checking the length of the value, but also support `z_check` for those that want to be semantically correct.

include/zenoh-pico/api/primitives.h
cguimaraes: z_query_value was agreed.

tests/z_msgcodec_test.c
cguimaraes: It would be more correct indeed.

tests/z_msgcodec_test.c
cguimaraes: As of now, I think it would just be useful for the tests.
It might be implemented in the future, if we see the need for it.

src/protocol/msgcodec.c
cguimaraes: Added in the upcoming commit.

PR #132, opened 2022-12-08
Adopt idiomatic C-style for error returns

cguimaraes:
This PR refactors core part of Zenoh-Pico to adopt a more C-style idiomatic for error returns.
Such approach allowed to gain a bit of performance in terms of speed but also memory usage.
It also provides a safer way to decode, copy, and clear messages.


include/zenoh-pico/protocol/msgcodec.h
```diff
@@ -23,190 +23,163 @@
...
 
-// /*------------------ Message Decorators ------------------*/
-_Z_DECLARE_ENCODE_NOH(attachment);
-_Z_DECLARE_P_DECODE(attachment);
-
-_Z_DECLARE_ENCODE_NOH(reply_context);
-_Z_DECLARE_P_DECODE(reply_context);
+int8_t _z_data_info_encode(_z_wbuf_t *buf, const _z_data_info_t *di);
+int8_t _z_data_info_decode(_z_data_info_t *di, _z_zbuf_t *buf);
+int8_t _z_data_info_decode_v2(_z_data_info_t *di, _z_zbuf_t *buf);
```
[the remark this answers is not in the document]
cguimaraes: Artifacts to be removed from the temporary performance tests.

include/zenoh-pico/protocol/msgcodec.h
```diff
@@ -23,190 +23,163 @@
...
+int8_t _z_data_info_decode_na(_z_data_info_t *di, _z_zbuf_t *buf);
+int8_t _z_data_info_decode_na_v2(_z_data_info_t *di, _z_zbuf_t *buf);
```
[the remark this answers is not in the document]
cguimaraes: Artifacts to be removed from the temporary performance tests.

include/zenoh-pico/utils/config.h
```diff
@@ -58,7 +58,7 @@ int8_t _zp_config_insert(_z_config_t *ps, unsigned int key, _z_string_t value);
  * Returns:
  *     The value of the property with key ``key`` in properties map ``ps``.
  */
-char *_z_config_get(const _z_config_t *ps, unsigned int key);
+char *_z_config_get(const _z_config_t *ps, uint8_t key);
```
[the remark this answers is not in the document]
cguimaraes: Up to now we don't have that many configuration.
The day we have more than that we can bump it to uint16_t without any impact for the user.

src/protocol/msgcodec.c
```diff
@@ -307,108 +295,91 @@ int8_t _z_res_decl_encode(_z_wbuf_t *wbf, uint8_t header, const _z_res_decl_t *d
...
 /*------------------ Subscriber Declaration ------------------*/
 int8_t _z_sub_decl_encode(_z_wbuf_t *wbf, uint8_t header, const _z_sub_decl_t *dcl) {
     int8_t ret = _Z_RES_OK;
     _Z_DEBUG("Encoding _Z_DECL_SUBSCRIBER\n");
 
-    // Encode the body
     _Z_EC(_z_keyexpr_encode(wbf, header, &dcl->_key))
     if (_Z_HAS_FLAG(header, _Z_FLAG_Z_S) == true) {
-        ret = _z_subinfo_encode(wbf, &dcl->_subinfo);
+        ret |= _z_subinfo_encode(wbf, &dcl->_subinfo);
```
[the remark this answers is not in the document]
cguimaraes: If an error occurs on the decoding, it must always overlap the success (that happen before or after). Also, it allow us to detect if any system error also occurred, which should be treated differently from other Zenoh errors. 

Before we were taking the suggested approach, but together with early returns in case of errors. With MISRA, we need to go all the way to the single return in the end of the function. I found this approach to be cleaner in terms of code, but also got a better performance than with nested `IFs`.

src/system/arduino/esp32/network.cpp
```diff
@@ -526,75 +531,72 @@ size_t _z_read_exact_udp_multicast(_z_sys_net_socket_t sock, uint8_t *ptr, size_
...
-            sock._bts->begin(gname, false);
+            sock->_bts->begin(gname, false);
         } else if (mode == _Z_BT_MODE_MASTER) {
-            sock._bts->begin(gname, true);
-            uint8_t connected = sock._bts->connect(gname);
+            sock->_bts->begin(gname, true);
+            uint8_t connected = sock->_bts->connect(gname);
             if (!connected) {
-                while (!sock._bts->connected(10000)) {
+                while (!sock->_bts->connected(10000)) {
```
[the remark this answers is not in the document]
cguimaraes: This is a timeout.
I should make it coming from the locator config.
Good catch.

include/zenoh-pico/collections/pointer.h
```diff
@@ -30,9 +30,15 @@
     static inline name##_sptr_t name##_sptr_new(type##_t val) {                                 \
         name##_sptr_t p;                                                                        \
         p.ptr = (type##_t *)z_malloc(sizeof(type##_t));                                         \
-        *p.ptr = val;                                                                           \
         p._cnt = (atomic_uint *)z_malloc(sizeof(atomic_uint));                                  \
```
Mallets, line 33: This malloc should be inside the `if (p.ptr != NULL)`. If for any obscure reason the first malloc fails and the second one succeeds, we have a memory leak because the corresponding cleanup is not done.

include/zenoh-pico/collections/pointer.h
```diff
@@ -73,9 +86,15 @@
     static inline name##_sptr_t name##_sptr_new(type##_t val) {                                 \
         name##_sptr_t p;                                                                        \
         p.ptr = (type##_t *)z_malloc(sizeof(type##_t));                                         \
-        *p.ptr = val;                                                                           \
         p._cnt = (uint8_t *)z_malloc(sizeof(uint8_t));                                          \
```
Mallets, line 89: This malloc should be inside the if (p.ptr != NULL). If for any obscure reason the first malloc fails and the second one succeeds, we have a memory leak because the corresponding cleanup is not done.

src/session/utils.c
```diff
@@ -44,9 +46,22 @@ void _z_timestamp_reset(_z_timestamp_t *tstamp) {
     tstamp->_time = 0;
 }
 
+int8_t _z_session_generate_pid(_z_bytes_t *bs, uint8_t size) {
```
Mallets, line 49: The correct name should be `zid` and not `pid`.

Mallets (changes requested):

src/session/utils.c
cguimaraes: Done in the upcoming commits.

include/zenoh-pico/collections/pointer.h
cguimaraes: Done in the upcoming commits.

include/zenoh-pico/collections/pointer.h
cguimaraes: Done in the upcoming commits.

Mallets (approved):
To be merged once the CI is happy on all the boards.

PR #135, opened 2022-12-26
Adding support for emscripten platform and WebSocket transport (#138)

gabrik:
This PR supports the [emscripten](https://github.com/emscripten-core/emscripten) platform in Zenoh-Pico.

Emscripten allows building C/C++ code to WASM. In the case of Zenoh-Pico this allows building a WASM/JS API on top of it.
It comes with implementation for WebSocket transport.


gabrik:
@cguimaraes seems the compiler in emscripten is not happy about the code:

```
/home/runner/work/zenoh-pico/zenoh-pico/src/net/config.c:20:29: error: a function declaration without a prototype is deprecated in all versions of C [-Werror,-Wstrict-prototypes]
_z_config_t *_z_config_empty() {
                            ^
                             void
/home/runner/work/zenoh-pico/zenoh-pico/src/net/config.c:28:31: error: a function declaration without a prototype is deprecated in all versions of C [-Werror,-Wstrict-prototypes]
_z_config_t *_z_config_default() { return _z_config_client(NULL); }
                              ^
                               void
2 errors generated.

```

While the others are happy about the same functions, any idea?

include/zenoh-pico/system/platform.h
```diff
@@ -31,6 +31,8 @@
 #include "zenoh-pico/system/platform/arduino/esp32.h"
 #elif defined(ZENOH_ARDUINO_OPENCR)
 #include "zenoh-pico/system/platform/arduino/opencr.h"
+#elif defined(ZENOH_WASM)
```
cguimaraes, line 34: What about changing `ZENOH_WASM` to `ZENOH_EMSCRIPTEN` since this port is tied to Emscripten as the compiler toolchain for WebAssembly?

include/zenoh-pico/system/platform/emscripten.h
```diff
@@ -0,0 +1,49 @@
...
+#if Z_LINK_WS == 1
+        int _fd;
+#endif
+    };
+} _z_sys_net_socket_t;
+
+typedef struct {
+    union {
+#if Z_LINK_WS == 1
+        struct addrinfo *_ipws;
```
cguimaraes, line 44: I would change it to `struct addrinfo *_iptcp;` to be consistent with the other ports. Also, it might be misleading in the sense that `struct addrinfo` has no specific information about the websocket other than tcp/ip information.

src/system/emscripten/network.c
```diff
@@ -0,0 +1,166 @@
...
+    // @TODO: To be implemented
+    ret = _Z_ERR_GENERIC;
+
+    return ret;
+}
+
+void _z_close_ws(_z_sys_net_socket_t *sock) { close(sock->_fd); }
+
+size_t _z_read_ws(const _z_sys_net_socket_t sock, uint8_t *ptr, size_t len) {
+    // WARNING: workaroud as the socket is non blocking. Try 500 times
```
cguimaraes, line 108: Instead of trying 500 times, it should try to read while the configured timeout does not expire. Since, `setsockopt` is not supported in emscripten, it should be manually done here.

src/system/emscripten/network.c
```diff
@@ -0,0 +1,166 @@
...
+
+        n = n + rb;
+        pos = _z_ptr_u8_offset(pos, n);
+    } while (n != len);
+
+    return n;
+}
+
+size_t _z_send_ws(const _z_sys_net_socket_t sock, const uint8_t *ptr, size_t len) {
+    // WARNING: workaroud as the socket is non blocking. Try 500 times
```
cguimaraes, line 139: Instead of trying 500 times, it should try to send while the configured timeout does not expire. Since, setsockopt is not supported in emscripten, it should be manually done here.

src/system/emscripten/network.c
```diff
@@ -0,0 +1,166 @@
...
+        c++;
+        z_sleep_ms(50);
+    } while (res <= 0 && c < 500);
+    return res;
+}
+
+#endif
+
+#if Z_LINK_TCP == 1
+#error "TCP not supported yet on emscripten port of Zenoh-Pico"
```
cguimaraes, line 153: Replace `emscripten` with `Emscripten` (everywhere).

src/system/emscripten/system.c
```diff
@@ -0,0 +1,115 @@
...
+int8_t _z_condvar_free(_z_condvar_t *cv) { return pthread_cond_destroy(cv); }
+
+int8_t _z_condvar_signal(_z_condvar_t *cv) { return pthread_cond_signal(cv); }
+
+int8_t _z_condvar_wait(_z_condvar_t *cv, _z_mutex_t *m) { return pthread_cond_wait(cv, m); }
+#endif  // Z_MULTI_THREAD == 1
+
+/*------------------ Sleep ------------------*/
+int z_sleep_us(unsigned int time) {
+    emscripten_sleep(1);
```
cguimaraes, line 83: 1) Set a comment mentioning that `emscripten_sleep` granualarity is `ms`. So, it sleeps the minimum possible if value is below `1ms`
2) Check if `time` is `>1ms` in that case it should sleep `at least` the configured time.

cguimaraes (changes requested):

src/system/emscripten/system.c
```diff
@@ -0,0 +1,119 @@
...
+
+int8_t _z_condvar_signal(_z_condvar_t *cv) { return pthread_cond_signal(cv); }
+
+int8_t _z_condvar_wait(_z_condvar_t *cv, _z_mutex_t *m) { return pthread_cond_wait(cv, m); }
+#endif  // Z_MULTI_THREAD == 1
+
+/*------------------ Sleep ------------------*/
+int z_sleep_us(unsigned int time) {
+    if (time > 1000) {
+        emscripten_sleep(time);
```
cguimaraes, line 84: Like this you will sleep the defined `time` in `ms` instead of `us`.
What about something like this to replace line 84:
`emscripten_sleep((time / 1000) + (time % 1000 == 0 ? 0 : 1));`
In this way, it will round up to the `ms` in such a way that it will sleep at least the configured `time`.

cguimaraes (changes requested):

src/system/emscripten/system.c
```diff
@@ -0,0 +1,121 @@
...
+    emscripten_sleep(time);
+    return 0;
+}
+
+int z_sleep_s(unsigned int time) { return z_sleep_ms(time * 1000); }
+
+/*------------------ Instant ------------------*/
+z_clock_t z_clock_now(void) { return z_time_now(); }
+
+unsigned long z_clock_elapsed_us(z_clock_t *instant) {
```
cguimaraes, line 97: Instead of not supporting, I am wondering if we could just rely on the `z_clock_elapsed_ms` and provide an approximation to the `us`. E.g.:
```
unsigned long z_clock_elapsed_us(z_clock_t *instant) { return z_clock_elapsed_ms(instant) * 1000; }
```

src/system/emscripten/system.c
```diff
@@ -0,0 +1,121 @@
...
+}
+
+unsigned long z_clock_elapsed_ms(z_clock_t *instant) { return z_time_elapsed_ms(instant); }
+
+unsigned long z_clock_elapsed_s(z_clock_t *instant) { return z_time_elapsed_ms(instant) * 1000; }
+
+/*------------------ Time ------------------*/
+z_time_t z_time_now(void) { return emscripten_get_now(); }
+
+unsigned long z_time_elapsed_us(z_time_t *time) {
```
cguimaraes, line 109: Instead of not supporting, I am wondering if we could just rely on the `z_time_elapsed_us` and provide an approximation to the `us`. E.g.:
```C
unsigned long z_time_elapsed_us(z_time_t *time) {
    return z_time_elapsed_ms(time) * 1000;
}
```

cguimaraes:
Integration tests are failing due to a change in Zenoh (Rust).
To be merged when fixed.

PR #150, opened 2023-01-22
User have to deallocate memory allocated by zenoh-pico (#145)

### LINKED ISSUE #145: [Bug] User have to deallocate memory allocated by zenoh-pico
### Describe the bug

The function `z_keyexpr_to_string` / `zp_keyexpr_resolve` requires user to drop data (`char *`) allocated by it.

This is definitely wrong for Windows platform (memory allocation should not cross DLL boundary) and in certain situations may be unsafe on Unix platform (if caller or callee is compiled with custom malloc/free for example). See https://stackoverflow.com/questions/64574662/allocating-and-deallocating-memory-across-shared-lib-boundaries

Related to https://github.com/eclipse-zenoh/zenoh-c/issues/91 .
Also https://github.com/eclipse-zenoh/zenoh-c/pull/134 .

### To reproduce

N/A

### System info

N/A

cguimaraes:
Fixes #145

### DIFF
--- a/src/collections/string.c
+++ b/src/collections/string.c
@@ -103,6 +103,8 @@ void _z_str_free(char **src) {
     }
 }
 
+void _z_str_copy(char *dst, const char *src) { strncpy(dst, src, strlen(src)); }
+
 char *_z_str_clone(const char *src) {
     size_t str_len = _z_str_size(src);
     char *dst = (char *)z_malloc(str_len);
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -98,9 +98,9 @@ To check if ``val`` is still valid, you may use ``z_X_check(&val)`` or ``z_check
 
   A zenoh-allocated :c:type:`z_bytes_t`.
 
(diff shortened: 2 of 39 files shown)


milyin:
Looks good for me. But I see one difference between zenoh-c and zenoh-pico api. There is no specific `z_XXX_move` functions in zenoh-c, there is only `z_move` macro which just takes reference to it's argument. So the code
```
z_drop(z_str_move(&str))
```
will not be compiled with zenoh-c. Though it's not a specific string-related issue

src/collections/string.c
```diff
@@ -103,6 +103,8 @@ void _z_str_free(char **src) {
     }
 }
 
+void _z_str_copy(char *dst, const char *src) { strncpy(dst, src, strlen(src)); }
```
p-avital, line 106: So this is just `strcpy`? Is this an attempt at bypassing MISRA's forbidding of it? :p

p-avital (approved):

Issue #156, opened 2023-01-27
[Bug] Zephyr pub/sub example not communicating


dcorbeil:
### Describe the bug

Hi all,

I am trying to get the pub/sub example to work on the Nucleo-F429zi. I already have a Zephyr Cmake project that I integrated zenoh-pico into and got it to build without major issues. The problem comes when I try to get my two F429zi to communicate through the pub and sub example. I use the `prj.conf` example from `docs/zephyr/nucleo_f767zi` since the two boards are pretty similar to each other. Although there are some configs that are turned off by zephyr due to dependencies not being met (might or might not be related to my issue). That applies to the `nucleo_f767zi` as well so I commented out `ETH_NATIVE_POSIX`, `ETH_NATIVE_POSIX_RANDOM_MAC` and `NET_SOCKETS_POSIX_NAMES`. I then set my two devices as peers by setting `#define CLIENT_OR_PEER 1`. I tried connecting my two boards either directly to each other or both of them in a network switch. In both these cases they can ping each other so I know that at least there is _some_ communication working. Then when I start the publisher and subscriber, both are declared subscriber and publisher correctly and the publisher is publishing its messages but the subscriber doesn't receive anything. 

This is my first time playing with zenoh so I might be missing some important detail. Are there tips and tricks to help debug what's happening? Or rather _not_ happening.

I also tried one F429zi in client mode connected directly on my host computer with a router running but had a different problem. The router was built from [zenoh](https://github.com/eclipse-zenoh/zenoh/commit/b06d91f5986fb87b607af92be874ecda28d021d9)'s latest commit the day of writing this. The client F429 wasn't able to create a zenoh session.

Here is my `prj.conf`

```
# Important for thread debugging info by doing:
#       info threads
# To switch threads:
#       thread <thread ID>
CONFIG_DEBUG_THREAD_INFO=y

# Kernel options
CONFIG_MAIN_STACK_SIZE=8192
CONFIG_HEAP_MEM_POOL_SIZE=8192
CONFIG_ENTROPY_GENERATOR=y
CONFIG_TEST_RANDOM_GENERATOR=y
CONFIG_INIT_STACKS=y

# Generic library options
CONFIG_NEWLIB_LIBC=y
CONFIG_NEWLIB_LIBC_NANO=n
CONFIG_POSIX_API=y

# Generic networking options
CONFIG_SERIAL=y
CONFIG_NETWORKING=y

CONFIG_NET_L2_ETHERNET=y
[...]

PR #158, opened 2023-02-01
Improve sleep resilience considering only their granularity (#156)

cguimaraes:


src/system/zephyr/system.c
```diff
@@ -103,7 +103,13 @@ int8_t _z_condvar_wait(_z_condvar_t *cv, _z_mutex_t *m) { return pthread_cond_wa
...
 int z_sleep_us(unsigned int time) { return usleep(time); }
 
-int z_sleep_ms(unsigned int time) { return usleep(time * 1000U); }
+int z_sleep_ms(unsigned int time) {
+    for (unsigned int i = 0; i < time; i++) {
+        z_sleep_us(1000);
+    }
+
+    return 0;
+}
```
dcorbeil, line 112: As discussed in #156, I had a [fix](https://github.com/dcorbeil/zenoh-pico/commit/b67b34013c531c88fcbc18ffe9f0a4dc80780570) for this but you were faster on the trigger than me. Feel free to use it if you wish.

My thinking was to keep using Zephyr's POSIX API as you were already doing for `z_sleep_us` and `z_sleep_s` and because the now deprecated [`usleep`](https://man7.org/linux/man-pages/man3/usleep.3.html) recommends using `nanosleep` instead, I went with that.

Although, under the hood they end up doing mostly the same thing.

cguimaraes: You can still do the PR. We really appreciate the engagement from the community and users.

I believe you can do the PR on top of this one, and you would appear as co-author when merged. I also departed from your feedback to fix an eventual issue on the other platforms.

dcorbeil: Ok I'll try that.

cguimaraes: Thanks. I have accepted it into this PR.
However, in order to be merged upstream into the Eclipse repositories you need to sign the ECA with the same email used in the commit message. Check more details here: https://accounts.eclipse.org/legal/eca/validation/138875

cguimaraes:
@dcorbeil in addition, I decided to have a deeper look on Zephyr native APIs, and I decided to make use of the Zephyr Sleep API instead of the POSIX compatible one.

If you can test it again, it would be great.

dcorbeil (approved):

src/system/arduino/esp32/system.c
```diff
@@ -117,7 +117,13 @@ int8_t _z_condvar_wait(_z_condvar_t *cv, _z_mutex_t *m) { return pthread_cond_wa
 /*------------------ Sleep ------------------*/
 int z_sleep_us(unsigned int time) { return usleep(time); }
 
-int z_sleep_ms(unsigned int time) { return usleep(time * 1000U); }
+int z_sleep_ms(unsigned int time) {
+    for (unsigned int i = 0; i < time; i++) {
+        z_sleep_us(1000);
```
p-avital, line 122: Most sleep APIs promise to sleep _at least_ whatever you asked them to. This may compound, so this approach may make sleeps longer than expected:
Arguably, we might just decide to accept this compounding if we measure it to be small enough for our application.

cguimaraes: Nice catch! You are right.
I will include an extra check on how much it slept and continue subsequent sleeps accordingly, so we can minimize the amount of extra time.
The important point is that we still sleep at least `time`.

p-avital: Yeah, current implementation will oversleep by x‰, where x is the average oversleep in µs of `z_sleep_us`. In micros, I would expect this number could be surprisingly high, so a deadline might be a good thing to add

cguimaraes: Indeed, this was accumulating more than I expected:
  - Sleeping 10000 ms OP (i.e., `for` loop): 13.071s 12.625s 12.623s
  - Sleeping 10000 ms now: 10.363s 10.006s 10.005s

p-avital (approved):

PR #159, opened 2023-02-02
Align log output with Zenoh Rust

cguimaraes:
It now includes the date, time, debug level, and component.

```bash
[2023-02-02T17:52:03Z DEBUG ::_z_scout_encode] Encoding _Z_MID_SCOUT
[2023-02-02T17:52:03Z DEBUG ::_z_hello_decode_na] Decoding _Z_MID_HELLO
[2023-02-02T17:52:03Z DEBUG ::_z_locators_decode_na] Decoding _LOCATORS
[2023-02-02T17:52:03Z INFO ::__z_scout_loop] Received _Z_HELLO message
[2023-02-02T17:52:03Z INFO ::_z_transport_unicast_open_client] Sending Z_INIT(Syn)
[2023-02-02T17:52:03Z DEBUG ::_z_init_encode] Encoding _Z_MID_INIT
[2023-02-02T17:52:03Z DEBUG ::_z_init_decode_na] Decoding _Z_MID_INIT
[2023-02-02T17:52:03Z INFO ::_z_transport_unicast_open_client] Received Z_INIT(Ack)
[2023-02-02T17:52:03Z INFO ::_z_transport_unicast_open_client] Sending Z_OPEN(Syn)
[2023-02-02T17:52:03Z DEBUG ::_z_open_encode] Encoding _Z_MID_OPEN
[2023-02-02T17:52:03Z DEBUG ::_z_open_decode_na] Decoding _Z_MID_OPEN
[2023-02-02T17:52:03Z INFO ::_z_transport_unicast_open_client] Received Z_OPEN(Ack)
```


include/zenoh-pico/system/platform.h
```diff
@@ -86,6 +86,7 @@ unsigned long z_clock_elapsed_s(z_clock_t *time);
 
 /*------------------ Time ------------------*/
 z_time_t z_time_now(void);
+char *z_time_now_as_str(char *buf, unsigned long buflen);
```
p-avital, line 89: Taking is a `const char*` (not a `char const*`) and returning `void` would be clearer as to ownership status of the passed and returned pointers

cguimaraes: I think you mean `const char *z_time_now_as_str(char *const buf, unsigned long buflen);`.

 - By returning the pointer, we can directly use the transformation in the printing functions. E.g.: `printf("[%s ", z_time_now_as_str(ret, sizeof(ret)));`
 - `z_time_now_as_str` is indeed changing the content of the `buf`, so it cannot be a `const char*` (which is the same as `char const *`). But making it a `char * const` (similar to reference) is what we actually want, meaning the value being pointed at can change but not the pointer itself.

I am pretty sure this is what you were thinking, so good catch again :-)

p-avital: The return thing is a nice trick, but it might result in slightly poorer code generation (because the fact that the return value is always the first parameter is hidden in another compilation unit, so we have to rely on LTO to spot that).

But as long as the ownership is clear, I'm fine with either :)

p-avital (approved):

PR #161, opened 2023-02-04
Broken non test build on newer Zephyr commits

dcorbeil:
On Zephyr commit [e83a13aa](https://github.com/zephyrproject-rtos/zephyr/commit/e83a13aabf7f477c408a20bc09885384f39e372f), Zephyr changed `CONFIG_TEST_EXTRA_STACKSIZE` to `CONFIG_TEST_EXTRA_STACK_SIZE`. This has the effect of breaking `zenoh-pico` builds when using newer zephyr commits because this config is not turned on by default.


cguimaraes (approved):
LGTM

Issue #163, opened 2023-02-06
Ease the integration of Zenoh-Pico into cpp projects


cguimaraes:
### Describe the feature

**Is your feature request related to a problem?**
Integrating Zenoh-Pico into a CPP project might be a frustrating experience in what concerns handle compilation and linking errors. Zenoh is made to be simple in all its embodiment.

**Describe the solution you'd like**
Include Zenoh-Pico header files as `#include <zenoh-pico.h>` no matter if it is a C or CPP project. Meaning, no need to wrap it around an `extern "C"` directive.

**Additional context**
Zenoh-C is implemented something similar as it is being used as the baseline for Zenoh-CPP

PR #164, opened 2023-02-06
Ease the integration of Zenoh-Pico into cpp projects (#163)

cguimaraes:
Related to #163 .
Working in my tests. @gabrik to test on his environment since it was one of the failing ones.

### DIFF
--- a/include/zenoh-pico/collections/pointer.h
+++ b/include/zenoh-pico/collections/pointer.h
@@ -21,9 +21,10 @@
 #if ZENOH_C_STANDARD != 99
 #ifndef __cplusplus
 #include <stdatomic.h>
+#define z_atomic(X) _Atomic X
 #else
 #include <atomic>
-#define _Atomic(X) std::atomic<X>
+#define z_atomic(X) std::atomic<X>
 using namespace std;
 #endif
 
@@ -31,13 +32,13 @@ using namespace std;
 #define _Z_POINTER_DEFINE(name, type)                                                           \
     typedef struct {                                                                            \
         type##_t *ptr;                                                                          \
-        _Atomic unsigned int *_cnt;                                                             \
+        z_atomic(unsigned int) * _cnt;                                                          \
     } name##_sptr_t;                                                                            \
     static inline name##_sptr_t name##_sptr_new(type##_t val) {                                 \
         name##_sptr_t p;                                                                        \
         p.ptr = (type##_t *)z_malloc(sizeof(type##_t));                                         \
         if (p.ptr != NULL) {                                                                    \
-            p._cnt = (_Atomic unsigned int *)z_malloc(sizeof(_Atomic unsigned int *));          \
+            p._cnt = (z_atomic(unsigned int) *)z_malloc(sizeof(z_atomic(unsigned int) *));      \
             if (p._cnt != NULL) {                                                               \
                 *p.ptr = val;                                                                   \
                 atomic_store_explicit(p._cnt, 1, memory_order_relaxed);                         \
--- a/include/zenoh-pico/collections/pointer.h
+++ b/include/zenoh-pico/collections/pointer.h
@@ -31,13 +31,13 @@ using namespace std;
 #define _Z_POINTER_DEFINE(name, type)                                                           \
     typedef struct {                                                                            \
         type##_t *ptr;                                                                          \
-        _Atomic(unsigned int) *_cnt;                                                            \
+        _Atomic unsigned int *_cnt;                                                             \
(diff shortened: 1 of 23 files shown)


include/zenoh-pico/collections/pointer.h
```diff
@@ -19,19 +19,25 @@
...
+#define _Atomic(X) std::atomic<X>
+using namespace std;
+#endif
 
 /*------------------ Internal Array Macros ------------------*/
 #define _Z_POINTER_DEFINE(name, type)                                                           \
     typedef struct {                                                                            \
         type##_t *ptr;                                                                          \
-        atomic_uint *_cnt;                                                                      \
+        _Atomic unsigned int *_cnt;                                                             \
```
p-avital, line 34: This looks odd to me: without the parentheses, is this interpreted as `_Atomic(unsigned int *)_cnt`, `_Atomic(unsigned int) *_cnt` or `_Atomic(unsigned) int *_cnt`?

p-avital (changes requested):

include/zenoh-pico/collections/pointer.h
cguimaraes: It is interpreted as `_Atomic(unsigned int) *_cnt`.

I would have preferred the more explicit and use `_Atomic(unsigned int) *_cnt` but Zephyr did not agree with me. I had to do this change in [cf6341c](https://github.com/eclipse-zenoh/zenoh-pico/pull/164/commits/cf6341ce084de4cc24edc63673977a80fd469e45) to make all the ports happy.

p-avital: I think
```c++
#ifndef __cplusplus
#include <stdatomic.h>
#define z_atomic(X) _Atomic X
#else
#include <atomic>
#define z_atomic(X) std::atomic<X>
using namespace std;
#endif
```
would solve the Zephyr issue, and make which part is atomic clearer.

include/zenoh-pico/collections/pointer.h
```diff
@@ -19,19 +19,25 @@
 #include <stdint.h>
 
 #if ZENOH_C_STANDARD != 99
+#ifndef __cplusplus
 #include <stdatomic.h>
+#else
+#include <atomic>
+#define _Atomic(X) std::atomic<X>
+using namespace std;
```
dcorbeil, line 27: That doesn't seem to be the right fix, at least for Zephyr because it doesn't have `std::atomic` and the whole of `std`.

Here is the compile error:

```
In file included from ../../lib/zenoh-pico/include/zenoh-pico/session/session.h:22,
                 from ../../lib/zenoh-pico/include/zenoh-pico/net/session.h:19,
                 from ../../lib/zenoh-pico/include/zenoh-pico/net/publish.h:18,
                 from ../../lib/zenoh-pico/include/zenoh-pico/api/types.h:17,
                 from ../../lib/zenoh-pico/include/zenoh-pico/api/primitives.h:21,
                 from ../../lib/zenoh-pico/include/zenoh-pico/api/macros.h:17,
                 from ../../lib/zenoh-pico/include/zenoh-pico.h:28,
                 from ../../src/main.cpp:24:
../../lib/zenoh-pico/include/zenoh-pico/collections/pointer.h:25:10: fatal error: atomic: No such file or directory
   25 | #include <atomic>
      |          ^~~~~~~~
compilation terminated.
```

cguimaraes: I am assuming you set the following configs in your `prf.conf`
```
CONFIG_CPP=y
CONFIG_CPP_MAIN=y
```

if that is the case you might want to enable C++ standard library by setting [CONFIG_LIB_CPLUSPLUS](https://docs.zephyrproject.org/3.1.0/kconfig.html#CONFIG_LIB_CPLUSPLUS) in the application configuration file.

Although this solves that particular include error, it dump many other that I am now tackling it.

include/zenoh-pico/collections/pointer.h
cguimaraes: This is done.

include/zenoh-pico/collections/pointer.h
cguimaraes: And the latest commit should fix the remaining compilation issues.
If you want to give it a try again @dcorbeil

dcorbeil: Oh no my bad. Turning on `CONFIG_LIB_CPLUSPLUS` fixes it. No problems to report. Thanks

dcorbeil (approved):

p-avital (approved):

PR #172, opened 2023-03-14
canonizer fix for one-character chunks

milyin:
Fix for canonization of expressions with one-character chunks. For example expression `a/**/**/b` was canonized incorrectly because single "b" was incorrectly treated as "*"
Also fixed bug in test: assert actually didn't work in release configuration. To make them work in release the '#undef NDEBUG' must be added before including "assert.h"


cguimaraes (approved):
LGTM.
@p-avital if you can give a 2nd look since you implemented the keyexpr helpers.

p-avital:
Thanks for catching it, LGTM, merging :)

Issue #173, opened 2023-03-14
z_keyexpr_equals API differs from zenoh-c [Bug] 


milyin:
### Describe the bug

In zenoh-pico `z_keyexpr_equals` returns 0 in case of success: 
https://github.com/eclipse-zenoh/zenoh-pico/blob/5b4734114fad6f7fc484322b53f0f02872171ac5/include/zenoh-pico/api/primitives.h#L212

In zenoh-c the success return value is 1

https://github.com/eclipse-zenoh/zenoh-c/blob/master/include/zenoh_commons.h#L1118

Supposedly there are more API differences like this. The common solution should be found

### To reproduce

-

### System info

-

p-avital:
`zenoh-c` is the one that's wrong here, would you mind patching it to follow pico?

It's kinda counter intuitive for `0` to be "equals" in a function that looks like a predicate for equality, but it's to allow error reporting and avoid treating errors as equality.

milyin:
I agree, but this also implies that return value should be bool, so the fix should be made in zenoh-c too. The int8_t return value is supposed to be used in sitiations where 0 is success and another value is error code.

p-avital:
you're correct, only zenoh-c needs fixing :)

milyin:
So what;s the decision: int8_t and 0 for success or bool and true for success for z_keyexpr_vals and z_keyexpr_intersects? I'd prefer bool, but it's up to you to decide

cguimaraes:
I am also inline with the latest agreement. We already open the space for more detailed error codes in the future.
Still, we need to remember to explicit mention this in the changelog and documentation, as it might impose changes in user code.

milyin:
The API change necessary is not clear yet. The zenohc makes the differences between false return and error return:
```
#[allow(clippy::missing_safety_doc)]
#[no_mangle]
/// Returns ``1`` if `left` and `right` define sets that have at least one key in common, ``0`` if they don't.
/// Returns negative values in case of error (if one of the key expressions is in an invalid state).
pub extern "C" fn z_keyexpr_intersects(left: z_keyexpr_t, right: z_keyexpr_t) -> i8 {
    match (&*left, &*right) {
        (Some(l), Some(r)) => l.intersects(r) as i8,
        _ => i8::MIN,
    }
}
```

The zenoh-pico always returns negative value, no matter is it the error case or normal negative result:
```
[...]

PR #174, opened 2023-03-16
return 0 on true, -1 on false, other negative on error

milyin:
API update accordingly to variant (1) proposed in this discussion https://github.com/eclipse-zenoh/zenoh-pico/issues/173


p-avital:
This would break any code that relies on `z_keyexpr_intersect(a, b) >= 0` to assert intersection. Do we take that risk? @cguimaraes

cguimaraes:
What worries me is that we might break the logic of anyone already making use of them. So, we will need to be clear on the change log and documentation.

But at the same time, I also agree that the functions below should return the same type:
```
int8_t z_keyexpr_intersects(z_keyexpr_t l, z_keyexpr_t r);
_Bool zp_keyexpr_intersect_null_terminated(const char *l, const char *r);
```

In between, Zenoh-Pico and Zenoh-C what does provide less impact for the user?

src/api/api.c
```diff
@@ -94,48 +94,77 @@ int8_t zp_keyexpr_canonize_null_terminated(char *start) {
...
+    return ret;
 }
 
-_Bool zp_keyexpr_intersect_null_terminated(const char *l, const char *r) {
-    return _z_keyexpr_intersects(l, strlen(l), r, strlen(r));
+int8_t zp_keyexpr_intersect_null_terminated(const char *l, const char *r) {
+    int8_t ret = -1;
+
+    if (l != NULL && r != NULL) {
+        ret = (_z_keyexpr_intersects(l, strlen(l), r, strlen(r)) == true) ? 0 : -1;
```
p-avital, line 136: why use `== true` for a ternary? MISRA?

p-avital (approved):

src/api/api.c
cguimaraes: Indeed, the explicitly comparison was introduced on the codebase with the MISRA compliance.

cguimaraes (approved):

PR #176, opened 2023-03-20
Fixed wrong type in serial CRC function

hairymnstr:
The CRC function was calculating the wrong CRC due to a regression in the variable type.

### DIFF
--- a/src/utils/checksum.c
+++ b/src/utils/checksum.c
@@ -15,11 +15,11 @@
 #include "zenoh-pico/utils/checksum.h"
 
 uint32_t _z_crc32(const uint8_t *message, size_t len) {
-    int32_t crc = 0xFFFFFFFF;
+    uint32_t crc = 0xFFFFFFFF;
     for (size_t i = 0; i < len; i++) {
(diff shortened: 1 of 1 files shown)


cguimaraes:
Is this introducing any any warning or error in your setup / scenario?
This change of type was intentionally introduced by the Windows port.

hairymnstr:
Without this fix I get CRC errors when talking to zenohd over the serial transport:

```[2023-03-21T07:48:37Z ERROR zenoh_link_serial::unicast] Read error on Serial link serial//dev/ttyACM0 => serial/f7c0887d-f068-4ba2-910a-b2e7deb108ee: CRC does not match Received 54CF142 Computed F82CABD3 at io/zenoh-links/zenoh-link-serial/src/unicast.rs:150.```

hairymnstr:
I've tested the code with your commits and it's working for me.

The bit with the issue was `(crc >> 1)` which is an arithmetic shift (i.e. sign extends in signed types).

cguimaraes:
Fixes #177 .

cguimaraes (commented):
LGTM

PR #182, opened 2023-03-25
Align drop macros with Zenoh-C

### LINKED ISSUE #181: difference in z_drop macro in zenohc and zenoh-pico
In zenoh-pico macro z_drop calls one of z_XXX_drop functions:

https://github.com/eclipse-zenoh/zenoh-pico/blob/master/include/zenoh-pico/api/macros.h#L52-L65

in zenoh-c the same macro calls higher level function depending on type, if available (z_close for session, z_udeclare_subscriber for subscriber)

https://github.com/eclipse-zenoh/zenoh-c/blob/master/include/zenoh_macros.h#L16-L35

Both approaches makes sense but it's important to have same logic in both projects.

Also for the zenoh-cpp project the zenoh-c approach is preferrable, as it allows to make correct resource cleaning with single template.

cguimaraes:
Fixes #181

### DIFF
--- a/src/api/api.c
+++ b/src/api/api.c
@@ -29,6 +29,7 @@
 #include "zenoh-pico/session/queryable.h"
 #include "zenoh-pico/session/resource.h"
 #include "zenoh-pico/session/utils.h"
+#include "zenoh-pico/utils/result.h"
 
 /********* Data Types Handlers *********/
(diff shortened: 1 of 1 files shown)


cguimaraes:
@milyin can you review it please?

milyin:
Looks good for me, thank you!

cguimaraes:
Superseded by #183

PR #184, opened 2023-03-28
user callback can take ownership on reply or hello

milyin:
zenohc allows user callback to take ownership of owned structure passed to callback with mutable reference. This is the case for `z_hello_t` and `z_reply_t`. This behavior is explicitly documented: https://zenoh-c.readthedocs.io/en/latest/api.html#id6

This fix enables the same functionality for zenoh-pico

### DIFF
--- a/examples/unix/c11/z_get.c
+++ b/examples/unix/c11/z_get.c
@@ -33,6 +33,11 @@ void reply_handler(z_owned_reply_t *reply, void *ctx) {
     } else {
         printf(">> Received an error\n");
     }
+
+    // Callback is allowed to take ownership of the reply and drop it later
+    z_owned_reply_t saved = *reply;
+    z_null(reply);
+    z_drop(z_move(saved));
 }
 
 int main(int argc, char **argv) {
--- a/examples/unix/c11/z_get.c
+++ b/examples/unix/c11/z_get.c
@@ -33,11 +33,6 @@ void reply_handler(z_owned_reply_t *reply, void *ctx) {
     } else {
         printf(">> Received an error\n");
     }
-
-    // Callback is allowed to take ownership of the reply and drop it later
-    z_owned_reply_t saved = *reply;
-    z_null(reply);
-    z_drop(z_move(saved));
 }
 
 int main(int argc, char **argv) {
--- a/include/zenoh-pico/session/session.h
+++ b/include/zenoh-pico/session/session.h
@@ -130,7 +130,7 @@ _Z_ELEM_DEFINE(_z_questionable_sptr, _z_questionable_sptr_t, _z_noop_size, _z_qu
 _Z_LIST_DEFINE(_z_questionable_sptr, _z_questionable_sptr_t)
 
 typedef struct {
-    _z_reply_t _reply;
+    _z_reply_t *_reply;
     _z_timestamp_t _tstamp;
 } _z_pending_reply_t;
 
--- a/include/zenoh-pico/session/session.h
+++ b/include/zenoh-pico/session/session.h
@@ -130,7 +130,7 @@ _Z_ELEM_DEFINE(_z_questionable_sptr, _z_questionable_sptr_t, _z_noop_size, _z_qu
 _Z_LIST_DEFINE(_z_questionable_sptr, _z_questionable_sptr_t)
 
 typedef struct {
-    _z_reply_t *_reply;
+    _z_reply_t _reply;
     _z_timestamp_t _tstamp;
 } _z_pending_reply_t;
 
--- a/src/collections/list.c
+++ b/src/collections/list.c
@@ -48,25 +48,16 @@ size_t _z_list_len(const _z_list_t *xs) {
 
 _Bool _z_list_is_empty(const _z_list_t *xs) { return _z_list_len(xs) == (size_t)0; }
 
-// pops element from _z_list_t and frees it
-_z_list_t *_z_list_pop_drop(_z_list_t *xs, z_element_free_f f_f)
+_z_list_t *_z_list_pop(_z_list_t *xs, z_element_free_f f_f, void** x)
 {
     _z_list_t *l = (_z_list_t *)xs;
     _z_list_t *head = xs;
     l = head->_tail;
-    f_f(&head->_val);
-    z_free(head);
-
-    return l;
-}
-
-// pops element from _z_list_t and returns it in second argument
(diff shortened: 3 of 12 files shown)


src/session/query.c
```diff
@@ -252,8 +267,9 @@ int8_t _z_trigger_query_reply_final(_z_session_t *zn, const _z_reply_context_t *
             _z_pending_reply_t *pen_rep = _z_pending_reply_list_head(pen_qry->_pending_replies);
 
             // Trigger the query handler
-            pen_qry->_callback(&pen_rep->_reply, pen_qry->_call_arg);
-            pen_qry->_pending_replies = _z_pending_reply_list_pop(pen_qry->_pending_replies);
+            pen_qry->_callback(pen_rep->_reply, pen_qry->_call_arg);
+            pen_rep->_reply = NULL; // The  callback took ownership of the reply
```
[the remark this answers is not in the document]
cguimaraes: Indeed. This callback is the one that is wrapping the reply into an owned type, which is part of the API layer.

src/collections/list.c
```diff
@@ -48,25 +48,16 @@ size_t _z_list_len(const _z_list_t *xs) {
...
-}
-
-// pops element from _z_list_t and returns it in second argument
-_z_list_t *_z_list_pop_get(_z_list_t *xs, void **x)
-{
-    _z_list_t *l = (_z_list_t *)xs;
-    _z_list_t *head = xs;
-    l = head->_tail;
-    *x = head->_val;
+    if (x) {
```
cguimaraes, line 56: Let's make the `x != NULL` explicit.
MISRA aspects.

examples/unix/c11/z_get.c
```diff
@@ -33,6 +33,11 @@ void reply_handler(z_owned_reply_t *reply, void *ctx) {
     } else {
         printf(">> Received an error\n");
     }
+
+    // Callback is allowed to take ownership of the reply and drop it later
+    z_owned_reply_t saved = *reply;
+    z_null(reply);
+    z_drop(z_move(saved));
```
cguimaraes, line 40: I would not put this in the examples, but instead mention how to do it in the documentation.

src/net/primitives.c
```diff
@@ -31,9 +31,9 @@ void _z_scout(const z_whatami_t what, const char *locator, const uint32_t timeou
     _z_hello_list_t *hellos = _z_scout_inner(what, locator, timeout, false);
 
     while (hellos != NULL) {
-        _z_hello_t *hello = _z_hello_list_head(hellos);
-        (*callback)(hello, arg_call);
-        hellos = _z_hello_list_pop(hellos);
+        _z_hello_t *hello;
```
cguimaraes, line 34: `_z_hello_t *hello = NULL`

include/zenoh-pico/session/session.h
```diff
@@ -130,7 +130,7 @@ _Z_ELEM_DEFINE(_z_questionable_sptr, _z_questionable_sptr_t, _z_noop_size, _z_qu
 _Z_LIST_DEFINE(_z_questionable_sptr, _z_questionable_sptr_t)
 
 typedef struct {
-    _z_reply_t _reply;
+    _z_reply_t *_reply;
```
cguimaraes, line 133: Just to confirm, you had to change it to a pointer in the heap in order to allow the user-code to get the ownership of the reply without an extra copy. 
Can you confirm if this was the case?

src/session/query.c
```diff
@@ -22,6 +22,15 @@
...
 
+_z_reply_t *_z_reply_new(_z_reply_t _reply)
+{
+    _z_reply_t *reply = (_z_reply_t *)z_malloc(sizeof(_z_reply_t));
+    if (reply != NULL) {
+        *reply = _reply;
+    }
+    return reply;
+}
+
```
cguimaraes, line 33: This might be error prone if reply is cleared it will affect the new reply.
I think here we should use the smart pointer helpers that @Mallets implemented some time ago.

src/session/query.c
```diff
@@ -41,7 +50,10 @@ _Bool _z_pending_reply_eq(const _z_pending_reply_t *one, const _z_pending_reply_
 
 void _z_pending_reply_clear(_z_pending_reply_t *pr) {
     // Free reply
-    _z_reply_clear(&pr->_reply);
+    if (pr->_reply != NULL) {
+        _z_reply_clear(pr->_reply);
+        z_free(pr->_reply);
```
cguimaraes, line 55: `_z_reply_free` is the same as these two operations.
Use it instead.

This applies to any type that you can clear you can also free it.

src/session/query.c
```diff
@@ -198,9 +210,12 @@ int8_t _z_trigger_query_reply_partial(_z_session_t *zn, const _z_reply_context_t
...
                     partial_reply.data.sample.keyexpr = _z_keyexpr_duplicate(&reply.data.sample.keyexpr);
-                    pen_rep->_reply = partial_reply;
+                    pen_rep->_reply = _z_reply_new(partial_reply);
                 } else {
-                    pen_rep->_reply = reply;  // Store the whole reply in the latest mode
+                    pen_rep->_reply = _z_reply_new(reply); // Store the whole reply in the latest mode
+                }
+                if (pen_rep->_reply == NULL) {
+                    ret = _Z_ERR_SYSTEM_OUT_OF_MEMORY;
                 }
```
cguimaraes, line 219: if it fails due to out of memory, we should clean any function state and go to the return point.
Here the timestamp is duplicating as well, and it will still be a problem with out of memory.

cguimaraes (changes requested):
Small changes requests

include/zenoh-pico/session/session.h
milyin: Yes, that's the goal. _z_reply_t as a part of larger structure can't change ownership

src/session/query.c
milyin: Could you please point me some sample of these smart pointers usage?

include/zenoh-pico/session/session.h
cguimaraes: Smart pointer definition: https://github.com/eclipse-zenoh/zenoh-pico/blob/master/include/zenoh-pico/collections/pointer.h
Example of usage: https://github.com/eclipse-zenoh/zenoh-pico/search?q=sptr

cguimaraes (approved):
LGTM

PR #189, opened 2023-03-30
make CMakeLists.txt includable

milyin:
Allow external projects to use zenohpico project including it as a subdirectory into customer's CMake project.
This mode of usage is demonstrated in separate CMakeLists.txt in examples directory. In addition to normal build of examples, where examples included as a subdirectrory of zenohpico project, "examples" project can be built as a root project itself. In this case it includes it's parent directory as a dependency.

```sh
git clone https://github.com/eclipse-zenoh/zenoh-pico
mkdir foo
cd foo
cmake ../zenoh-pico/examples
cmake --build .
```


CMakeLists.txt
```diff
@@ -37,49 +61,34 @@ if(NOT CMAKE_BUILD_TYPE)
...
-option(BUILD_TOOLS "Use this to also build the tools." OFF)
-option(BUILD_TESTING "Use this to also build tests." ON)
-option(BUILD_INTEGRATION "Use this to also build integration tests." OFF)
-option(ZENOH_DEBUG "Use this to set the ZENOH_DEBUG variable." 0)
-option(WITH_ZEPHYR "Build for Zephyr RTOS" OFF)
-
 if(CMAKE_SYSTEM_NAME MATCHES "Linux")
-  add_definitions(-DZENOH_LINUX)
+  add_definition(ZENOH_LINUX)
   set(JNI_PLATFORM_NAME "linux")
```
cguimaraes, line 66: We can leverage on this PR to remove this that is an old artifact.

cguimaraes: I refer to the JNI_PLATFORM_NAME

CMakeLists.txt
cguimaraes, line 1: Can you apply the same indentation space as the other CMake-related files?

milyin: Done - formatted using https://cmake-format.readthedocs.io/en/latest/cmake-format.html, added format configuration for future use

cguimaraes (approved):

cguimaraes (approved):

PR #203, opened 2023-04-13
Added a way to manually batch put messages

hairymnstr:
This adds a z_put_multi API call which puts all the payloads passed into a single zenoh frame to reduce overhead and improve UDP efficiency without reducing granularity of keyexpr.

I considered 2 approaches,
1. Payload as a vector (which I implemented)
2. Add a batching and flush mechanism

The first approach uses a bit more memory since you must have all payload and keys in memory at once before calling.  However the goal was to reduce overhead when pushing lots of small payloads so the memory footprint is still small.

The second approach raised issues for synchronisation, if another thread wanted to do a put before flushing you could get a mix of priorities or something.

Known issues: I've not implemented encoding as a vector but it should be so different topics can have different data types even if they are published in a group.


cguimaraes:
Thank you @hairymnstr .
We have been discussing if we shall adopt your proposal or to go for something more similar to your second approach.

In the end, we decided to go for the second approach and we will be working on it. In any case, we really value your contribution and we would like to have some feedback from you when it is done.

I will keep this PR open until further notice.

Mallets:
Reverted to issue #298

PR #208, opened 2023-04-25
Added CLOSE message handling

hairymnstr:
There was nothing making the read task finish if it received a CLOSE message.  This meant that if zenohd was restarted a UDP publisher may continue bombarding it with publications indefinitely while it replied with CLOSE which was ignored.


src/transport/multicast/link/task/read.c
```diff
@@ -91,14 +91,18 @@ void *_zp_multicast_read_task(void *ztm_arg) {
             ret = _z_transport_message_decode_na(&t_msg, &zbuf);
             if (ret == _Z_RES_OK) {
                 ret = _z_multicast_handle_transport_message(ztm, &t_msg, &addr);
-
+                bool should_close = (_Z_MID(t_msg._header) == _Z_MID_CLOSE);
```
Mallets, line 94: The read task in multicast should never be stopped in response of a close message.
The same read task is used to read all multicast messages from all the peers, so one peer could stop the communication with all the other peers.

Hence, proposed changes to multicast should be reverted.

src/transport/unicast/link/task/read.c
```diff
@@ -89,12 +89,17 @@ void *_zp_unicast_read_task(void *ztu_arg) {
 
         if (ret == _Z_RES_OK) {
             ret = _z_unicast_handle_transport_message(ztu, &t_msg);
+            bool should_close = (_Z_MID(t_msg._header) == _Z_MID_CLOSE);
```
Mallets, line 92: The message inspection should be done by `_z_unicast_handle_transport_message` which in turn should return an appropriate error code. If `ret != _Z_RES_OK` then the task is already marked as non running at line `95` below.

Mallets (changes requested):

cguimaraes:
I think the solution shall be target in a different place of the codebase. Namely, on the handler of the CLOSE message, by triggering the `_z_close` operation.
This is signaling the tasks to terminate on their next loop, thus not requiring to be complexified with extra control variables and checks.

cguimaraes (changes requested):
I think the solution shall be target in a different place of the codebase. Namely, on the handler of the CLOSE message, by triggering the _z_close operation.
This is signaling the tasks to terminate on their next loop, thus not requiring to be complexified with extra control variables and checks.

PR #212, opened 2023-04-27
provide ping/pong examples

p-avital:


examples/CMakeLists.txt
```diff
@@ -53,4 +57,6 @@ elseif(MSVC)
     add_example(z_queryable windows/z_queryable.c)
     add_example(z_info windows/z_info.c)
     add_example(z_scout windows/z_scout.c)
+    add_example(z_ping windows/z_ping.c)
```
cguimaraes, line 60: These are not windows friendly :-)
Exclude them here if not ported for Windows.

examples/unix/c99/z_ping.c
```diff
@@ -0,0 +1,128 @@
...
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <time.h>
+
+#include "zenoh-pico.h"
+#include "zenoh-pico/api/primitives.h"
+#include "zenoh-pico/system/platform.h"
+
+_z_condvar_t cond;
```
cguimaraes, line 11: We shall consider making all these types public, if we plan to make use of them.
I might target it on a different PR to not pollute this one.

examples/CMakeLists.txt
p-avital: Since they are ported using pico's abstraction of mutex and condvar, they should work on any platform pico supports, so they are windows friendly (I had just forgotten to remove the include x) )

examples/unix/c99/z_ping.c
p-avital: Agreed, and zenoh-c should probably export the same condvar/mutex/malloc equivalents too :)

cguimaraes: well...then let's hold and do them at the same time for the after release IMO.

Mallets (approved):

Issue #214, opened 2023-05-11
.cmake file for find_package() is not being installed [Bug] 


milyin:
The projects https://github.com/eclipse-zenoh/zenoh-c and https://github.com/eclipse-zenoh/zenoh-cpp on `install` command installs not only library and headers, but also fies
```
/usr/local/lib/pkgconfig/${libname}.pc
/usr/local/lib/cmake/${libname}/${libname}Config.cmake
/usr/local/lib/cmake/${libname}/${libname}ConfigVersion.cmake
```
These files are used by CMake's `find_package` command. 

`zenohpico` doens't do that, so `find_package(zenohpico)` doesn't work.

In https://github.com/eclipse-zenoh/zenoh-cpp/blob/main/README.md#library-usage it's recommended to include `zenohpico` with `find_package`, which doesn't work in fact and causes confusion to users

PR #222, opened 2023-06-16
Add pc files upon installation (#214)

cguimaraes:
Fixes https://github.com/eclipse-zenoh/zenoh-pico/issues/214


cguimaraes:
@milyin can you double check if it solves the issue you identified in https://github.com/eclipse-zenoh/zenoh-pico/issues/214?

CMakeLists.txt
```diff
@@ -217,6 +219,31 @@ install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/include/zenoh-pico
   COMPONENT Headers
 )
 
+set(CMAKE_INSTALL_CMAKEDIR "/usr/lib/cmake/zenohpico")
```
milyin, line 222: It's better to use ${CMAKE_INSTALL_LIBDIR} instead of "/usr/lib"

milyin (changes requested):

CMakeLists.txt
cguimaraes: It shall be fixed now.

milyin (approved):

PR #223, opened 2023-06-19
WIP: Protocol 0.8

p-avital:


src/transport/unicast/link/rx.c
```diff
@@ -120,7 +121,8 @@ int8_t _z_unicast_handle_transport_message(_z_transport_unicast_t *ztu, _z_trans
             size_t len = _z_vec_len(&t_msg->_body._frame._messages);
             for (size_t i = 0; i < len; i++) {
                 _z_handle_zenoh_message(ztu->_session,
```
Mallets, line 123: Not the goal of this commit, but for consistency with the new protocol concepts `_z_handle_zenoh_message` should be renamed `_z_handle_network_message`.

include/zenoh-pico/collections/bytes.h
```diff
@@ -34,9 +34,11 @@ typedef struct {
 } _z_bytes_t;
 
 _z_bytes_t _z_bytes_empty(void);
+inline static _Bool _z_bytes_check(_z_bytes_t value) { return value.start != NULL; }
 int8_t _z_bytes_init(_z_bytes_t *bs, size_t capacity);
 _z_bytes_t _z_bytes_make(size_t capacity);
 _z_bytes_t _z_bytes_wrap(const uint8_t *bs, size_t len);
+_z_bytes_t _z_bytes_steal(_z_bytes_t *b);
```
cguimaraes, line 41: Wasn't '_z_bytes_move` already implementing the same behavior?

include/zenoh-pico/net/query.h
```diff
@@ -23,7 +25,7 @@
 typedef struct {
     _z_value_t _value;
     _z_keyexpr_t _key;
-    _z_zint_t _qid;
+    uint32_t _rid;
```
cguimaraes, line 28: `z_declare_resource` returns an `uint16_t`-size `id`.
Any reason for a wider range here?

include/zenoh-pico/net/query.h
```diff
@@ -33,7 +35,7 @@ typedef struct {
  * Return type when declaring a queryable.
  */
 typedef struct {
-    _z_zint_t _id;
+    uint32_t _id;
```
cguimaraes, line 38: Similar comment here, but I did not yet check the qid return size.

include/zenoh-pico/net/subscribe.h
```diff
@@ -15,14 +15,16 @@
...
+
 #include "zenoh-pico/net/session.h"
 #include "zenoh-pico/protocol/core.h"
 
 /**
  * Return type when declaring a subscriber.
  */
 typedef struct {
-    _z_zint_t _id;
+    uint32_t _id;
```
cguimaraes, line 27: Same comment here as for the queryable

src/link/link.c
```diff
@@ -148,8 +148,7 @@ size_t _z_link_recv_exact_zbuf(const _z_link_t *link, _z_zbuf_t *zbf, size_t len
 
 int8_t _z_link_send_wbuf(const _z_link_t *link, const _z_wbuf_t *wbf) {
     int8_t ret = _Z_RES_OK;
-
-    for (size_t i = 0; (i < _z_wbuf_len_iosli(wbf)) || (ret == -1); i++) {
+    for (size_t i = 0; (i < _z_wbuf_len_iosli(wbf)); i++) {
```
cguimaraes, line 151: if an `iosli` fails to send and a subsequent one succeeds, this function will return `_Z_RES_OK` and it shall not be the case.

src/protocol/core.c
```diff
@@ -0,0 +1,52 @@
...
+}
+_Bool _z_id_check(_z_id_t id) {
+    _Bool ret = false;
+    for (int i = 0; !ret && i < 16; i++) {
+        ret |= id.id[i] != 0;
+    }
+    return ret;
+}
+_z_id_t _z_id_empty() {
+    return (_z_id_t){.id = {
```
cguimaraes, line 24: Please confirm, but I think: `return (_z_id_t){.id = {0}};` would initialize the entire array with 0s.

cguimaraes (commented):
This was a quick and light review on my side.
I suggest @Mallets to do a more detailed one, including double checking the message formats and constant values with the Rust implementation.

include/zenoh-pico/net/query.h
p-avital: Here, rid stands for request id. Changing the name to avoid confusions

include/zenoh-pico/net/query.h
p-avital: This here is an entity id, they're u32

include/zenoh-pico/collections/bytes.h
p-avital: it was indeed, didn't spot it because the behaviour differs to other moves. I think `_z_bytes_move` will be renamed to the steal version, to avoid being confused for the move convention

include/zenoh-pico/net/subscribe.h
p-avital: entity_id, u32

src/link/link.c
p-avital: so here, the behaviour should depend on whether or not the link is streamed:
- on streamed links, keep trying to send what hasn't been sent.
- on datagram links, give up after the first failure, as only fragmented payloads have multi-slice wbufs, which means missing one is a full on transmit failure for the whole thing

PR #225, opened 2023-06-28
Append newline to some messages

clachan:


cguimaraes:
I would prefer to see this fixed generally in the logging macros available at `include/zenoh-pico/utils/logging.h`.

In doing so, we can omit the `\n` whenever making use of the logging macros.

cguimaraes (changes requested):
I would prefer to see this fixed generally in the logging macros available at `include/zenoh-pico/utils/logging.h`.

In doing so, we can omit the `\n` whenever making use of the logging macros.

clachan:
> I would prefer to see this fixed generally in the logging macros available at `include/zenoh-pico/utils/logging.h`.
> 
> In doing so, we can omit the `\n` whenever making use of the logging macros.

Hi @cguimaraes, thanks for your feedback.

I agree with you that one could omit `\n` by adding some changes in `logging.h` that inserts the trailing `\n` automatically. The concern is, once added, one no longer has a way to explicitly remove `\n` if they really want to.

Is this really preferred?

cguimaraes:
The macros will pre-append the timestamp, log level, and function to the argument message.
 See example below:

```
[2023-07-03T15:08:04Z DEBUG ::_z_scout_encode] Encoding _Z_MID_SCOUT
[2023-07-03T15:08:04Z DEBUG ::_z_hello_decode_na] Decoding _Z_MID_HELLO
[2023-07-03T15:08:04Z DEBUG ::_z_locators_decode_na] Decoding _LOCATORS
```

Having the `\n` automatically set would avoid the log to seem wrongly formatted.
This is mainly my concern since we are touching this part of the codebase.

p-avital:
While I agree with @cguimaraes, I think the work he proposes has a different scale to this PR, I'll open an issue on that mentioning this PR.

A lot of this code has been changed in a very large PR that should land soon-ish, so I'd rather close this PR than merge it.

PR #235, opened 2023-08-03
"docker rmi " of 'FROM' docker fails now

milyin:
Since today it seems that base docker image (the one which was used in "FROM" clause) doesn't appear in the list of the images. Attempt to remove it causes build failure.
Removed command "dorcker rmi" for this image to fix the build failure.


cguimaraes (changes requested):
Shouldn't we fix the image being fetched instead?

Mallets:
I've reproduced the same behaviour on an Ubuntu machine. It seems that Docker behaviour changed and the image does not appear anymore on the image list and `docker rmi` fails. Nevertheless, the whole crossbuilding process is successful. Therefore, I'm in favour for approving this PR.

cguimaraes (approved):

PR #237, opened 2023-08-04
Fix the ReadTheDocs clang dependency

evshary:
It seems like ReadTheDocs uses Ubuntu-22.04 instead Ubuntu-18.04 now.
The clang should also be updated.
I generated my own ReadTheDocs to see whether it makes any difference.

Original one: https://zenoh-pico.readthedocs.io/en/0.7.2-rc/index.html
After update: https://evshary-zenoh-pico.readthedocs.io/en/latest/index.html

There are some display differences on the web page. 
@Mallets Could you please check whether this is acceptable or not?

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -10,6 +10,8 @@ build:
   os: ubuntu-22.04
   tools:
     python: "3.11"
+  apt_packages:
+    - libclang-dev
 
 # Build documentation in the docs/ directory with Sphinx
 sphinx:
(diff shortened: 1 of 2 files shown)


cguimaraes (approved):
LGTM. Were you able to test it in a private repo?

evshary:
@cguimaraes Thank you!
I'm not sure whether ReadTheDocs accepts the private repo or not. I need to check.
Anyway, the test page is only for you to check the differences, so I'll delete it after the PR is merged to avoid confusion among users.

Mallets (approved):

PR #239, opened 2023-08-11
FreeRTOS-Plus-TCP port

bjsowa:
Features to implement:

- [x] UDP Unicast transport
- [x] Build tests
- [ ] ~UDP Multicast transport~
- [x] TCP transport
- [x] Multi threading
- [ ] ~(optional) IPv6 support~


cguimaraes:
@bjsowa check failing test regarding the Eclipse ECA. Otherwise, we will not be able to merge this PR.

Do you have a timeline for the remaining features?

bjsowa:
> @bjsowa check failing test regarding the Eclipse ECA. Otherwise, we will not be able to merge this PR.

Fixed

> Do you have a timeline for the remaining features?

I don't.

src/system/freertos_plus_tcp/system.c
```diff
@@ -0,0 +1,86 @@
...
+
+int z_sleep_s(size_t time) {
+    vTaskDelay(pdMS_TO_TICKS(time * 1000));
+    return 0;
+}
+
+/*------------------ Clock ------------------*/
+z_clock_t z_clock_now(void) { return z_time_now(); }
+
+unsigned long z_clock_elapsed_us(z_clock_t *instant) { return z_clock_elapsed_ms(instant) * 1000; }
```
cguimaraes, line 63: This will sleep seconds instead of useconds.

src/system/freertos_plus_tcp/network.c
```diff
@@ -0,0 +1,106 @@
...
+#if Z_LINK_UDP_UNICAST == 1 || Z_LINK_UDP_MULTICAST == 1
+/*------------------ UDP sockets ------------------*/
+int8_t _z_create_endpoint_udp(_z_sys_net_endpoint_t *ep, const char *s_address, const char *s_port) {
+    int8_t ret = _Z_RES_OK;
+
+    if (FreeRTOS_getaddrinfo(s_address, NULL, NULL, &ep->_iptcp) < 0) {
+        ret = _Z_ERR_GENERIC;
+    }
+
+    ep->_iptcp->ai_addr->sin_family = FREERTOS_AF_INET4;
```
cguimaraes, line 20: isn't this value set by `getaddrinfo`?

cguimaraes (changes requested):
A first quick review.
Tests must be also provided...at least build tests.

src/system/freertos_plus_tcp/system.c
bjsowa: Not sure I follow, the conversion here seems correct to me. 1 ms = 1000 us

cguimaraes: This comment should have been here:
```
unsigned long z_clock_elapsed_s(z_clock_t *instant) { return z_time_elapsed_ms(instant) * 1000; }
```

Moving it to the proper line.
You can mark it as resolved here.

src/system/freertos_plus_tcp/system.c
```diff
@@ -0,0 +1,86 @@
...
+
+unsigned long z_clock_elapsed_ms(z_clock_t *instant) { return z_time_elapsed_ms(instant); }
+
+unsigned long z_clock_elapsed_s(z_clock_t *instant) { return z_time_elapsed_ms(instant) * 1000; }
```
cguimaraes, line 67: This will return seconds but useconds instead.

cguimaraes (changes requested):

src/system/freertos_plus_tcp/system.c
bjsowa: Fixed it. It is worth noting that the emscripten port, which I used for reference, has the same issue.

src/system/freertos_plus_tcp/network.c
bjsowa: Unfortunately not, `getaddrinfo` only sets `ai_family` but does not write it to `ai_addr->sin_family`.  I changed it so it at least uses the value returned by getaddrinfo.

bjsowa:
@cguimaraes I added TCP support, some examples and CI workflow. Could you review the changes?

The examples use POSIX port of FreeRTOS and network interface implementation that uses SLiRP library to emulate a userspace NAT network. I was able to use the examples on Linux in client mode with a Zenoh Router running on the host. Peer mode won't work as this port does not support incoming connections.

bjsowa:
@cguimaraes I added Multi threading support and extended Zenoh-pico API to allow passing port-specific task attributes when creating read and lease tasks (example usage in `z_pub.c`). Let me know what you think about it.

Multicast is not fully supported in FreeRTOS-Plus-TCP (see FreeRTOS/FreeRTOS-Plus-TCP#962) so I'll have to pass on UDP Multicast for now. I'll also pass on IPv6 support as I don't need it.

I'll mark this PR as Ready for review once I'll have it tested on a microcontroller.

bjsowa:
I updated the PR. @p-avital Could you rerun the workflows?

Mallets:
@bjsowa I've just realised that this PR is still open. In the meanwhile we have released zenoh 0.10.0 and it seems there are some conflicts to be resolved before merging.

bjsowa:
> @bjsowa I've just realised that this PR is still open. In the meanwhile we have released zenoh 0.10.0 and it seems there are some conflicts to be resolved before merging.

@Mallets I resolved the conflicts.

cguimaraes (approved):
My previous comments are addressed

Mallets (approved):

PR #240, opened 2023-08-16
Fix transport/rx unused-variable warnings

PetervdPerk-NXP:
I'm compiling with werror thus the unused zbuf variable come up as an errror.
Since they aren't used I'm just calling `_z_zbuf_view`


p-avital:
Hi, 
Thanks for the fix :)
However, these line will conflict with the massive PR that's coming for the new protocol, so we won't merge it. I'm keeping the PR open as a reminder to fix that on the new-protocol branch, although I think I already did :)

PetervdPerk-NXP:
@p-avital what would be the schedule for this ''new protocol" PR?

p-avital:
@PetervdPerk-NXP most likely, we'll merge it into main near the start of September.

PR #248, opened 2023-09-13
add -l option to examples

p-avital:


examples/unix/c11/z_get.c
```diff
@@ -53,11 +53,15 @@ int main(int argc, char **argv) {
             case 'm':
                 mode = optarg;
                 break;
+            case 'l':
+                locator = optarg;
+                mode = "peer";
```
Mallets, line 58: In addition to that, configuration should be updated accordingly. I.e.: in case of `-e` the config is:
 `zp_config_insert(z_loan(config), Z_CONFIG_CONNECT_KEY, z_string_make(locator));`
 
In case of `-l` the config should be:
 ``zp_config_insert(z_loan(config), Z_CONFIG_LISTEN_KEY, z_string_make(locator));``
 
 This comment applies also to all the other examples.

Mallets (changes requested):

examples/unix/c11/z_get.c
cguimaraes: It seems to me that using `Z_CONFIG_LISTEN_KEY` would imply more changes. At least from a quick search it is not used anywhere in the code.

Mallets (approved):

PR #256, opened 2023-10-04
Remove typos in constants and comments

jean-roland:
Going through the repo, I noticed some typos in both comments and constants and decided to fix them.

### DIFF
--- a/include/zenoh-pico/link/link.h
+++ b/include/zenoh-pico/link/link.h
@@ -47,18 +47,18 @@
  *
  * Enumerators:
  *     Z_LINK_CAPABILITY_NONE: Bitmask to define that link has no capabilities.
- *     Z_LINK_CAPABILITY_RELIEABLE: Bitmask to define and check if link is reliable.
+ *     Z_LINK_CAPABILITY_RELIABLE: Bitmask to define and check if link is reliable.
  *     Z_LINK_CAPABILITY_STREAMED: Bitmask to define and check if link is streamed.
  *     Z_LINK_CAPABILITY_MULTICAST: Bitmask to define and check if link is multicast.
  */
 typedef enum {
     Z_LINK_CAPABILITY_NONE = 0x00,       // 0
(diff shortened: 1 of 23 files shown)


include/zenoh-pico/link/link.h
```diff
@@ -47,18 +47,18 @@
...
  *     Z_LINK_CAPABILITY_NONE: Bitmask to define that link has no capabilities.
- *     Z_LINK_CAPABILITY_RELIEABLE: Bitmask to define and check if link is reliable.
+ *     Z_LINK_CAPABILITY_RELIABLE: Bitmask to define and check if link is reliable.
  *     Z_LINK_CAPABILITY_STREAMED: Bitmask to define and check if link is streamed.
  *     Z_LINK_CAPABILITY_MULTICAST: Bitmask to define and check if link is multicast.
  */
 typedef enum {
     Z_LINK_CAPABILITY_NONE = 0x00,       // 0
-    Z_LINK_CAPABILITY_RELIEABLE = 0x01,  // 1 << 0
+    Z_LINK_CAPABILITY_RELIABLE = 0x01,  // 1 << 0
```
p-avital, line 56: @Mallets considering out-of-tree transports, isn't that technically a breaking change?

If so, we could `#define Z_LINK_CAPABILITY_RELIEABLE Z_LINK_CAPABILITY_RELIABLE` just to keep from a typo-related API break :)

cguimaraes: Is there any out-of-tree transports that we are aware of?
If not, I would accept this change.
If so, I would suggest that they contribute it back.

cguimaraes (approved):
LGTM

include/zenoh-pico/link/link.h
Mallets: All Zenoh-Pico transports we are aware of are part of the zenoh-pico repository. So I'm ok with fixing the typo as suggested by the PR.

Mallets (approved):

PR #260, opened 2023-10-11
Add query & queryable token config

jean-roland:
This adds compile time config tokens to add or remove queries & queryables independently. The measured impact on program memory was, if you look at the sum of the symbol sizes from the Z_pub example elf, in O3 on amd64 linux:
* Full feature: 127 685B
* No queryable: 124 353B (-3332, 2.6%)
* No query & queryable: 120 087B (-4266, 3.4%)


src/api/api.c
```diff
@@ -654,14 +657,17 @@ typedef struct __z_reply_handler_wrapper_t {
 } __z_reply_handler_wrapper_t;
 
 void __z_reply_handler(_z_reply_t *reply, __z_reply_handler_wrapper_t *wrapped_ctx) {
```
cguimaraes, line 659: If queries are not supported, the entire reply handler can be ommitted.

cguimaraes, line 659: If queries are not supported, the entire reply handler can be ommitted.

src/api/api.c
```diff
@@ -654,14 +657,17 @@ typedef struct __z_reply_handler_wrapper_t {
...
+#if Z_FEATURE_QUERIES == 1
     z_owned_reply_t oreply = {._value = reply};
 
     wrapped_ctx->user_call(&oreply, wrapped_ctx->ctx);
     z_reply_drop(&oreply);  // user_call is allowed to take ownership of the reply by setting oreply._value to NULL
+#endif
 }
 
 int8_t z_get(z_session_t zs, z_keyexpr_t keyexpr, const char *parameters, z_owned_closure_reply_t *callback,
```
cguimaraes, line 668: If queries are disabled, I would completely omit any related APIs.

In doing so, the user will have an explicit error at compilation. Otherwise,  it will become harder for users (and us) to debug non-compliant behavior.

Also, you can save some extra bytes in the final binary.

src/api/api.c
```diff
@@ -896,6 +905,7 @@ z_queryable_options_t z_queryable_options_default(void) {
 
 z_owned_queryable_t z_declare_queryable(z_session_t zs, z_keyexpr_t keyexpr, z_owned_closure_query_t *callback,
```
cguimaraes, line 906: If queryables are disabled, I would completely omit any related APIs.

In doing so, the user will have an explicit error at compilation. Otherwise, it will become harder for users (and us) to debug non-compliant behavior.

Also, you can save some extra bytes in the final binary.

src/api/api.c
```diff
@@ -923,14 +933,20 @@ z_owned_queryable_t z_declare_queryable(z_session_t zs, z_keyexpr_t keyexpr, z_o
 
     return (z_owned_queryable_t){
         ._value = _z_declare_queryable(zs._val, key, opt.complete, callback->call, callback->drop, ctx)};
+#else
+    return (z_owned_queryable_t){._value = NULL};
+#endif
 }
 
 int8_t z_undeclare_queryable(z_owned_queryable_t *queryable) {
```
cguimaraes, line 941: If queryables are disabled, I would completely omit any related APIs.

In doing so, the user will have an explicit error at compilation. Otherwise, it will become harder for users (and us) to debug non-compliant behavior.

Also, you can save some extra bytes in the final binary.

src/api/api.c
```diff
@@ -940,6 +956,7 @@ z_query_reply_options_t z_query_reply_options_default(void) {
 
 int8_t z_query_reply(const z_query_t *query, const z_keyexpr_t keyexpr, const uint8_t *payload, size_t payload_len,
                      const z_query_reply_options_t *options) {
```
cguimaraes, line 958: If queryables are disabled, I would completely omit any related APIs.

In doing so, the user will have an explicit error at compilation. Otherwise, it will become harder for users (and us) to debug non-compliant behavior.

Also, you can save some extra bytes in the final binary.

src/api/api.c
```diff
@@ -949,6 +966,9 @@ int8_t z_query_reply(const z_query_t *query, const z_keyexpr_t keyexpr, const ui
                             },
                         .encoding = {.prefix = opts.encoding.prefix, .suffix = opts.encoding.suffix}};
     return _z_send_reply(query, keyexpr, value);
+#else
+    return _Z_ERR_GENERIC;
+#endif
 }
 
 _Bool z_reply_is_ok(const z_owned_reply_t *reply) {
```
cguimaraes, line 974: If queryables are disabled, I would completely omit any related APIs.

In doing so, the user will have an explicit error at compilation. Otherwise, it will become harder for users (and us) to debug non-compliant behavior.

Also, you can save some extra bytes in the final binary.

src/session/rx.c
```diff
@@ -154,9 +162,13 @@ int8_t _z_handle_network_message(_z_session_t *zn, _z_zenoh_message_t *msg, uint
             }
         } break;
         case _Z_N_RESPONSE_FINAL: {
+#if Z_FEATURE_QUERIES == 1
```
Mallets, line 165: `Response` and `ResponseFinal` are not limited to queries/queryables.

jean-roland:
Here are the modifications on the update:
* APIs were removed, this breaks examples compilation so for the moment being I replace the affected examples with a dummy main when necessary
* I tried to regroup the feature code to reduce #if directives as they make code harder to read
* I added some feature-agnostic "reply" and "push" functions

include/zenoh-pico/session/reply.h
```diff
@@ -0,0 +1,29 @@
...
+
+#include "zenoh-pico/net/session.h"
+#include "zenoh-pico/protocol/core.h"
+#include "zenoh-pico/protocol/definitions/message.h"
+#include "zenoh-pico/protocol/definitions/network.h"
+
+#ifndef ZENOH_PICO_SESSION_REPLY_H
+#define ZENOH_PICO_SESSION_REPLY_H
+
+int8_t _z_trigger_reply_partial(_z_session_t *zn, _z_zint_t id, _z_keyexpr_t key, _z_msg_reply_t *reply);
```
cguimaraes, line 25: Check for the features here as well.

src/session/reply.c
```diff
@@ -0,0 +1,44 @@
...
+
+int8_t _z_trigger_reply_partial(_z_session_t *zn, _z_zint_t id, _z_keyexpr_t key, _z_msg_reply_t *reply) {
+    int8_t ret = _Z_RES_OK;
+
+    // TODO check id to know where to dispatch
+
+#if Z_FEATURE_QUERY == 1
+    ret = _z_trigger_query_reply_partial(zn, id, key, reply->_value.payload, reply->_value.encoding, Z_SAMPLE_KIND_PUT,
+                                         reply->_timestamp);
+#endif
```
cguimaraes, line 30: Return error if not supported?

src/session/reply.c
```diff
@@ -0,0 +1,44 @@
...
+
+int8_t _z_trigger_reply_final(_z_session_t *zn, _z_n_msg_response_final_t *final) {
+    int8_t ret = _Z_RES_OK;
+
+    // TODO check id to know where to dispatch
+    _z_zint_t id = final->_request_id;
+
+#if Z_FEATURE_QUERY == 1
+    _z_trigger_query_reply_final(zn, id);
+#endif
```
cguimaraes, line 42: Return error if not supported?

cguimaraes (commented):
LGTM. A couple of comments only.

cguimaraes (approved):

include/zenoh-pico/session/reply.h
jean-roland: The overarching goal here, as mentioned by @Mallets, is to dissociate replies with requests as in the future there might be reply packets for other features.

Same for push packet and subscriptions.

Mallets (approved):

PR #262, opened 2023-10-12
fix memory leak in z_declare_subscriber

### LINKED ISSUE #259: [Bug] z_sub example leaks allocated suffix in z_declare_subscriber
### Describe the bug

The example z_sub.c calls [`z_declare_subscriber()`](https://github.com/eclipse-zenoh/zenoh-pico/blob/c4f78b2165741578a259a72eab140afa5233303b/examples/unix/c11/z_sub.c#L87) with the keyexpr "demo/example/**".
This causes the `z_declare_subscriber()` to [allocate memory](https://github.com/eclipse-zenoh/zenoh-pico/blob/c4f78b2165741578a259a72eab140afa5233303b/src/api/api.c#L827) for a suffix that is never freed before the program ends.
Calling `_z_keyexpr_clear(&keyexpr)` after `_z_declare_resource()` removes the leak but I am unsure if this is the correct solution as I am not familiar with the zenoh-pico codebase.

### To reproduce

1. Console 1: zenohd -l tcp/localhost:7447 --no-multicast-scouting
2. Console 2: build/examples/z_sub -m client -e tcp/127.0.0.1:7447
3. Console 2: Press 'q' to quit

### System info

Distributor ID: Ubuntu
Description:    Ubuntu 22.04.3 LTS
Release:    22.04
Codename:   jammy

CPU: Intel(R) Core(TM) i9-7920X CPU @ 2.90GHz

zenoh commit: 1f935b16baaaf45c0c18c8fafd0d03b774f4b4ca
zenoh-pico commit: c4f78b2165741578a259a72eab140afa5233303b

p-avital:
fixes #259

### DIFF
--- a/src/api/api.c
+++ b/src/api/api.c
@@ -826,10 +826,12 @@ z_owned_subscriber_t z_declare_subscriber(z_session_t zs, z_keyexpr_t keyexpr, z
                 wild -= 1;
                 size_t len = wild - keyexpr._suffix;
                 suffix = z_malloc(len + 1);
-                memcpy(suffix, keyexpr._suffix, len);
-                suffix[len] = 0;
-                keyexpr._suffix = suffix;
-                _z_keyexpr_set_owns_suffix(&keyexpr, false);
+                if (suffix != NULL) {
+                    memcpy(suffix, keyexpr._suffix, len);
+                    suffix[len] = 0;
+                    keyexpr._suffix = suffix;
+                    _z_keyexpr_set_owns_suffix(&keyexpr, false);
+                }
             }
             uint16_t id = _z_declare_resource(zs._val, keyexpr);
             key = _z_rid_with_suffix(id, wild);
--- a/src/api/api.c
+++ b/src/api/api.c
@@ -822,6 +822,7 @@ z_owned_subscriber_t z_declare_subscriber(z_session_t zs, z_keyexpr_t keyexpr, z
         _z_resource_t *r = _z_get_resource_by_key(zs._val, &keyexpr);
         if (r == NULL) {
             char *wild = strpbrk(keyexpr._suffix, "*$");
+            _Bool do_keydecl = true;
             if (wild != NULL && wild != keyexpr._suffix) {
                 wild -= 1;
                 size_t len = wild - keyexpr._suffix;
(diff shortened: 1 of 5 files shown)


cguimaraes (commented):
From a quick look seems good.
But I suggest you to run valgrind on the z_client_test at least.

hwoithe:
Although not necessary to fix the memory leak issue, this is one of the few places where the result of `z_malloc` is not checked.

Thank you very much for the fix.

src/api/api.c
```diff
@@ -824,11 +825,11 @@ z_owned_subscriber_t z_declare_subscriber(z_session_t zs, z_keyexpr_t keyexpr, z
             if (wild != NULL && wild != keyexpr._suffix) {
                 wild -= 1;
                 size_t len = wild - keyexpr._suffix;
-                char *suffix = z_malloc(len + 1);
+                suffix = z_malloc(len + 1);
```
cguimaraes, line 828: You can make it more robust by checking the allocation as suggested by @hwoithe

cguimaraes (approved):
LGTM. But have a look at the inline comment.

src/api/api.c
```diff
@@ -826,10 +826,12 @@ z_owned_subscriber_t z_declare_subscriber(z_session_t zs, z_keyexpr_t keyexpr, z
...
-                keyexpr._suffix = suffix;
-                _z_keyexpr_set_owns_suffix(&keyexpr, false);
+                if (suffix != NULL) {
+                    memcpy(suffix, keyexpr._suffix, len);
+                    suffix[len] = 0;
+                    keyexpr._suffix = suffix;
+                    _z_keyexpr_set_owns_suffix(&keyexpr, false);
+                }
             }
             uint16_t id = _z_declare_resource(zs._val, keyexpr);
```
cguimaraes, line 836: What is declared if suffix allocation fails? It seems that it will declare a `keyexpr` with id but no suffix...

cguimaraes (changes requested):

src/api/api.c
p-avital: It will declare the exact keyexpr that was passed to the subscriber, but will keep the wild suffix, so the bug is not the one you expected, but it is buggy

p-avital: Actually was a bug, but not the one you expected. Let's just cancel attempting to declare the resource if we failed to allocate for the suffix, the other steps would probably also fail anyway.

cguimaraes (approved):

PR #268, opened 2023-10-24
Add modularity checks in CI

jean-roland:
As the number of possible compile configuration evolve in 2^n, with n the number of configuration tokens, it will quickly become impossible to test them all manually. The goal with this automated check in CI is to:
* Verify that all configurations compile, at least for linux.
* Run a simple program along a zenoh router to test basic functionality


jean-roland:
The thing is examples z_sub, z_query, and z_queryable don't return by themselves, they need to receive a 'q' through scanf

p-avital:
In zenoh-python, we have a python script that spawns the things in parallel, writes to `stdin`, and checks outputs

cguimaraes:
I agree with @Mallets on this. Either on the examples or the `z_client_test`.

Also, if I recall correctly, e.g. the `queryable` example shall not compile if the feature is disable (being explicitly on misconfiguration at compile time instead of runtime). But all the others examples shall compile as work as expected, including the existing tests.

Otherwise, @p-avital approach seems the way to go.

cguimaraes (commented):
I agree with @Mallets on this. Either on the examples or the `z_client_test`.

Also, if I recall correctly, e.g. the `queryable` example shall not compile if the feature is disable (being explicitly on misconfiguration at compile time instead of runtime). But all the others examples shall compile as work as expected, including the existing tests.

Otherwise, @p-avital approach seems the way to go.

jean-roland:
Ok so I added a Python script to run the examples, write to their "stdin" pipes and check their outputs according to the compile-time feature configurations value. 

A bit of issues translating from my machine to the CI but it should be good now.
Also modified the recently PR'ed freertos examples with the configuration options.

p-avital (approved):

Mallets (approved):

PR #275, opened 2023-11-07
removed using namespace std

milyin:
Headers should not leak global "using namespace" statements. This causes unexpected issues in the code which include such headers. See explanation here for example https://stackoverflow.com/questions/5849457/using-namespace-in-c-headers


milyin:
See https://github.com/eclipse-zenoh/zenoh-pico/issues/276

Mallets:
Closing since https://github.com/eclipse-zenoh/zenoh-pico/pull/281 has been merged.

PR #279, opened 2023-11-13
Refactor transport layer

jean-roland:
Transport layer refactor to improve maintainability and modularity in prevision of adding a new transport.

Instead of compile-time specific code removal spread everywhere, files call functions that are swapped to dummy versions when the feature is removed. 

This may increase slightly code size when removing the feature but makes it much easier to expand the number of transports.

+ added a few "early return" pattern here and there

PR #280, opened 2023-11-15
Refactor link capabilities

jean-roland:
Similarly to #279, the goal of this refactor is to improve link capabilities modularity in prevision of adding a new transport.

The bitfield is replaced with a Unicast/Multicast;Stream/Datagram "matrix" enum and a `is_reliable` flag. 
The bitfield macros were removed.

This way the if/else + macro are replaced by a switch case that can be expanded easily.


include/zenoh-pico/link/link.h
```diff
@@ -46,22 +46,18 @@
...
  */
 typedef enum {
-    Z_LINK_CAPABILITY_NONE = 0x00,      // 0
-    Z_LINK_CAPABILITY_RELIABLE = 0x01,  // 1 << 0
-    Z_LINK_CAPABILITY_STREAMED = 0x02,  // 1 << 1
-    Z_LINK_CAPABILITY_MULTICAST = 0x04  // 1 << 2
+    Z_LINK_CAP_UNICAST_STREAM = 0,
+    Z_LINK_CAP_UNICAST_DATAGRAM,
+    Z_LINK_CAP_MULTICAST_STREAM,
+    Z_LINK_CAP_MULTICAST_DATAGRAM,
```
cguimaraes, line 58: I would opt to always have the enumeration values set, as a best practice.
We had some misalignments in the past because of this, due to changing the order which reassigned different values.

include/zenoh-pico/link/link.h
```diff
@@ -105,6 +101,7 @@ typedef struct _z_link_t {
 
     uint16_t _mtu;
     uint8_t _capabilities;
+    bool _is_reliable;
```
cguimaraes, line 104: `_is_reliable` should be a bit in the `_capabilities`.

src/link/link.c
```diff
@@ -148,7 +148,19 @@ size_t _z_link_recv_exact_zbuf(const _z_link_t *link, _z_zbuf_t *zbf, size_t len
 
 int8_t _z_link_send_wbuf(const _z_link_t *link, const _z_wbuf_t *wbf) {
     int8_t ret = _Z_RES_OK;
-    _Bool link_is_streamed = _Z_LINK_IS_STREAMED(link->_capabilities);
+    _Bool link_is_streamed = false;
+
+    switch (link->_capabilities) {
```
cguimaraes, line 153: This should be a bitwise operation as defined in the existing `_Z_LINK_IS_STREAMED`

src/transport/common/rx.c
```diff
@@ -28,32 +28,41 @@ int8_t _z_link_recv_t_msg(_z_transport_message_t *t_msg, const _z_link_t *zl) {
...
     _z_zbuf_reset(&zbf);
 
-    if (_Z_LINK_IS_STREAMED(zl->_capabilities) == true) {
-        // Read the message length
-        if (_z_link_recv_exact_zbuf(zl, &zbf, _Z_MSG_LEN_ENC_SIZE, NULL) == _Z_MSG_LEN_ENC_SIZE) {
-            size_t len = 0;
-            for (uint8_t i = 0; i < _Z_MSG_LEN_ENC_SIZE; i++) {
-                len |= (size_t)(_z_zbuf_read(&zbf) << (i * (uint8_t)8));
-            }
+    switch (zl->_capabilities) {
```
cguimaraes, line 31: Same as before.

src/transport/common/tx.c
```diff
@@ -27,14 +27,23 @@
...
-void __unsafe_z_prepare_wbuf(_z_wbuf_t *buf, _Bool is_streamed) {
+void __unsafe_z_prepare_wbuf(_z_wbuf_t *buf, uint8_t link_capabilities) {
     _z_wbuf_reset(buf);
 
-    if (is_streamed == true) {
-        for (uint8_t i = 0; i < _Z_MSG_LEN_ENC_SIZE; i++) {
-            _z_wbuf_put(buf, 0, i);
-        }
-        _z_wbuf_set_wpos(buf, _Z_MSG_LEN_ENC_SIZE);
+    switch (link_capabilities) {
```
cguimaraes, line 33: same as before

src/transport/common/tx.c
```diff
@@ -43,12 +52,22 @@ void __unsafe_z_prepare_wbuf(_z_wbuf_t *buf, _Bool is_streamed) {
  * Make sure that the following mutexes are locked before calling this function:
  *  - ztu->mutex_tx
  */
-void __unsafe_z_finalize_wbuf(_z_wbuf_t *buf, _Bool is_streamed) {
-    if (is_streamed == true) {
-        size_t len = _z_wbuf_len(buf) - _Z_MSG_LEN_ENC_SIZE;
-        for (uint8_t i = 0; i < _Z_MSG_LEN_ENC_SIZE; i++) {
-            _z_wbuf_put(buf, (uint8_t)((len >> (uint8_t)8 * i) & (uint8_t)0xFF), i);
+void __unsafe_z_finalize_wbuf(_z_wbuf_t *buf, uint8_t link_capabilities) {
+    switch (link_capabilities) {
```
cguimaraes, line 56: The same as before.
Also in all the occurrences afterwards.

cguimaraes (changes requested):

include/zenoh-pico/link/link.h
jean-roland: _capabilities is no longer a bitfield though, eventually capabilities could be a structure including an enum and the reliable flag

cguimaraes: The idea of the bitfield was to reduce the overhead. These values will be booleans, so they can be store in a single bit.
If this was discussed within the team, you can ignore. Otherwise, I suggest to be rethought because this was done intentionally.

src/link/link.c
cguimaraes: Future transports will support or not a given set of functionalities. And the way, the core code should be handled is accordingly to the support or not these functionalities.
As far as I remember, this is how the Rust implementation is also tackling this issue.

Can you give an example of a future transport that does not map? Note that if new properties must be added, they can be done is a clean way with the current approach.

include/zenoh-pico/link/link.h
cguimaraes: tagging @p-avital

src/link/link.c
cguimaraes: tagging @Mallets

include/zenoh-pico/link/link.h
jean-roland: Also invoking @Mallets

include/zenoh-pico/link/link.h
```diff
@@ -46,22 +46,18 @@
...
+ *     Z_LINK_CAP_MULTICAST_DATAGRAM: Link has multicast datagram capabilities.
  */
 typedef enum {
-    Z_LINK_CAPABILITY_NONE = 0x00,      // 0
-    Z_LINK_CAPABILITY_RELIABLE = 0x01,  // 1 << 0
-    Z_LINK_CAPABILITY_STREAMED = 0x02,  // 1 << 1
-    Z_LINK_CAPABILITY_MULTICAST = 0x04  // 1 << 2
+    Z_LINK_CAP_UNICAST_STREAM = 0,
+    Z_LINK_CAP_UNICAST_DATAGRAM = 1,
+    Z_LINK_CAP_MULTICAST_STREAM = 2,
```
Mallets, line 57: I think this does not make much sense... As far as I know there are no `multicast` protocols that are byte-streamed.

include/zenoh-pico/link/link.h
Mallets: In this specific case I don't think it will make any difference. 
`capabilities` is `uint8_t` and `bool` takes the same space of a `uint8_t`. 
Unless you are running on a 8-bit platform, and because of alignment, the two solutions are equivalent w.r.t. space usage. 

The bitfield usually is used to express the combination of different independent capabilities, which may be invalid. Looking at the code re-organisation I believe that having specific enum variants instead of bitfields will make the logic more explicit.

src/link/link.c
Mallets: In accordance with my previous comment, bitwise operations have the tendance to be interpreted as a combination of indipendent parameters. This has led to some confusion and misconception on how things should be done and structured. I believe in this case having a clear variant separation helps in understanding what are the real cardinality of available options.

include/zenoh-pico/link/link.h
cguimaraes: As this is zenoh-pico, I see 8-bit as a target.
Also, you are just considering one capability, so you are not being fair in the analysis.

Today, we are considering 3 capabilities: `_Z_LINK_IS_STREAMED`, `_Z_LINK_IS_RELIABLE`, `_Z_LINK_IS_MULTICAST`.
Tomorrow we may have others capabilities like `_Z_LINK_IS_SECURE` or `_Z_LINK_IS_UNIDIRECTIONAL`.

I am still in favor of the bitfield, also because it scales better than to enumerate all the possible combinations.

cguimaraes: Let me know what is your decision, so I can remove or keep the requested changes in my previous code review.

src/link/link.c
cguimaraes: These parameters are and should be interpreted independently.

If you notice on your own review, this new approach has in fact introduced confusion and misconceptions. @jean-roland was considering `Z_LINK_CAP_MULTICAST_STREAM` as a possible capability that, in your own words, does not make much sense.

cguimaraes: Let me know what is your decision, so I can remove or keep the requested changes in my previous code review.

Mallets: That's because bitfields are interpreted as combinatory parameters and the enum was defined from that combination of bits. As a result, the invalid variant was added.

cguimaraes: The usage here is intended to be a flag structure.

Moreover, if the scope of the refactor is to ease future additions, I find the switch case approach counter-intuitive since it will require to modify more parts that were supposed to be isolated for a given capability.

I would ask @p-avital opinion here as well.

src/link/multicast/bt.c
```diff
@@ -115,7 +115,8 @@ uint16_t _z_get_link_mtu_bt(void) { return SPP_MAXIMUM_PAYLOAD; }
 int8_t _z_new_link_bt(_z_link_t *zl, _z_endpoint_t endpoint) {
     int8_t ret = _Z_RES_OK;
 
-    zl->_capabilities = Z_LINK_CAPABILITY_STREAMED | Z_LINK_CAPABILITY_MULTICAST;
```
jean-roland, line 118: @Mallets @cguimaraes Here

src/link/link.c
jean-roland: I added MULTICAST_STREAM only because bluetooth was declared multicast and streamed.

include/zenoh-pico/link/link.h
jean-roland: Just to weigh in on the discussion, the current bitfield structure isn't super well suited because only RELIABLE is a true flag/bool, for multicast and streamed it really represents enums:
```
enum LINK_CAPABILITY_TRANSPORT {
    UNICAST,
    MILTICAST,
    <New values>,
}
```
and
```
enum LINK_CAPABILITY_FLOW (just made this up) {
    DATAGRAM,
    STREAM,
}
```
Now there might be other flags in the future as you stated but in that case, if we really want to represent things correctly and use as little memory as possible then we have the possibility to do a register like structure:
```
struct capability {
    uint8_t transport: 2;
    uint8_t flow: 1;
    uint8_t reliable: 1;
    uin8_t reserved: 4;
}
```
It does open a lot of problems and future headache if we have to change the fields but that's a possibility.

Another one is to have a bitfield and a combined enum which is almost this PR but the bitfield is only the reliable flag.

p-avital: I do like the register-like structure proposal a bit better than the enum one. Not extremely familiar with it, but then couldn't we do
```c
struct capability {
    enum LINK_CAPABILITY_TRANSPORT transport: 2;
    enum LINK_CAPABILITY_FLOW flow: 1;
    _Bool reliable: 1;
    uin8_t reserved: 4;
}
```
to make each field unambiguous as to its purpose? (since C enums are dumb, this might backfire and make `capability` int-sized though)

Mallets: My main goal is to avoid ambiguity as well as avoiding the possibility to represent invalid states. For that reason, I believe the current PR is a good tradeoff, as mentioned by @jean-roland . 

I don't expect the enum to significantly grow and have a large number of variants. We are targeting fundamental characteristics of transports and links, which have an impact on the Zenoh protocol itself...

src/link/link.c
Mallets: Then I withdraw my comment: ``As far as I know there are no multicast protocols that are byte-streamed.`` 
Now I know :)

p-avital: My take on this is that regardless of the existence of a multicast-streamed transport, I think we always treat multicast-ness and streamed-ness independently (since one affects routing and the other serialization), so while the enum might highlight that some don't exist, is that really a relevant fact we need to encode in here?

In general, I take an enum going through the combinations of X and Y as `X_Y` to be a sign that these properties are rather independent, angling toward the bitfield. But bitfields are a bit ugly: I like the register-struct proposal @jean-roland made in the other thread

Mallets: Then I propose to move forward with @jean-roland proposal then.

jean-roland:
Changes done, tried to declare the fields as enum types and `_Bool` but got a 4 bytes structure, so had to declare them all as `uint8_t`

Mallets (approved):

p-avital (approved):

PR #281, opened 2023-11-17
Remove 'using namespace std' statement. Fixes #276

### LINKED ISSUE #276: remove "using namespace std" statement [Bug]
It's incorrect to expose "using namespace std" in header here:
https://github.com/eclipse-zenoh/zenoh-pico/blob/master/include/zenoh-pico/collections/pointer.h#L29
This affects compilation of other projects which may purposedly avoid including ``std`` namepace.See explanation here 
https://stackoverflow.com/questions/5849457/using-namespace-in-c-headers

Unfortunately just removing this line is not enough because macro ``_Z_POINTER_DEFINE`` accesses ``memory_order_relaxed`` without ``std`` prefix and we can't just add ``std::`` to it as this should work with C code too. A bit more clever solution is required.

cguimaraes:
Fixes #276

### DIFF
--- a/include/zenoh-pico/collections/pointer.h
+++ b/include/zenoh-pico/collections/pointer.h
@@ -23,15 +23,21 @@
 #ifndef __cplusplus
 #include <stdatomic.h>
 #define z_atomic(X) _Atomic X
+#define _z_atomic_store_explicit atomic_store_explicit
+#define _z_atomic_fetch_add_explicit atomic_fetch_add_explicit
+#define _z_atomic_fetch_sub_explicit atomic_fetch_sub_explicit
+#define _z_memory_order_acquire memory_order_acquire
+#define _z_memory_order_release memory_order_release
+#define _z_memory_order_relaxed memory_order_relaxed
 #else
 #include <atomic>
 #define z_atomic(X) std::atomic<X>
-#define atomic_store_explicit std::atomic_store_explicit
-#define atomic_fetch_add_explicit std::atomic_fetch_add_explicit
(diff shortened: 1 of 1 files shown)


Mallets (approved):

Mallets:
LGTM but I'd really like a review from @milyin

cguimaraes:
I forgot to mention that the compilation as backend for zenoh-c also passes.

milyin:
@cguimaraes this fix still doesn't work beacuse it leaks out defines 
```C
#define atomic_store_explicit std::atomic_store_explicit
#define atomic_fetch_add_explicit std::atomic_fetch_add_explicit
#define atomic_fetch_sub_explicit std::atomic_fetch_sub_explicit
#define memory_order_acquire std::memory_order_acquire
#define memory_order_release std::memory_order_release
#define memory_order_relaxed std::memory_order_relaxed
```
so the following code fails to compile with zenoh-pico:
```C++
    std::memory_order order = std::memory_order_relaxed;
```
with error:
```
[build] /home/milyin/ZS/zenoh-cpp/examples/universal/z_get.cxx:28:31: error: no member named 'std' in namespace 'std'; did you mean simply 'std'?
[build]     std::memory_order order = std::memory_order_relaxed;
```

milyin (changes requested):
the defines with names same as functions of std library still leaks to global namespace

milyin:
Looks good now, thank you!

cguimaraes:
@Mallets @p-avital this is ready to merge. Can you take care of it please?

p-avital (approved):

PR #283, opened 2023-11-27
Add raweth transport

jean-roland:
We had a request to send data on an Ethernet link with VLAN support on multicast addresses. This should have been a multicast transport on a Ethernet link but we needed to map each keyexpr to a multicast address, from a specific config file.

This is not possible to do from the link layer since links don't deal with keyexpr, so we went the route of an ad hoc multicast transport that only deal with raw Ethernet frames and VLAN.

This PR only include the transport implementation, not the code binding the transport to the rest of Zenoh to focus the review on the transport itself.

Here is what's included:
* system: open, write and send on raw sockets (UNIX only)
* link:  check the locator which is only the root "reth", everything else is given by the config
* config: a ad hoc config file with the keyexpr mapping information
* transport: a multicast transport that encodes Ethernet and VLAN headers from the config data.


include/zenoh-pico/transport/multicast/tx.h
jean-roland, line 1: Typo piggybacking

include/zenoh-pico/transport/unicast/tx.h
jean-roland, line 1: Typo piggybacking

src/system/freertos_plus_tcp/network.c
jean-roland, line 1: Piggybacking some link errors

src/transport/multicast/tx.c
jean-roland, line 1: Typo piggybacking

include/zenoh-pico/system/link/raweth.h
```diff
@@ -0,0 +1,71 @@
...
+
+// Address Sizes
+#define _ZP_MAC_ADDR_LENGTH 6
+
+// Max frame size
+#define _ZP_MAX_ETH_FRAME_SIZE 1500
+
+typedef struct {
+    uint16_t tpid;      // Tag ethertype
+    uint16_t pcp : 3;   // Priority code point
```
Mallets, line 37: I wonder about the portability of this code. As far as I know, bitfields are not guaranteed to be portable (see [here](https://stackoverflow.com/questions/1490092/c-c-force-bit-field-order-and-alignment)) between different platforms, endianness, and compilers. Should we just use `uint16_t` and do some bitmasking?

include/zenoh-pico/transport/raweth/config.h
```diff
@@ -0,0 +1,44 @@
...
+#include "zenoh-pico/protocol/core.h"
+#include "zenoh-pico/system/link/raweth.h"
+#include "zenoh-pico/transport/transport.h"
+#include "zenoh-pico/utils/result.h"
+
+#if Z_FEATURE_RAWETH_TRANSPORT == 1
+
+typedef struct {
+    _z_keyexpr_t _keyexpr;
+    uint16_t _vlan;  // vlan tag (pcp + dei + id)
```
Mallets, line 29: Previously you defined `_zp_vlan_tag_t` type. Should we use the same type here? Or have `_zp_vlan_tag_t` simply a `uint16_t`?

include/zenoh-pico/system/link/raweth.h
```diff
@@ -0,0 +1,71 @@
...
+
+#include <stdint.h>
+
+#include "zenoh-pico/collections/string.h"
+#include "zenoh-pico/system/platform.h"
+
+#if Z_FEATURE_RAWETH_TRANSPORT == 1
+
+// Ethernet types
+#define _ZP_ETH_TYPE_VLAN 0x8100   // 2048  (0x0800) IPv4
```
Mallets, line 26: I believe the `EthType` should be configurable as part of the raweth configuration.

include/zenoh-pico/system/link/raweth.h
jean-roland: Yeah portability isn't guaranteed in bitfields, now since we don't want to send this structure, the problem mainly resides in platform-dependent behaviors and bugs.

Bitmasking would be more portable but introduce more clutter and possibility for human mistakes.

jean-roland: Now the other thing is, we don't really need to interact with the vlan tag we just need to write it as specified in a config (or locator in the future) so we might as well remove this structure entirely.

include/zenoh-pico/system/link/raweth.h
jean-roland: Not sure what you mean, the VLAN type is fixed and I doubt we'll need double tagging.

For the frame, we'll set the field as the data length so there is no eth type.

cguimaraes (commented):
Why not to consider each `raweth` as a new `Zenoh link`, and each `keyexpr` mapping to a multicast ethernet address to be the instantiation of a new link?

It seems to me that it would imply less code replication and less cross-layer dependencies between code components.

include/zenoh-pico/system/link/raweth.h
Mallets: So are you suggesting to go with a 802.3 header instead of a Ethernet II header?

cguimaraes:
@jean-roland that in indeed a current limitation. But, eventually, we need to extend it to support `1:1:n` or even `1:n(2):n`. The overhead will happen but only if more than one link or transport is used. But well, there are no free meals :-)

It seems that now would be the time to support multiple links given the constraints you are having. I would like to avoid shortcuts now that could become headaches in the future.

Also, in Zenoh there are only two types of transports: `unicast` and `multicast`. It seems that you are implicitly adding a new type, and I am not sure how these changes are planned to be propagated to the rest of Zenoh suite.

Mallets:
> @jean-roland that in indeed a current limitation. But, eventually, we need to extend it to support `1:1:n` or even `1:n(2):n`. The overhead will happen but only if more than one link or transport is used. But well, there are no free meals :-)
> 
> It seems that now would be the time to support multiple links given the constraints you are having. I would like to avoid shortcuts now that could become headaches in the future.
> 
> Also, in Zenoh there are only two types of transports: `unicast` and `multicast`. It seems that you are implicitly adding a new type, and I am not sure how these changes are planned to be propagated to the rest of Zenoh suite.

The main goals of the `raweth` transport are:
- working directly on raw ethernet sockets
- mapping different key expressions to different destination MAC addresses and VLAN IDs for transmission
- being able to receive different key expressions from different destination MAC addresses and VLAN IDs
- the sets of MAC addresses used for transmission and reception may be disjoint
- one single thread for receiving data

Based on the above, I believe in this particular case moving to `1:1:n` support will make things more complex:
- there might be links used only for TX and some others only for RX
- thread management may become more complex since we might need to handle different links separately
- since we only have one raw socket, we might need to do some internal dispatching to preserve the link abstraction
- mapping and ensuring correct src/dst MAC address would be an issue since the sets of MAC addresses used for TX and RX may be disjoint (we need to break somehow the abstraction here)
- memory footprint would increase since several *links objects* would need to be created runtime while with a single transport/link a static configuration could be used

include/zenoh-pico/system/link/raweth.h
Mallets: If it is not needed, then I suggest to remove it.

src/transport/raweth/rx.c
```diff
@@ -0,0 +1,365 @@
...
+
+        if (memcmp(val->_remote_addr.start, remote_addr->start, remote_addr->len) == 0) {
+            ret = val;
+        }
+    }
+    return ret;
+}
+
+/*------------------ Reception helper ------------------*/
+int8_t _z_raweth_recv_t_msg_na(_z_transport_multicast_t *ztm, _z_transport_message_t *t_msg, _z_bytes_t *addr) {
```
Mallets, line 75: Regarding code replication, I believe we could move this code in some common structure between multicast and raweth? I believe the really required function is `_z_raweth_link_recv_zbuf` that should perform the right filtering and checking.

src/system/unix/link/raweth.c
```diff
@@ -0,0 +1,111 @@
...
+    }
+    return (size_t)wb;
+}
+
+size_t _z_receive_raweth(const _z_sys_net_socket_t *sock, void *buff, size_t buff_len, _z_bytes_t *addr) {
+    size_t bytesRead = recvfrom(sock->_fd, buff, buff_len, 0, NULL, NULL);
+    if (bytesRead < 0) {
+        return SIZE_MAX;
+    }
+    // Soft Filtering ?
```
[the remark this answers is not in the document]
jean-roland: Something like that yes

jean-roland:
@Mallets: Some update on the PR:
* Use a configurable ethertype for the frame header
* Remove the vlan struct as we don't need it
* Remove code duplication, especially the multicast rx fsm

src/system/unix/link/raweth.c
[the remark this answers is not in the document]
Mallets: Ok, that should be done according to the provided configuration.

src/transport/raweth/config.c
```diff
@@ -0,0 +1,35 @@
...
+const uint16_t _ZP_RAWETH_CFG_ETHTYPE = 0x72e0;
+
+// Should be generated
+const uint8_t _ZP_RAWETH_CFG_SMAC[_ZP_MAC_ADDR_LENGTH] = {0x30, 0x03, 0xc8, 0x37, 0x25, 0xa1};
+
+// Should be generated
+const _zp_raweth_cfg_entry _ZP_RAWETH_CFG_ARRAY[] = {
+    {{0, {0}, ""}, 0x00, {0xaa, 0xbb, 0xcc, 0xdd, 0xee, 0xff}, false},                // Default mac addr
+    {{0, {0}, "some/key/expr"}, 0x8c, {0x00, 0x11, 0x22, 0x33, 0x44, 0x55}, true},    // entry1
+    {{0, {0}, "another/keyexpr"}, 0x43, {0x01, 0x23, 0x45, 0x67, 0x89, 0xab}, true},  // entry2
```
Mallets, line 29: The config should also include a list of allowed MAC addresses in RX.

src/transport/raweth/lease.c
```diff
@@ -0,0 +1,208 @@
+//
```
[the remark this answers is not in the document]
jean-roland: For lease/read/join the only differences is the send_t function if I'm not mistaken, so there certainly is a way to mutualize the code.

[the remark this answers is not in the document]
Mallets: I see. In the meanwhile I'd like to the `raweth` being validated. My proposal is the following:
- We keep this PR open (and not merging it yet)
- You can work on an additional PR (based on this one - or directly on this one) that implements the actual logic and code glue
- Once `raweth` is validated, we can think about merging into master

jean-roland:
Alright so moving from the draft to testing phase some modifications needed to be made on the tx side:
* Added an interface string in the config to bind the socket, uses a linux specific structure
* Throw an error if activating the feature not on linux (should revisit this later to support more OS)
* Added missing vlan ether type in the header
* Fix some issues with the config parsing

PR #294, opened 2023-12-11
Add raweth data length header field

jean-roland:
Because ethernet frame minimum size is 64 bytes, some device will pad frames. This forces us to add a field containing our actual payload length.

### DIFF
--- a/include/zenoh-pico/protocol/definitions/transport.h
+++ b/include/zenoh-pico/protocol/definitions/transport.h
@@ -477,6 +477,8 @@ typedef struct {
 } _z_t_msg_fragment_t;
 void _z_t_msg_fragment_clear(_z_t_msg_fragment_t *msg);
 
+#define _Z_FRAGMENT_HEADER_SIZE 12
+
 /*------------------ Transport Message ------------------*/
 typedef union {
     _z_t_msg_join_t _join;
--- a/include/zenoh-pico/system/link/raweth.h
+++ b/include/zenoh-pico/system/link/raweth.h
@@ -29,7 +29,7 @@
 #define _ZP_MAC_ADDR_LENGTH 6
 
 // Max frame size
-#define _ZP_MAX_ETH_FRAME_SIZE 1500
+#define _ZP_MAX_ETH_FRAME_SIZE 1514
 
 // Ethernet header structure type
 typedef struct {
--- a/include/zenoh-pico/system/link/raweth.h
+++ b/include/zenoh-pico/system/link/raweth.h
@@ -36,6 +36,7 @@ typedef struct {
     uint8_t dmac[_ZP_MAC_ADDR_LENGTH];  // Destination mac address
     uint8_t smac[_ZP_MAC_ADDR_LENGTH];  // Source mac address
     uint16_t ethtype;                   // Ethertype of frame
+    uint16_t data_length;               // Payload length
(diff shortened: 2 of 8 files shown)


include/zenoh-pico/protocol/definitions/transport.h
```diff
@@ -477,6 +477,8 @@ typedef struct {
 } _z_t_msg_fragment_t;
 void _z_t_msg_fragment_clear(_z_t_msg_fragment_t *msg);
 
+#define _Z_FRAGMENT_HEADER_SIZE 12
```
Mallets, line 480: This is actually not true, the fragment header size may vary depending on the used protocol extension.
As a consequence, basing the code logic on this value may lead to some bugs.

include/zenoh-pico/system/link/raweth.h
```diff
@@ -44,6 +45,7 @@ typedef struct {
     uint16_t vlan_type;                 // Vlan ethtype
     uint16_t tag;                       // Vlan tag
     uint16_t ethtype;                   // Ethertype of frame
```
Mallets, line 47: Do we need `ethtype` then? As far as I understand it, `data_legnth` replaces `ethtype`.

src/transport/multicast/tx.c
```diff
@@ -116,7 +116,7 @@ int8_t _z_multicast_send_n_msg(_z_session_t *zn, const _z_network_message_t *n_m
             } else {
                 // The message does not fit in the current batch, let's fragment it
                 // Create an expandable wbuf for fragmentation
-                _z_wbuf_t fbf = _z_wbuf_make(ztm->_wbuf._capacity - 12, true);
+                _z_wbuf_t fbf = _z_wbuf_make(ztm->_wbuf._capacity - _Z_FRAGMENT_HEADER_SIZE, true);
```
Mallets, line 119: The fragmentation buffer should allocate a small amount of memory and it shouldn't be dependent on the fragment header size.

src/transport/unicast/tx.c
```diff
@@ -125,7 +125,7 @@ int8_t _z_unicast_send_n_msg(_z_session_t *zn, const _z_network_message_t *n_msg
             } else {
                 // The message does not fit in the current batch, let's fragment it
                 // Create an expandable wbuf for fragmentation
-                _z_wbuf_t fbf = _z_wbuf_make(ztu->_wbuf._capacity - 12, true);
+                _z_wbuf_t fbf = _z_wbuf_make(ztu->_wbuf._capacity - _Z_FRAGMENT_HEADER_SIZE, true);
```
Mallets, line 128: See my comment on the fragment header size above.

include/zenoh-pico/protocol/definitions/transport.h
jean-roland: I'd say we should fix this in another PR, I opened an issue on the overall issue: #295

include/zenoh-pico/system/link/raweth.h
jean-roland: 802.3 raw requires us to use a 802.2 LLC header which would be 3 bytes of useless information for us, so I figured an ethertype and a frame length was the most efficient choice.

Mallets (approved):

PR #296, opened 2023-12-14
Update z_api_alignment_test for compatiblity with zenoh-c build

sashacmc:


tests/z_api_alignment_test.c
```diff
@@ -215,9 +221,9 @@ int main(int argc, char **argv) {
     assert(hellos >= 1);
 
     uint32_t _scouting_timeout = strtoul(SCOUTING_TIMEOUT, NULL, 10);
-    z_sleep_ms(_scouting_timeout);
+    sleep(_scouting_timeout / 1000);
```
Mallets, line 224: In Zenoh-Pico/Zenoh-C we should better use `z_sleep_*` and avoid using platform-specific syscalls.
However, those functions are still missing in Zenoh-C (see https://github.com/eclipse-zenoh/zenoh-c/issues/183).
Nevertheless, they are present in Zenoh-Pico, so better keep them.

sashacmc: But as far as I understand, the idea is for the files to be identical in both repos, do you think adding a `#define` here would be reasonable?

cguimaraes: I would say that the additional functions shall be added in zenoh-c, as referred by @Mallets.

tests/z_api_alignment_test.c
```diff
@@ -26,7 +26,11 @@
 #include <unistd.h>
 #endif
 
+#ifdef ZENOH_PICO
 #include "zenoh-pico.h"
```
cguimaraes, line 30: Tests are also failing because the `ZENOH_PICO` checked before its definition in `zenoh-pico.h` include

cguimaraes (changes requested):

tests/z_api_alignment_test.c
sashacmc: Yes, sure, but until we have no them in zenohc, api alignment test on them has no sense.

sashacmc:
Close because seems currently on zenoh-c side this test require a lot of rework.

Issue #291, opened 2023-12-06
[Bug] Failure to receive large samples


gmartin82:
### Describe the bug

In client mode, I have encountered an issue with receiving samples >= 32 KB. Samples appear to be published successfully to a Zenoh Router but are not received by a subscribing application. The exact size which triggers the issue seems to vary from application to application.

### To reproduce

Start a Zenoh router:

```./target/release/zenohd -l tcp/127.0.0.1:7447 --no-multicast-scouting```

Start a Zenoh Pico subscriber:

```./build/examples/z_sub -e tcp/127.0.0.1:7447```

Start a Zenoh Pico publisher with large, randomly generated character data:

```
rand=$(cat /dev/urandom | tr -dc 'a-zA-Z0-9' | fold -w 65536 | head -n 1)
./build/examples/z_pub -e tcp/127.0.0.1:7447 -v $rand
```

Observe that samples are not received by ```z_sub```.

### System info

Platform: Ubuntu 22.04 64-bit
CPU: Intel i7-1260P
Zenoh Pico version: master (https://github.com/eclipse-zenoh/zenoh-pico/commit/ae3c2376709c6e90875d49197935dbc4d5ba2105)

jean-roland:
I confirm, this looks like a regression from Zenoh starting from hop-to-hop compression.

Mallets:
The size of the RX buffer should depend on wether the link is streamed or not. In case it is streamed, we need to account also for the 2 bytes indicating the batch length. The compression PR fixed that behaviour in Zenoh Rust. However, by doing so, it broke fragmentation compatibility with Zenoh-Pico since those 2 bytes are not accounted.

As a temporary fix, I've created this PR: https://github.com/eclipse-zenoh/zenoh/pull/630

Mallets:
I've tested https://github.com/eclipse-zenoh/zenoh-pico/pull/297/commits/479c1737e1a8ed599c6def2d6c99f978af4fd64b s and it works against https://github.com/eclipse-zenoh/zenoh/commit/780ec606ba4000a17b2da0e2b0b668e11d37c65e.
Once https://github.com/eclipse-zenoh/zenoh-pico/pull/297 lands, this issue should be solved. In the meanwhile, I'm waiting to merge https://github.com/eclipse-zenoh/zenoh/pull/630.

Mallets:
#297 has been merged. Before closing this issue, @gmartin82 can you please verify on your side that everything is now working as expected?

[...]

PR #297, opened 2023-12-15
Fix fragmentation

jean-roland:
There are currently a few issues regarding fragmentation messages:
* #291 
* Fragment header buffer allocation size values were weird.
* Fragments were recopying data

This PR is attempting to resolve that and also adds a CI non-regression test for this feature.


src/protocol/codec.c
```diff
@@ -296,8 +296,7 @@ int8_t _z_bytes_val_encode(_z_wbuf_t *wbf, const _z_bytes_t *bs) {
     int8_t ret = _Z_RES_OK;
 
     if ((wbf->_expansion_step != 0) && (bs->len > Z_TSID_LENGTH)) {
-        // ret |= _z_wbuf_wrap_bytes(wbf, bs->start, 0, bs->len);
-        ret |= _z_wbuf_write_bytes(wbf, bs->start, 0, bs->len);
+        ret |= _z_wbuf_wrap_bytes(wbf, bs->start, 0, bs->len);
```
cguimaraes, line 299: It seems that this line is being changed frequently. Last time it was from `wrap` to `write` in https://github.com/eclipse-zenoh/zenoh-pico/commit/5238f256f8ccb4f737e82552fb0241f721b8d691 .

tests/z_test_fragment_rx.c
```diff
@@ -0,0 +1,90 @@
+//
```
cguimaraes, line 1: Why not to integrate the test in the `client` and/or `multicast` tests instead of having a new one? Also, because this shall be tested while interoperating with the Rust implementation.

include/zenoh-pico/protocol/definitions/message.h
```diff
@@ -45,6 +45,8 @@
 #define _Z_FLAG_Z_T 0x20  // 1 << 5 | QueryTarget       if T==1 then the query target is present
 #define _Z_FLAG_Z_X 0x00  // Unused flags are set to zero
 
+#define _Z_FRAG_BUFF_BASE_SIZE 128  // Base size of the buffer to encode a fragment message header
+
 // Flags:
```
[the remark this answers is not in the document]
jean-roland: It's an arbitrary value we decided with @Mallets, that seemed not to big and not to small. The buffer being expandable if needed.

src/protocol/codec.c
jean-roland: Yah that's one of the whole point of this PR, reverting this change from @p-avital that outlived its usefulness.

include/zenoh-pico/protocol/definitions/message.h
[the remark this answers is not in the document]
Mallets: In fragmentation, the `zbuf` is used to copy the header bytes and to append the pointer to the payload so as to have a linear reading of the bytes.
`128` is a *rules of thumb* heuristic about the size of memory you need to serialize the header. If more is needed, another block can be allocated straight after and so on an so forth.

tests/z_test_fragment_rx.c
Mallets: I agree with both, but the fact we don't have a test that actually test fragmentation with the Rust implementation is the at root of https://github.com/eclipse-zenoh/zenoh-pico/issues/291. So, if I have to prioritise, having a test that tests against Rust is more important than on UDP multicast. Keeping in mind that we don't have reliability on UDP multicast.

Mallets (approved):

Mallets:
Approving this PR. @jean-roland I'd like to see in a separate PR the fragmentation tests working as well with the Zenoh router.

PR #304, opened 2023-12-22
Add windows atomic

jean-roland:
Since C11 atomics were only added as an experimental feature in MSVC in February, Windows was excluded from atomics on zenoh-pico... until now.


cguimaraes (approved):
LGTM

Mallets (approved):

PR #307, opened 2023-12-27
Add newline to logs

### LINKED ISSUE #238: Make the logging macros handle newlining
### Describe the feature

As explained by @cguimaraes in #225, logging macros already do some formatting, so they should also do newlines.

jean-roland:
This fixes #238

In addition, it switches to a compiler based macro, to ensure no obsolete log will blow up 2 years after the renaming of a variable because we didn't build with the correct `ZENOH_DEBUG` value.

Also added a compiler symbol to identify a debug/release build.


include/zenoh-pico/utils/logging.h
```diff
@@ -17,55 +17,57 @@
...
-    printf(x, ##__VA_ARGS__);
-#define _Z_ERROR_CONTINUE(x, ...) printf(x, ##__VA_ARGS__);
+#define _Z_DEBUG(...)                          \
+    do {                                       \
+        if (ZENOH_DEBUG >= _Z_LOG_LVL_DEBUG) { \
+            _Z_LOG_PREFIX(DEBUG);              \
+            printf(__VA_ARGS__);               \
+            printf("\n");                      \
+        }                                      \
+    } while (false)
```
cguimaraes, line 52: What is the reason for this `do..while` if the loop only executes once?

include/zenoh-pico/utils/logging.h
```diff
@@ -17,55 +17,57 @@
...
-    printf(x, ##__VA_ARGS__);
-#define _Z_ERROR_CONTINUE(x, ...) printf(x, ##__VA_ARGS__);
+#define _Z_INFO(...)                          \
+    do {                                      \
+        if (ZENOH_DEBUG >= _Z_LOG_LVL_INFO) { \
+            _Z_LOG_PREFIX(INFO);              \
+            printf(__VA_ARGS__);              \
+            printf("\n");                     \
+        }                                     \
+    } while (false)
```
cguimaraes, line 61: What is the reason for this `do..while` if the loop only executes once?

include/zenoh-pico/utils/logging.h
```diff
@@ -17,55 +17,57 @@
...
-#define _Z_ERROR(x, ...) (void)(0)
-#endif
+#define _Z_ERROR(...)                          \
+    do {                                       \
+        if (ZENOH_DEBUG >= _Z_LOG_LVL_ERROR) { \
+            _Z_LOG_PREFIX(ERROR);              \
+            printf(__VA_ARGS__);               \
+            printf("\n");                      \
+        }                                      \
+    } while (false)
```
cguimaraes, line 70: What is the reason for this `do..while` if the loop only executes once?

include/zenoh-pico/utils/logging.h
```diff
@@ -17,55 +17,57 @@
...
-    _Z_LOG_PREFIX(INFO); \
-    printf(x, ##__VA_ARGS__);
-#define _Z_INFO_CONTINUE(x, ...) printf(x, ##__VA_ARGS__);
-#define _Z_ERROR(x, ...)  \
-    _Z_LOG_PREFIX(ERROR); \
-    printf(x, ##__VA_ARGS__);
-#define _Z_ERROR_CONTINUE(x, ...) printf(x, ##__VA_ARGS__);
+#define _Z_DEBUG(...)                          \
+    do {                                       \
+        if (ZENOH_DEBUG >= _Z_LOG_LVL_DEBUG) { \
```
cguimaraes, line 47: Why to do this check at runtime, instead of at compile time as before?

include/zenoh-pico/utils/logging.h
```diff
@@ -17,55 +17,57 @@
...
-#define _Z_DEBUG_CONTINUE(x, ...) (void)(0);
-#define _Z_INFO(x, ...) (void)(0);
-#define _Z_INFO_CONTINUE(x, ...) (void)(0);
-#define _Z_ERROR(x, ...)  \
-    _Z_LOG_PREFIX(ERROR); \
-    printf(x, ##__VA_ARGS__);
-#define _Z_ERROR_CONTINUE(x, ...) printf(x, ##__VA_ARGS__);
+#define _Z_INFO(...)                          \
+    do {                                      \
+        if (ZENOH_DEBUG >= _Z_LOG_LVL_INFO) { \
```
cguimaraes, line 56: Why to do this check at runtime, instead of compile time as before?

include/zenoh-pico/utils/logging.h
```diff
@@ -17,55 +17,57 @@
...
+    } while (false)
 
-#elif (ZENOH_DEBUG == 0)
-#define _Z_DEBUG(x, ...) (void)(0)
-#define _Z_INFO(x, ...) (void)(0)
-#define _Z_ERROR(x, ...) (void)(0)
-#endif
+#define _Z_ERROR(...)                          \
+    do {                                       \
+        if (ZENOH_DEBUG >= _Z_LOG_LVL_ERROR) { \
```
cguimaraes, line 65: Why to do this check at runtime, instead of compile time as before?

cguimaraes (commented):
Some clarification required.

include/zenoh-pico/utils/logging.h
jean-roland: It turns the block into a single statement and forces a `;`after the macro.

include/zenoh-pico/utils/logging.h
jean-roland: Actually it transitions from a pre-compiler thing to a compile time check, as the compiler realises this values won't change for the whole execution of the program, a bit like a C++ `constexpr`. But just to be extra sure I added the `(void)(0)` in release mode and no debug, in case some exotic compiler emits run time checks.

The problem with pre compiler stuff is that it doesn't check if the code is valid while the compiler do. So let's say in my log statement I print a variable, that is then changed (type or name), it will go unnoticed until the compiler sees it, so if I only compile without Debug it won't see it. Now it will always see this kind of issues.

include/zenoh-pico/utils/logging.h
cguimaraes: If MISRA-C is still followed for the codebase, this shall not be required.
But I get the point...

include/zenoh-pico/utils/logging.h
cguimaraes: I am not sure if it is already in place, but having compilation tests for both debug and release seems a simpler option.

Mallets (approved):

PR #308, opened 2024-01-02
Restore single thread compilation

### LINKED ISSUE #254: Should `platform/unix.h` avoid `#include <pthread.h>` if `Z_MULTI_THREAD == 0`?
### Describe the bug

As per title.

IIUC, setting `config.h::Z_MULTI_THREAD` to `0` should result in no use of threading.

I'm not sure whether I understand everything correctly, but if `pthread.h` is only used in case multithreading is enabled, perhaps 

https://github.com/eclipse-zenoh/zenoh-pico/blob/79bf88c6315119bf0236c93b3cd398168d2bc9c6/include/zenoh-pico/system/platform/unix.h#L18

should be changed to:

```c
#if Z_MULTI_THREAD == 1
#include <pthread.h>
#endif
```

? And similarly in other platform headers.

### To reproduce

1. clone `zenoh-pico` on a platform without `pthread.h`
2. edit `config.h` and set `Z_MULTI_THREAD = 0`
3. edit `CMakeLists.txt`, set `CHECK_THREADS` to `OFF`
4. attempt to build

Compilation will fail due to unconditional `#include` described above.

### System info

- platform: POSIX compatible-ish (but no `pthread.h`)
- cpu: `amd64`
- zenoh-pico version: 79bf88c6315119bf0236c93b3cd398168d2bc9c6 (current `HEAD`)

jean-roland:
Included in this PR:
* Restore single thread compilation that was broken due to raweth changes from #294 
* Fix #254 
* Add a single thread config build & test (pub_st & sub_st examples) in CI


src/system/unix/system.c
```diff
@@ -15,6 +15,7 @@
 #include <stddef.h>
 #include <stdio.h>
 #include <stdlib.h>
+#include <time.h>
```
p-avital, line 18: could this be accidental?

p-avital: Actually compensates for `pthread` becoming optional in an included header, which previously transitively included `time`

p-avital (approved):

Issue #271, opened 2023-10-26
Expose public functions for synchronisation


jean-roland:
### Describe the feature

Currently example z_ping use private functions/types:
* _z_task_t
* _z_condvar_t
* _z_mutex_t

For thread synchronisation, it was clearly planned to turn this into public functions but it has not yet been done.

jean-roland:
Of all systems currently supported only 2 don't support them:
* Arduino/opencr has no support for `mutex` and `condvar`
* FreeRTOS has no support for `condvar`

PR #309, opened 2024-01-04
Make all sync mechanisms public

jean-roland:
Some system-abstracted functions where public, like `z_time` and `z_clock` while the multi-thread ones `z_task`, `z_mutex` and `z_condvar` were not. This PR fixes this inconsistency that was noted in #271.

Note that 2 systems are not fully covered (functions and types are stubs):
* `Arduino/opencr` has no support for task, mutex and condvar
* `FreeRTOS` has no support for condvar


p-avital:
If we want them public, can it at least be restricted to `zp` namespace? Otherwise `zenoh-c` will also have to expose these

Mallets:
I agree with @p-avital . Let's expose them public but with the `zp_` prefix.

jean-roland:
But wasn't Zenoh C supposed to eventually expose at least some of these? Also it would be weird to have some of them in `z_` and some in `zp_` so here are all the mechanisms concerned:
* `z_random`
* `z_malloc/z_free`
* `z_task`
* `z_mutex`
* `z_condvar`

Mallets:
> But wasn't Zenoh C supposed to eventually or already expose at least some of these? Also it would be weird to have some of them in `z_` and some in `zp_` so here are all the mechanisms concerned:
> 
> * `z_random`
> * `z_malloc/z_free`
> * `z_task`
> * `z_mutex`
> * `z_condvar`
> * `z_sleep`
> * `z_clock`
> * `z_time`

I think we need a table checking what is supported and what not. The real API I'm quite confident restricting to `zp_` only for now is `z_task` since `zenoh-c` is a binding to `async` Rust. We might expose it in the future though. Comments?

jean-roland:
Agreed for `z_task`, and beside `z_random` that is not used in the examples, exposing the others on zenoh-c should allow the zenoh-pico examples to run on zenoh-c.

include/zenoh-pico/system/platform/arduino/esp32.h
```diff
@@ -24,10 +24,10 @@
 #endif  // Z_FEATURE_MULTI_THREAD == 1
 
 #if Z_FEATURE_MULTI_THREAD == 1
-typedef TaskHandle_t _z_task_t;
-typedef void *_z_task_attr_t;  // Not used in ESP32
-typedef pthread_mutex_t _z_mutex_t;
-typedef pthread_cond_t _z_condvar_t;
+typedef TaskHandle_t z_task_t;
```
Mallets, line 27: I believe the type should be called `zp_task_t`.

include/zenoh-pico/system/platform/arduino/esp32.h
```diff
@@ -24,10 +24,10 @@
...
+typedef void *z_task_attr_t;  // Not used in ESP32
```
Mallets, line 28: I believe the type should be called `zp_task_attr_t`.

include/zenoh-pico/system/platform/void.h
```diff
@@ -18,10 +18,10 @@
 #include "zenoh-pico/config.h"
 
 #if Z_FEATURE_MULTI_THREAD == 1
-typedef void *_z_task_t;
-typedef void *_z_task_attr_t;
-typedef void *_z_mutex_t;
-typedef void *_z_condvar_t;
+typedef void *z_task_t;
+typedef void *z_task_attr_t;
+typedef void *z_mutex_t;
```
Mallets, line 23: It seems that `z_mutex_t` and `z_condvar_t` are not available in `zenoh-c`.

Mallets:
Before merging, can you please confirm that:

`z_malloc/z_free`
`z_mutex`
`z_condvar`
`z_sleep`
`z_clock`
`z_time`

are all supported in `zenoh-c`?

jean-roland:
None of these types are currently supported in zenoh-c

Mallets:
Ok, then let's stick to `zp_xxx` namespace. 
Once those functions are added to `zenoh-c` we can always do a `#define zp_xxx z_xxx` for backward compatibility.

Mallets (approved):

Mallets (approved):

Mallets (approved):

Issue #272, opened 2023-10-26
Align z_time and z_clock primitives for Windows with Rust implementation


cguimaraes:
### Describe the feature

**Is your feature request related to a problem?**
No

**Describe the solution you'd like**
Currently on Windows, `z_time` and `z_clock` primitives are implemented in the same way. However, they have different purposes in mind, as mentioned in #270 .

**Additional context**
From @Mallets in #269: 
> For Windows, in [Rust](https://doc.rust-lang.org/std/time/struct.Instant.html#underlying-system-calls) they base the clock implementation on [QueryPerformanceCounter](https://learn.microsoft.com/en-us/windows/win32/api/profileapi/nf-profileapi-queryperformancecounter). We might think about updating the Windows port to that but I'd leave it as a new PR.

Mallets:
Closed by #310

PR #310, opened 2024-01-04
Switch Windows z_clock to QueryPerformanceCounter

jean-roland:
As stated in #272, the Windows `z_clock` implementation was the same as `z_time` which wasn't ideal.

 This PR aligns `z_clock` with the Rust clock implementation that uses Windows' QueryPerformanceCounter.
 
 As a bonus, it switches the scout timeout loop to `z_clock` instead of `z_time` which is more suited.

### DIFF
--- a/examples/windows/z_ping.c
+++ b/examples/windows/z_ping.c
@@ -113,7 +113,7 @@ int main(int argc, char** argv) {
         results[i] = z_clock_elapsed_us(&measure_start);
     }
     for (unsigned int i = 0; i < args.number_of_pings; i++) {
-        printf("%d bytes: seq=%d rtt=%luµs, lat=%luµs\n", args.size, i, results[i], results[i] / 2);
+        printf("%d bytes: seq=%d rtt=%luus, lat=%luus\n", args.size, i, results[i], results[i] / 2);
     }
     _z_mutex_unlock(&mutex);
     z_free(results);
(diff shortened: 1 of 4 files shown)


examples/windows/z_ping.c
```diff
@@ -113,7 +113,7 @@ int main(int argc, char** argv) {
         results[i] = z_clock_elapsed_us(&measure_start);
     }
     for (unsigned int i = 0; i < args.number_of_pings; i++) {
-        printf("%d bytes: seq=%d rtt=%luµs, lat=%luµs\n", args.size, i, results[i], results[i] / 2);
+        printf("%d bytes: seq=%d rtt=%luus, lat=%luus\n", args.size, i, results[i], results[i] / 2);
```
jean-roland, line 116: Windows shell is really bad at parsing anything else than ASCII

src/session/scout.c
```diff
@@ -55,8 +55,8 @@ _z_hello_list_t *__z_scout_loop(const _z_wbuf_t *wbf, const char *locator, unsig
                 // The receiving buffer
                 _z_zbuf_t zbf = _z_zbuf_make(Z_BATCH_UNICAST_SIZE);
 
-                z_time_t start = z_time_now();
-                while (z_time_elapsed_ms(&start) < period) {
+                z_clock_t start = z_clock_now();
```
jean-roland, line 58: `z_clock_t` is more adapted than `z_time_t` to a timeout implementation like here

Mallets (approved):

PR #313, opened 2024-01-06
add Z_FEATURE_ASSEMBLY_NOP

ffontaine:
Add `Z_FEATURE_ASSEMBLY_NOP` to allow the user to disable assembly `nop` code as it raises the following build failure on some "exotic" architectures such as or1k:

```
/home/autobuild/autobuild/instance-1/output-1/build/libzenoh-pico-0.10.0-rc/src/link/endpoint.c: Assembler messages:
/home/autobuild/autobuild/instance-1/output-1/build/libzenoh-pico-0.10.0-rc/src/link/endpoint.c:358: Error: unrecognized instruction `nop'
```

Fixes:
 - http://autobuild.buildroot.org/results/fd0b2c666a1dc1537162d15b27743abd270243ed

### DIFF
--- a/include/zenoh-pico/config.h
+++ b/include/zenoh-pico/config.h
@@ -225,6 +225,13 @@
 #define Z_FEATURE_RAWETH_TRANSPORT 0
 #endif
 
+/**
+ * Enable assembly nop.
+ */
+#ifndef Z_FEATURE_ASSEMBLY_NOP
+#define Z_FEATURE_ASSEMBLY_NOP 1
+#endif
+
 /*------------------ Compile-time configuration properties ------------------*/
 /**
  * Default length for Zenoh ID. Maximum size is 16 bytes.
--- a/src/link/endpoint.c
+++ b/src/link/endpoint.c
@@ -361,7 +361,9 @@ size_t _z_endpoint_config_strlen(const _z_str_intmap_t *s, const char *proto) {
         if (_z_str_eq(proto, RAWETH_SCHEMA) == true) {
         len = _z_raweth_config_strlen(s);
     } else {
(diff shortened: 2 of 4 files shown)


jean-roland:
Though I understand the problem you encounter and that PR do solve it, I'd prefer trying to get rid of the `nop` in the first place to avoid adding yet another pre-compiler token.

So unless we don't find a way to do this, I vote for postponing this PR.

ffontaine:
I can update the PR to remove all `nop` if you prefer this solution.

jean-roland:
I'll do it, I just need to check with the team if there was no particular reason we had to use the `nop` in the first place.

ffontaine:
Great, thank you. I'll wait for your feedback before sending a patch to buildroot.

p-avital:
As an alternative (since investigation reveals that the `nop`s may be required for certain compilers to not do goofs), we could have something like
```c
#ifndef _Z_ASM_NOP
#define _Z_ASM_NOP _asm_("nop")
#endif
```

This would let us keep the current nops, be less verbose, keep the symbol private, and allow users to define the nop as something else if needed for their platform.

Mallets:
I think having `nop` has some advantages in certain cases, e.g. to help the compiler or CPU. I'd rather keep the `nop` but behind a `#define` as proposed by @p-avital . @jean-roland what do you think?

jean-roland:
That makes sense for the loops but some `nop` could still be removed so I'll do a mix of both in #314

Mallets:
#314 has been merged. @ffontaine can you please check if the problem is fixed for you?

ffontaine:
Seems to work, I'll send a patch to buildroot, thanls

Issue #242, opened 2023-08-25
API inconsistency: need function to check if z_keyexpr_t is Z_RESOURCE_ID_NONE


milyin:
### Describe the feature

The `z_keyexpr_t` have the following definition:
```C
typedef struct {
    _z_zint_t _id;
    const char *_suffix;
} _z_keyexpr_t;
```
The `z_keyexpr_t` instance is considered valid if it is registered in the `Session` object (let's name it **declared**) or if it have the key expression string value stored inside (let's name it **standalone**):
```C
_Bool z_keyexpr_is_initialized(const z_keyexpr_t *keyexpr) {
    _Bool ret = false;
    if ((keyexpr->_id != Z_RESOURCE_ID_NONE) || (keyexpr->_suffix != NULL)) {
        ret = true;
    }
    return ret;
}
```
But all key expression processing functions, like `z_keyexpr_includes`, `z_keyexpr_intersects`, etc. are able to process only **standalone** instances, otherwise they return error:
```C
int8_t z_keyexpr_includes(z_keyexpr_t l, z_keyexpr_t r) {
    int8_t ret = 0;
    if ((l._id == Z_RESOURCE_ID_NONE) && (r._id == Z_RESOURCE_ID_NONE)) {
        ret = zp_keyexpr_includes_null_terminated(l._suffix, r._suffix);
    } else {
        ret = _Z_ERR_GENERIC;
    }
    return ret;
}
```
There is no direct way to determine through API, is the valid key expression **standalone** or **declared**. It's possible to check it through `z_keyexpr_as_bytes`, but this function performs extra work -  creates `z_bytes_t` structure and calls `strlen`.

I think we need API function like this:
```C
_Bool zp_keyexpr_is_declared(const z_keyexpr_t *keyexpr) {
    _Bool ret = false;
    if (keyexpr->_id != Z_RESOURCE_ID_NONE) {
        ret = true;
    }
    return ret;
}
```
If this function returns `TRUE`, this means that the `z_keyexpr_t` instance is valid and declared in `Session` and it's text value can be extracted with `zp_keyexpr_resolve`. Currently there is no way to determine this.

This functionality is necessary to fix issue https://github.com/eclipse-zenoh/zenoh-cpp/issues/20 in zenoh-cpp

milyin:
[...]

PR #315, opened 2024-01-09
Add keyexpr_was_declared api function

jean-roland:
Fixes issue #242

### DIFF
--- a/include/zenoh-pico/api/primitives.h
+++ b/include/zenoh-pico/api/primitives.h
@@ -93,19 +93,20 @@ z_owned_str_t z_keyexpr_to_string(z_keyexpr_t keyexpr);
  *
  * Returns:
  *   The :c:type:`z_bytes_t` pointing to key expression string representation if it's possible
-
  */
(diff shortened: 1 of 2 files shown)


Mallets:
LGTM but I'd like to have the review as well from @milyin .

include/zenoh-pico/api/primitives.h
```diff
@@ -97,6 +97,18 @@ z_owned_str_t z_keyexpr_to_string(z_keyexpr_t keyexpr);
  */
 z_bytes_t z_keyexpr_as_bytes(z_keyexpr_t keyexpr);
 
+/**
+ * Indicates if the key expression has been declared but don't guarantee it's still in session.
```
milyin, line 101: I would add that it's not possible to get key expression string representation if it's declared and pointed to `zp_resolve`. This will make clear why this function is needed and connect this function to two functions above

milyin (changes requested):
Documentation update requested

milyin (approved):

PR #317, opened 2024-01-12
Fix c99 compilation

jean-roland:
C99 compilation was broken because of the usage of c11 generics macros:
* Tests are only built with c11 standard as they use generics.
* c99 examples are patched to avoid using c11 generics.
* A c99 compilation test was added to the CI to avoid this kind of issue going unnoticed in the future.


Mallets (approved):

cguimaraes (commented):
LGTM

PR #318, opened 2024-01-16
Attachment support for pub/sub

p-avital:
Since we need pub/sub attachments earlier, and they're easier to implement, we merge them before enabling attachments in query/replies.

This PR also paves the way for query/replies attachment support, but their API still needs some work to ensure ownership stays clear.


p-avital:
@jean-roland can I have a review? :D

examples/unix/c11/z_put.c
```diff
@@ -14,11 +14,17 @@
...
+#include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <unistd.h>
 #include <zenoh-pico.h>
 
+#include "zenoh-pico/api/macros.h"
+#include "zenoh-pico/api/types.h"
+#include "zenoh-pico/collections/bytes.h"
+#include "zenoh-pico/protocol/core.h"
```
cguimaraes, line 26: We shouldn't make use of the internal API in the examples.

examples/unix/c11/z_sub.c
```diff
@@ -14,17 +14,31 @@
...
 #include <ctype.h>
 #include <stddef.h>
+#include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <unistd.h>
 #include <zenoh-pico.h>
 
+#include "zenoh-pico/api/types.h"
+#include "zenoh-pico/protocol/core.h"
```
cguimaraes, line 24: We shouldn't make use of the internal API in the examples.

jean-roland:
Now this is only the first phase of attachments so that'll do,  but for the final one we'd need a CI test checking that both configuration works and an attachment update on `z_msgcodec_test` and `z_client_test`.

src/api/api.c
```diff
@@ -1106,3 +1162,116 @@ int8_t zp_send_join(z_session_t zs, const zp_send_join_options_t *options) {
...
+        if (ret) {
+            return ret;
+        }
+        current = _z_bytes_pair_list_tail(current);
+    }
+    return 0;
+}
+z_owned_bytes_map_t z_bytes_map_new(void) { return (z_owned_bytes_map_t){._inner = _z_bytes_pair_list_new()}; }
+z_owned_bytes_map_t z_bytes_map_null(void) { return (z_owned_bytes_map_t){._inner = NULL}; }
+z_bytes_t z_bytes_new(const char *str) {
```
Mallets, line 1273: Why taking a `const char *str` as argument for a `z_bytes_t`? I would expect to take a `const uint8_t *bytes`.

src/api/api.c
```diff
@@ -1106,3 +1162,116 @@ int8_t zp_send_join(z_session_t zs, const zp_send_join_options_t *options) {
...
+    return (z_bytes_t){.len = strlen(str), ._is_alloc = false, .start = (unsigned char *)str};
```
Mallets, line 1274: Based on my previous comment, the `len` should be provided as argument since `strlen` could not be used.

src/api/api.c
p-avital: Mostly because this constructor is mostly aimed at constructing a borrowed slice from a null terminated string. We can remove it, or make it `uint8_t` instead, but since we never care about sign in anything `z_bytes_t` related, I think the `char*` constructor would be convenient for users.

Mallets: I agree with the rationale, then I'd call the function `z_bytes_from_str` or something like that. 
Regarding my comment, there is already an internal `_z_bytes_t _z_bytes_wrap(const uint8_t *bs, size_t len);` function that could be exposed for creating `z_bytes_t` from a `const uint8_t *bs`.

p-avital: I propose we expose `z_bytes_wrap` as a way to construct slices, and `z_bytes_from_str` for convenience (documenting that it is `strlen` based and therefore should only be run on trusted input).

`_z_bytes_make(size_t capacity)` _must_ remain private, as it's a way to construct an allocated slice, and breaks the ownership model we expose publicly.

src/api/api.c
p-avital: Resolving as it's somewhat duplicating the upper conversation about whether or not to add a `from_str` constructor for convenience with trusted input.

src/api/api.c
Mallets: I agree with what you propose.

Mallets (approved):

PR #319, opened 2024-01-16
Add atomic fallback for CXX without std::atomic

PetervdPerk-NXP:
When compiling zenoh-pico on an embedded rtos without std::atomic you'll get `<atomic>` include errors. This patch provides a way to implement some of the c11 atomics used in `pointer.h` in CXX.

It's just an attempt maybe someone has an idea for a cleaner approach, also it's the question other parts of zenoh-pico are going to utilize c11 atomic.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,7 +26,7 @@ set(FRAG_MAX_SIZE 0 CACHE STRING "Use this to override the maximum size for frag
 set(BATCH_UNICAST_SIZE 0 CACHE STRING "Use this to override the maximum unicast batch size")
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON CACHE INTERNAL "")
 if(CMAKE_EXPORT_COMPILE_COMMANDS)
-  set(CMAKE_CXX_STANDARD_INCLUDE_DIRECTORIES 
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -62,6 +62,11 @@ endif()
 set(CMAKE_C_STANDARD_REQUIRED TRUE)
 add_definition(ZENOH_C_STANDARD=${CMAKE_C_STANDARD})
 
+# Useful for CXX platforms that don't provide std::atomic
+if(CXX_ATOMIC_FALLBACK)
```
cguimaraes, line 66: Could you validate this using `check_cxx_source_compiles`? In that way, the fallback would happen automatically without the need for the user to specify this flag.

PetervdPerk-NXP: Updated to use `check_cxx_source_compiles` it works on my machine, but it seems that the CI doesn't expose it

AlexisTM:
I think this is fixed by: https://github.com/eclipse-zenoh/zenoh-pico/pull/552

PetervdPerk-NXP:
Indeed #547 hides the actual atomic functions from CXX thus we CXX atomics aren't needed anymore.

Issue #312, opened 2024-01-05
RefCount for zenoh entities


imstevenpmwork:
### Describe the feature

As an example, currently if we call ```z_publisher_put(...)``` after having closed the session we get a segmentation fault:
```
zp_stop_read_task(...);
zp_stop_lease_task(...);
z_close(...);
...
z_publisher_put(...)
```
This happens because ```z_publisher_t``` maintains a reference to the session, but if this one was closed, the reference is unchanged. This is only one of the examples and even thought the order of these calls doesn't make sense, our API should be robust enough to avoid these types of bad-usage.
This can be solved by implementing a reference counting mechanism for the declared zenoh entities. The architecture is still to be defined, but I'm opening this issue to keep track of it.

cguimaraes:
I believe that @p-avital implemented (or is implementing) the baseline for a refcount mechanism.

In this case, the `close` call would be postponed until no references to the session exist.

jean-roland:
As of now, refcounting is present but only used for subscriptions and queryables

imstevenpmwork:
Solved by https://github.com/eclipse-zenoh/zenoh-pico/pull/322

PR #320, opened 2024-01-17
Replace session pointer with refcount

jean-roland:
The entities `session` `publisher`, `subscriber` and `queryables` were sharing ownership of the same `_z_session_t` pointer, this caused issues like #312. This PR replaces the session pointer with a refcounter on the session pointer.

Had to modify how the refcount implementation was handled:
* Restricted the thread unsafe one to single thread mode only
* Added a c99 gcc only implementation

And renamed `pointer.h` / `_sptr_t` to `refcount.h` / `_rc_t`


include/zenoh-pico/collections/refcount.h
```diff
@@ -0,0 +1,261 @@
...
+#define _z_memory_order_acquire std::memory_order_acquire
+#define _z_memory_order_release std::memory_order_release
+#define _z_memory_order_relaxed std::memory_order_relaxed
+#endif  // __cplusplus
+
+/*------------------ Internal Array Macros ------------------*/
+#define _Z_REFCOUNT_DEFINE(name, type)                                                          \
+    typedef struct name##_rc_t {                                                                \
+        type##_t *ptr;                                                                          \
+        _z_atomic(unsigned int) * _cnt;                                                         \
```
p-avital, line 48: Why not define a `struct name##_rc_inner_t {type##_t value; _z_atomic(unsigned int) _cnt;};` and have a single pointer to it?

jean-roland: That's an interesting suggestion.

jean-roland:
Have to close this one with the main branch renaming.

PR #322, opened 2024-01-17
Replace session pointer with refcount

jean-roland:
Reopening of #320.

The entities `session` `publisher`, `subscriber` and `queryables` were sharing ownership of the same `_z_session_t` pointer, this caused issues like #312. This PR replaces the session pointer with a refcounter on the session pointer.

Had to modify how the refcount implementation was handled:
* Restricted the thread unsafe one to single thread mode only
* Added a c99 gcc only implementation

And renamed `pointer.h` / `_sptr_t` to `refcount.h` / `_rc_t`


include/zenoh-pico/collections/refcount.h
```diff
@@ -0,0 +1,261 @@
...
+#define _z_atomic_fetch_sub_explicit std::atomic_fetch_sub_explicit
+#define _z_memory_order_acquire std::memory_order_acquire
+#define _z_memory_order_release std::memory_order_release
+#define _z_memory_order_relaxed std::memory_order_relaxed
+#endif  // __cplusplus
+
+/*------------------ Internal Array Macros ------------------*/
+#define _Z_REFCOUNT_DEFINE(name, type)                                                          \
+    typedef struct name##_rc_t {                                                                \
+        type##_t *ptr;                                                                          \
```
p-avital, line 47: This should be a single pointer to `struct name##_rc_inner_t {type##_t value; _z_atomic(unsigned int) _cnt;};`, making the type smaller, reducing the number of allocations and reducing the number of cache lines occupied by a single value.

p-avital (changes requested):
I'd really like that change to the structure of refcounts, but the rest is approved

p-avital (approved):

Issue #278, opened 2023-11-09
failed to use `find_package(zenohpico)` [Bug] 


milyin:
### Describe the bug

There are 3 proper ways to use zenoh-pico library in customer cmake project:
- `find_package(zenohpico)`
- `add_subdirectory(../zenoh-pico)`
- `FetchContent` from github repository, which is just a wrapper over the `add_subdirectory` command

The `find_package` includes file `lib/cmake/zenohpico/zenohpicoConfig.cmake` generated from https://github.com/eclipse-zenoh/zenoh-pico/blob/master/PackageConfig.cmake.in 

The `add_subdirectory` includes project's CMakeLists.txt: https://github.com/eclipse-zenoh/zenoh-pico/blob/master/CMakeLists.txt

This approach works well for zenoh-c and zenoh-cpp. See for example testing of these featues in zenoh-c CI:

Install zenohc package:
https://github.com/eclipse-zenoh/zenoh-c/blob/master/.github/workflows/ci.yml#L43-L45
Test building examples using zenoh-c directory as subproject:
https://github.com/eclipse-zenoh/zenoh-c/blob/master/.github/workflows/ci.yml#L72-L74
Test building examples using installed zenohc package:
https://github.com/eclipse-zenoh/zenoh-c/blob/master/.github/workflows/ci.yml#L79-L81

The CMakeLists.txt in `examples` directory of zenoh-pico also allows to test both approaches. At this moment building with `find_package` doesnt work for zenoh-pico

It's necessary to fix it and add testing of both approaches to CI, like it's done in zenoh-c

### To reproduce

Install zenohpico as a package locally
```
git clone git@github.com:eclipse-zenoh/zenoh-pico.git
cmake -S zenoh-pico -B build -DCMAKE_INSTALL_PREFIX=~/FOO/install
cmake --build build --target install
```

Build examples using zenoh-pico as package:
```
cmake -S zenoh-pico/examples -B build_examples -DZENOHPICO_SOURCE=PACKAGE -DCMAKE_INSTALL_PREFIX=~/FOO/install
cmake --build build_examples 
```

This fails due to misconfigured line
https://github.com/eclipse-zenoh/zenoh-pico/blob/master/examples/CMakeLists.txt#L7
(zenohc instead of zenohpico::static)

But even after fix this the more problem comes: include direcrories and preprocessor settings are not provided to dependent project correctly

### System info

Ubuntu 22.04

Mallets:
[...]

PR #324, opened 2024-01-22
Fix build examples using find_package

jean-roland:
This fixes issue #278, the examples can now build in standalone with a preinstalled zenohpico library.


examples/CMakeLists.txt
```diff
@@ -19,6 +19,7 @@ function(add_example name)
 endfunction()
 
 if(UNIX)
+    add_definitions(-DZENOH_LINUX)
```
cguimaraes, line 22: If you define it here as well, then I suggest that you use instead the `CMAKE_SYSTEM_NAME` in a similar way to the root `CMakeLists.txt`.

jean-roland:
For now we should merge the fix and make the deeper rework on CMake mentioned by @milyin at a later date, created issue #332 for this purpose.

Mallets (approved):

PR #327, opened 2024-01-24
Add owned_query entity

jean-roland:
This new entitiy based on a refcounter allows the query entitiy received in a queryable callback to be extracted from said callback in order to be processed in another context, if needed.

Requested by @p-avital.

### DIFF
--- a/src/net/query.c
+++ b/src/net/query.c
@@ -13,6 +13,10 @@
 
 #include "zenoh-pico/net/query.h"
 
+void _z_query_clear(_z_query_t *q) {
+    _ZP_UNUSED(q);  // Nothing to do
+}
+
 #if Z_FEATURE_QUERYABLE == 1
 void _z_queryable_clear(_z_queryable_t *qbl) {
     // Nothing to clear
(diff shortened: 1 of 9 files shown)


src/session/queryable.c
```diff
@@ -171,13 +172,17 @@ int8_t _z_trigger_queryables(_z_session_t *zn, const _z_msg_query_t *query, cons
         q._value.encoding = query->_ext_value.encoding;
         q._value.payload = query->_ext_value.payload;
```
p-avital, line 173: do we have ownership of this payload? We didn't before (otherwise we'd have leaks), so there should be a change here to get it. Otherwise, use-after-frees are coming our way.

This applies to the other fields as well :(

src/net/query.c
```diff
@@ -13,6 +13,10 @@
 
 #include "zenoh-pico/net/query.h"
 
+void _z_query_clear(_z_query_t *q) {
```
p-avital, line 16: This is the final destructor (called when refcount hits 0) for the query, right?

Then this is likely where all the memory freeing and `REPLY_FINAL` triggering logic should be.

p-avital (changes requested):
I don't think the owned query actually owns anything right now

PR #336, opened 2024-01-30
Renaming questionable

jean-roland:
In zenoh-pico, for zenoh entities like `Queryable` or `Subscriber`, we need to distinguish between the object given to the user and the object we store in the zenoh session.

For `Subscriber` we have `_z_subscriber` and `_z_subscription` which is fine but for `Queryable` we had `_z_queryable` and `_z_questionable` which isn't great as the link between the two isn't clear.

As the problem may arise again for the yet-to-be-implemented `Interest` entity, `_z_questionable` should be renamed to `_z_session_queryable` to more accurately represent the link between those objects and this scheme is applicable to future entities.

As such this PR renames:
* `_z_questionable` -> `_z_session_queryable`
* `questionable_handler` -> `queryable_handler` (the callback type)
* `local_questionables` -> `local_queryable` (the session's list of entity)

### DIFF
--- a/include/zenoh-pico/api/primitives.h
+++ b/include/zenoh-pico/api/primitives.h
@@ -589,7 +589,7 @@ z_owned_closure_sample_t z_closure_sample(_z_data_handler_t call, _z_dropper_han
  * Returns:
  *   Returns a new query closure.
  */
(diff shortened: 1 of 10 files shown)


Mallets (approved):

p-avital:
RIP questionables :(

PR #348, opened 2024-02-15
Add support for qos settings in sample

DenisBiryukov91:
Add support for qos settings in sample (see https://github.com/eclipse-zenoh/zenoh/pull/730)

### DIFF
--- a/include/zenoh-pico/protocol/core.h
+++ b/include/zenoh-pico/protocol/core.h
@@ -215,14 +215,14 @@ _z_keyexpr_t _z_rid_with_suffix(uint16_t rid, const char *suffix);
  * QoS settings of zenoh message.
  *
  * Members:
- *   _z_priority_t priority: Priority of the message.
- *   _z_congestion_control_t congestion_control: Congestion control of the message.
- *   bool express: If true, the message is not batched during transmission, in order to reduce latency.
+ *   z_priority_t priority: Priority of the message.
+ *   z_congestion_control_t congestion_control: Congestion control of the message.
+ *   _Bool express: If true, the message is not batched during transmission, in order to reduce latency.
  */
 typedef struct {
     z_priority_t priority;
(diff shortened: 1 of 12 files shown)


include/zenoh-pico/protocol/core.h
```diff
@@ -211,6 +211,20 @@ _z_keyexpr_t _z_rname(const char *rname);
...
+ *   _z_priority_t priority: Priority of the message.
+ *   _z_congestion_control_t congestion_control: Congestion control of the message.
+ *   bool express: If true, the message is not batched during transmission, in order to reduce latency.
+ */
+typedef struct {
+    z_priority_t priority;
+    z_congestion_control_t congestion_control;
+    bool express;
+} _z_qos_t;
+
```
jean-roland, line 227: The comment and the structure do not agree on type visibility (`_z_priority_t` vs `z_priority_t`), also we use `_Bool` for booleans generally.

include/zenoh-pico/protocol/core.h
```diff
@@ -227,6 +241,7 @@ typedef struct {
     _z_timestamp_t timestamp;
     _z_encoding_t encoding;
     z_sample_kind_t kind;
+    _z_qos_t qos;
```
jean-roland, line 244: Pretty sure by putting the private type here, the public alias `z_qos_t` is never used. So I'm tempted to use `z_qos_t` instead.

jean-roland (changes requested):

include/zenoh-pico/protocol/core.h
DenisBiryukov91: fixed

jean-roland (approved):

jean-roland (approved):

jean-roland (approved):

milyin (approved):

Issue #331, opened 2024-01-26
Protocol update:  Unicast Interests


jean-roland:
### Describe the feature

Add Interests support for unicast transport on zenoh pico

PR #349, opened 2024-02-20
Add unicast interest feature

jean-roland:
This PR is the first concerning the upcoming `Declare interest` feature, it contains mainly the modification for unicast/client but also prepare the terrain for multicast/peer.

* Add a config token to toggle the non-mandatory part of the feature
* Add the interest module that handles interest management at the session level and the processing of interest related messages (tx and rx)
* Add the filtering module that build on the interest module to add a writer side filtering for publishers (they won't write on the network until they receive a declare subscriber).
* Add the basic multicast logic to reply to interest messages.

This is a partial implementation of #331


src/api/api.c
```diff
@@ -728,33 +743,33 @@ z_publisher_delete_options_t z_publisher_delete_options_default(void) {
...
     // Trigger local subscriptions
     _z_trigger_local_subscriptions(&pub._val->_zn.in->val, pub._val->_key, payload, len
 #if Z_FEATURE_ATTACHMENT == 1
                                    ,
                                    opt.attachment
 #endif
     );
-
-    return ret;
+    // Check if write filter is active before writing
```
OlivierHecart, line 761: Push messages are now sent to the network after local subscriptions trigger. We tend to have the previous order on purpose.

OlivierHecart:
I guess we should add a toggle for `Z_FEATURE_INTEREST` in `CMakeLists.txt` and `GNUMakefile` (enabled by default ?).
And probably enable it in `.github/workflows/build-check.yaml`, `.github/workflows/freertos_plus_tcp.yaml` and `docs/conf.py`.

OlivierHecart (approved):

PR #354, opened 2024-02-24
Fix/remove qos designated initializers

### LINKED ISSUE #352: remove C++17 incompatible designated initializers from headers [Bug]
When compiling zenoh-pico with C++ compiler the following error appears:
```
 /home/runner/work/zenoh-cpp/zenoh-cpp/build/_deps/zenohpico-src/include/zenoh-pico/protocol/definitions/network.h:66:23: error: C++ designated initializers only available with ‘-std=c++20’ or ‘-std=gnu++20’ [-Werror=pedantic]
   66 |     _z_n_qos_t ret = {._val = (uint8_t)((express << 4) | (nodrop << 3) | priority)};
```

This should be corrected like it was done in this fix:
https://github.com/eclipse-zenoh/zenoh-pico/commit/e76e367c3f9df68cf631872ce970b87f3be8803a

To reproduce the issue build https://github.com/eclipse-zenoh/zenoh-cpp project

DenisBiryukov91:
move qos-related functions from header into .c to hide usage of designated initializers from c++ (closes #352)

### DIFF
--- a/include/zenoh-pico/protocol/definitions/network.h
+++ b/include/zenoh-pico/protocol/definitions/network.h
@@ -60,10 +60,19 @@
 
 typedef _z_qos_t _z_n_qos_t;
 
-_z_qos_t _z_n_qos_create(_Bool express, z_congestion_control_t congestion_control, z_priority_t priority);
-z_priority_t _z_n_qos_get_priority(_z_n_qos_t n_qos);
-z_congestion_control_t _z_n_qos_get_congestion_control(_z_n_qos_t n_qos);
-_Bool _z_n_qos_get_express(_z_n_qos_t n_qos);
+static inline _z_qos_t _z_n_qos_create(_Bool express, z_congestion_control_t congestion_control, z_priority_t priority) {
+    _z_n_qos_t ret;
+    _Bool nodrop = congestion_control == Z_CONGESTION_CONTROL_DROP ? 0 : 1;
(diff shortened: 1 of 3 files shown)


milyin (changes requested):
I think it's better to make the fix in the header, without moving implementation to C, like here https://github.com/eclipse-zenoh/zenoh-pico/commit/e76e367c3f9df68cf631872ce970b87f3be8803a
Moving code to C may affect the performance (additional call instead of inlining)

include/zenoh-pico/protocol/definitions/network.h
```diff
@@ -62,8 +62,9 @@ typedef _z_qos_t _z_n_qos_t;
 
 static inline _z_qos_t _z_n_qos_create(_Bool express, z_congestion_control_t congestion_control,
                                        z_priority_t priority) {
+    _z_n_qos_t ret;
     _Bool nodrop = congestion_control == Z_CONGESTION_CONTROL_DROP ? 0 : 1;
-    _z_n_qos_t ret = {._val = (uint8_t)((express << 4) | (nodrop << 3) | priority)};
+    ret._val = (uint8_t)((express << 4) | (nodrop << 3) | priority);
```
milyin, line 67: I'd prefer the form 
```
_z_n_qos_t ret = { uint8_t)((express << 4) | (nodrop << 3) | priority };
```
This would guarantee that all fields in `_z_n_qos_t` are set.
Though this variant works too.

milyin (approved):

PR #359, opened 2024-02-28
Align examples and remove reading from stdin

oteffahi:
These changes remove reading from stdin for all examples, and align their behaviors and outputs across projects.

This PR is part of an effort across currently active Zenoh projects. To maintain examples' implementations as close to each other as possible, any changes identified in other projects will be replicated in this PR. As such, **please do not merge these changes until all PRs are ready for merging**.

Related PRs:
- https://github.com/eclipse-zenoh/zenoh/pull/768
- https://github.com/eclipse-zenoh/zenoh-c/pull/255
- https://github.com/eclipse-zenoh/zenoh-cpp/pull/103
- https://github.com/eclipse-zenoh/zenoh-java/pull/45
- https://github.com/eclipse-zenoh/zenoh-kotlin/pull/47
- https://github.com/eclipse-zenoh/zenoh-python/pull/150


jean-roland (approved):

jean-roland (approved):

oteffahi:
The CI failed because the c11/z_get example currently does not block to wait for responses; I worked under the assumption that the `z_get` function is a blocking call that awaits a final queryable response before returning to the main function. I will fix this as soon as possible.

examples/unix/c11/z_get.c
```diff
@@ -12,16 +12,30 @@
...
+        exit(EXIT_FAILURE);      \
+    } while (0)
+pthread_cond_t cond;
+pthread_mutex_t mutex;
+
 void reply_dropper(void *ctx) {
     (void)(ctx);
     printf(">> Received query final notification\n");
+    pthread_cond_signal(&cond);
+    pthread_cond_destroy(&cond);
```
jean-roland, line 38: This should use the public sync mechanisms found in `include/zenoh-pico/system/platform.h`, this impacts all the following pthread_cond and pthread_mutex calls

examples/unix/c11/z_get.c
```diff
@@ -112,6 +130,14 @@ int main(int argc, char **argv) {
         printf("Unable to send query.\n");
         return -1;
     }
+    struct timespec query_timeout;
+    clock_gettime(CLOCK_REALTIME, &query_timeout);
```
jean-roland, line 134: This should use the public timer mechanisms found in `include/zenoh-pico/system/platform.h`, this impacts all the following clock calls

examples/unix/c11/z_get.c
```diff
@@ -12,16 +12,30 @@
...
+#include <errno.h>
+#include <pthread.h>
 #include <stddef.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <unistd.h>
 #include <zenoh-pico.h>
 
 #if Z_FEATURE_QUERY == 1
+#define QUERY_TIMEOUT 10  // query timeout in seconds
```
jean-roland, line 24: Note: we'll have to revise this example when timeout is exposed by zenoh pico

jean-roland (changes requested):

jean-roland (approved):

Mallets:
This job is failing: https://github.com/eclipse-zenoh/zenoh-pico/actions/runs/8297508365/job/22713592044?pr=359

jean-roland (approved):

jean-roland (approved):

examples/freertos_plus_tcp/z_pub_st.c
```diff
@@ -54,7 +55,7 @@ void app_main(void) {
 
     char *buf = (char *)pvPortMalloc(256);
     z_clock_t now = z_clock_now();
-    for (int idx = 0; 1;) {
+    for (int idx = 0; idx < N;) {
```
Mallets, line 58: In this case after `10` publications the example will exit, however all other examples kepp publishing indefinitely.

examples/freertos_plus_tcp/z_sub_st.c
```diff
@@ -27,13 +27,17 @@
 #endif
 
 #define KEYEXPR "demo/example/**"
+#define N 10
```
Mallets, line 30: We should align the behaviour with all other bindings, I believe we don't wait for just `10` samples by default.

oteffahi: Other bindings do not implement single-threaded pub/sub examples, therefore we only need to align the configuration of these examples across their different plateform-specific implementations. In this PR, the configuration is as follows:
- On unix (x11, x99): `N` must be provided as a CLI argument to run the example. This is also used in our CI to run the test with a known number of messages. 
- On windows: since `getopt` is unavailable, `N` is initialized to `int32::max`.
- On freertos: I understand that CLI argument-passing may not be possible on certain embedded targets that run on freertox, therefore `N` is initialized to 10 to avoid running the example indefinitely. Should we change it to `int32::max` ?

oteffahi: As per the other comment on this file, I'll update all implementations to start with a default number of samples of `int32:max`, and allow unix implementations to overwrite it through CLI argument.

jean-roland (approved):

examples/freertos_plus_tcp/z_sub_st.c
Mallets: The behaviour of a `z_pub` in all languages is to publish indefinitely, so `uint32::MAX` is close enough :)

Mallets (approved):

PR #364, opened 2024-03-01
Add config endpoint in raweth locator

jean-roland:
The first version of endpoint configuration was in a static source file, this has been removed and now the whole config is sent through the locator.


gmartin82 (approved):
The changes look good to me. I've tested various combinations of the locator string and the mapping and whitelist functions are working. Also confirmed the correct source MAC is used by observing the traffic in Wireshark.

Mallets (approved):

PR #375, opened 2024-03-19
Expose timeout option in z_get_options_t

### LINKED ISSUE #215: Expose timeout option in z_get_options_t
### Describe the feature

Describe the feature

The timeout option (from rust: https://docs.rs/zenoh/0.7.0-rc/zenoh/query/struct.GetBuilder.html#method.timeout) should be exposed in zenoh-pico get options `z_get_options_t`.

sashacmc:
Closes #215

### DIFF
--- a/src/api/api.c
+++ b/src/api/api.c
@@ -788,7 +788,7 @@ z_get_options_t z_get_options_default(void) {
 #if Z_FEATURE_ATTACHMENT == 1
         // TODO:ATT.attachment = z_attachment_null()
 #endif
-        .timeout_ms = 0
+        .timeout_ms = Z_GET_TIMEOUT_DEFAULT
     };
 }
 
--- a/src/api/api.c
+++ b/src/api/api.c
@@ -788,6 +788,7 @@ z_get_options_t z_get_options_default(void) {
 #if Z_FEATURE_ATTACHMENT == 1
         // TODO:ATT.attachment = z_attachment_null()
 #endif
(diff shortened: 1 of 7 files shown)


src/api/api.c
```diff
@@ -788,6 +788,7 @@ z_get_options_t z_get_options_default(void) {
 #if Z_FEATURE_ATTACHMENT == 1
         // TODO:ATT.attachment = z_attachment_null()
 #endif
+        .timeout_ms = 0
```
jean-roland, line 791: To make it consistent with Zenoh it should have the same default value of 10 seconds. Although you might need to double check that value.

Mallets:
This PR does not seem to implement the timeout behaviour but only to add the option in the options. Will a query ever timeout locally? Or it will always depend on a router sending the timeout? What about p2p multicast case?

jean-roland:
Right, it eluded me but yeah we need to implement the timeout locally, like zenoh iirc.

sashacmc:
@Mallets as far as I see, there we have no any timeout inside, just send the request (response comes asynchronously)
Should we introduce some waiting?

Mallets:
We could piggyback on the `lease_task` which basically acts as a timer for events, what do you think?

jean-roland (approved):

Mallets:
Let's get this PR in and leave the internal timeout implementation to a different PR.

Mallets (approved):

PR #391, opened 2024-03-27
Rework owned sample

sashacmc:
Rework owned sample processing for pull and sub_channel:
 - Add multithreaded collections: `fifo_mt` and `ring_mt`
 - Add channel definition macro: `_Z_CHANNEL_DEFINE`
 - Add closure for `z_owned_sample_t`: `z_owned_closure_owned_sample_t` 
 - Define `z_owned_sample_ring_channel_t` and `z_owned_sample_fifo_channel_t` channels

_(NB: channels for queries and replies will be provided in the separate PR)_


examples/unix/c11/z_sub_channel.c
```diff
@@ -64,7 +64,7 @@ int main(int argc, char **argv) {
     }
 
     printf("Declaring Subscriber on '%s'...\n", keyexpr);
-    z_owned_sample_channel_t channel = z_sample_channel_fifo_new(3);
+    z_owned_sample_fifo_channel_t channel = z_sample_fifo_channel(3);
```
Mallets, line 67: In zenoh-c we have `z_*_new()`: https://github.com/eclipse-zenoh/zenoh-c/blob/c1498fcaf08d3dad15e49b1ac15cd380c0dc175c/examples/z_queryable_with_channels.c#L57

We can decide to keep the new or to remove it. In any case, zenoh-pico and zenoh-c should be aligned.

include/zenoh-pico/api/handlers.h
```diff
@@ -15,109 +15,74 @@
...
+        if (res) {                                                                                                \
+            _Z_ERROR("%s failed: %i", #collection_pull_f, res);                                                   \
+        }                                                                                                         \
+    }                                                                                                             \
+                                                                                                                  \
+    static inline z_owned_##name##_t z_##name(size_t capacity) {                                                  \
+        z_owned_##name##_t channel;                                                                               \
+        channel.collection = collection_new_f(capacity);                                                          \
+        channel.send = z_##send_closure_name(_z_##name##_send, NULL, channel.collection);                         \
+        channel.recv = z_##recv_closure_name(_z_##name##_recv, NULL, channel.collection);                         \
```
Mallets, line 68: Should we have as well a `try_recv` method? It could be quite handy similar to what is being done in https://github.com/eclipse-zenoh/zenoh/pull/903/files.

It could come in a separate PR.

src/collections/fifo_mt.c
```diff
@@ -0,0 +1,150 @@
...
+
+int8_t _z_fifo_mt_push(const void *elem, void *context, z_element_free_f element_free) {
+    if (elem == NULL || context == NULL) {
+        return _Z_ERR_GENERIC;
+    }
+
+    _z_fifo_mt_t *f = (_z_fifo_mt_t *)context;
+
+#if Z_FEATURE_MULTI_THREAD == 1
+    int res = zp_mutex_lock(&f->_mutex);
```
Mallets, line 85: Mainly a naming thing. `res` is actually `err` here, isn't it?

src/collections/fifo_mt.c
```diff
@@ -0,0 +1,150 @@
...
+
+    return _Z_RES_OK;
+}
+
+int8_t _z_fifo_mt_pull(void *dst, void *context, z_element_move_f element_move) {
+    _z_fifo_mt_t *f = (_z_fifo_mt_t *)context;
+
+#if Z_FEATURE_MULTI_THREAD == 1
+    void *src = NULL;
+    int res = zp_mutex_lock(&f->_mutex);
```
Mallets, line 119: Mainly a naming thing. `res` is actually `err` here, isn't it?

src/collections/fifo_mt.c
Mallets: The return type is `int8_t` but `res` has `int` type.

src/collections/ring_mt.c
```diff
@@ -0,0 +1,114 @@
...
+    }
+#endif
+    return _Z_RES_OK;
+}
+
+int8_t _z_ring_mt_pull(void *dst, void *context, z_element_move_f element_move) {
+    _z_ring_mt_t *r = (_z_ring_mt_t *)context;
+
+#if Z_FEATURE_MULTI_THREAD == 1
+    int res = zp_mutex_lock(&r->_mutex);
```
Mallets, line 95: The return type is `int8_t` but `res` has `int` type.

tests/z_channels_test.c
```diff
@@ -0,0 +1,118 @@
...
+#include <assert.h>
+
+#define SEND(channel, v)                           \
+    do {                                           \
+        z_sample_t sample;                         \
+        sample.payload.start = (const uint8_t *)v; \
+        sample.payload.len = strlen(v);            \
+        sample.keyexpr = _z_rname("key");          \
+        z_call(channel.send, &sample);             \
+    } while (0);
```
Mallets, line 31: Just a curiosity. Are you using here ` do { ... } while (0)` as a way to implement scoping `{ ... }`?

src/api/handlers.c
```diff
@@ -11,176 +11,22 @@
...
-#endif
-        if (res == _Z_RES_OK) {
-            z_owned_closure_sample_t send = z_closure(z_sample_channel_ring_push, NULL, ring);
-            ch.send = send;
-            z_owned_closure_owned_sample_t recv = z_closure(z_sample_channel_ring_pull, NULL, ring);
-            ch.recv = recv;
-        } else {
-            zp_free(ring);
-        }
+    if (dst && src) {
```
Mallets, line 27: In zenoh-pico we prefer explicit pointer checks, i.e. `if (dst != NULL && src != NULL)`

src/collections/fifo_mt.c
```diff
@@ -0,0 +1,150 @@
...
+
+/*-------- Fifo Buffer Multithreaded --------*/
+int8_t _z_fifo_mt_init(_z_fifo_mt_t *fifo, size_t capacity) {
+    int8_t res = _z_fifo_init(&fifo->_fifo, capacity);
+    if (res) {
+        return res;
+    }
+
+#if Z_FEATURE_MULTI_THREAD == 1
+    res = zp_mutex_init(&fifo->_mutex);
```
jean-roland, line 28: all the zp_* system calls are deprecated, you should use z_mutex & z_condvar instead

src/api/handlers.c
```diff
@@ -11,176 +11,22 @@
...
+        dst->_value = (_z_sample_t *)zp_malloc(sizeof(_z_sample_t));
```
Mallets, line 28: If think we should check if `dst->_value != NULL` before doing the copy.

examples/unix/c11/z_sub_channel.c
sashacmc: It seemed to me that I saw a creation without `_new`, but I now double-checked, this is not so, indeed in most cases we have a creation with `_new`, thanks!

src/collections/fifo_mt.c
sashacmc: Yes, err will be better, thanks!

tests/z_channels_test.c
```diff
@@ -0,0 +1,118 @@
...
+            z_drop(z_move(sample));                                                                       \
+        } else {                                                                                          \
+            buf[0] = '\0';                                                                                \
+        }                                                                                                 \
+    } while (0);
+
+void sample_fifo_channel_test(void) {
+    z_owned_sample_fifo_channel_t channel = z_sample_fifo_channel(10);
+
+    SEND(channel, "v1");
```
jean-roland, line 49: With a closing `;` on the macro and another one on the call, `-Wpedantic` won't be happy

tests/z_channels_test.c
jean-roland: I explained the rationale: https://github.com/eclipse-zenoh/zenoh-pico/pull/307#pullrequestreview-1797921122

src/collections/fifo_mt.c
sashacmc: I don't see them in this branch :(

src/collections/fifo_mt.c
```diff
@@ -14,44 +14,33 @@
...
-    }
-    res = zp_condvar_init(&fifo->_cv_not_full);
-    if (res) {
-        return res;
-    }
-    res = zp_condvar_init(&fifo->_cv_not_empty);
-    if (res) {
-        return res;
-    }
+    _Z_RETURN_IF_ERR(zp_mutex_init(&fifo->_mutex))
```
jean-roland, line 26: Still needs to replace the deprecrated systems calls: https://github.com/eclipse-zenoh/zenoh-pico/blob/main/src/deprecated/platform.c

jean-roland: Just realized this changes won't be in this branch until #394 is merged. So we either accept as is and modify later  or we wait till #394 is merged and rebase on that. Opinions @Mallets?

Mallets: I suggest to ago ahed with this PR and we then update #394. @jean-roland what do you think?

jean-roland (approved):

src/net/memory.c
```diff
@@ -49,6 +49,24 @@ void _z_sample_free(_z_sample_t **sample) {
     }
 }
 
+void _z_sample_copy(_z_sample_t *dst, const _z_sample_t *src) {
+    dst->keyexpr = _z_keyexpr_duplicate(src->keyexpr);
+    dst->payload = _z_bytes_duplicate(&src->payload);
+    dst->timestamp = _z_timestamp_duplicate(&src->timestamp);
+    // TODO(sashacmc): should be implemented after encoding rework
```
Mallets, line 56: So `dst->encoding` is left unitialized?

sashacmc: Yes, this is not very good, I agree, I added a copy of the current encoding implementation

Mallets (approved):

PR #392, opened 2024-03-28
Decode fails when interest id flag is set

Mallets:
https://github.com/eclipse-zenoh/zenoh/pull/870 adds an interest id in the declare message to disambiguate wether the declaration is sent in a response of a declare interest with `future == false` or not.

However, zenoh-pico never sends an interest with `future == false`, so the interest id should never be set. If set, it means the protocol is misbehaving. Therefore, we return an error during decoding.


Mallets:
@jean-roland can you help reviewing this?

jean-roland (approved):

PR #399, opened 2024-04-09
Update encoding format on the wire

jean-roland:
The way `encoding_t` prefix is sent on the wire changed to be 31bits + suffix flag. Suffix is only written when present.

Because of misalignment on interest format with Zenoh, interests are deactivated by default for now.


jean-roland:
@Mallets could you review this PR?

Mallets:
Definition of `_z_encoding_t` is still:

```C
typedef struct {
    _z_bytes_t suffix;
    z_encoding_prefix_t prefix;
} _z_encoding_t;
```

It should become:
```C
typedef struct {
    _z_bytes_t schema;
    uint16_t id;
} _z_encoding_t;
```

`z_encoding_prefix_t` is meant to be deprecated and replaced with a set of constants.

Mallets (approved):

PR #400, opened 2024-04-09
Add query and reply channels

sashacmc:
Add query and reply channels:
 - z_owned_query_ring_channel_t
 - z_owned_query_fifo_channel_t
 - z_owned_reply_ring_channel_t
 - z_owned_reply_fifo_channel_t

Add examples:
 - z_queryable_channel.c
 - z_get_channel.c


examples/unix/c11/z_get_channel.c
```diff
@@ -0,0 +1,128 @@
...
+        opts.value.payload = _z_bytes_wrap((const uint8_t *)value, strlen(value));
+    }
+    z_owned_reply_ring_channel_t channel = z_reply_ring_channel_new(1);
+    if (z_get(z_loan(s), ke, "", z_move(channel.send), &opts) < 0) {
+        printf("Unable to send query.\n");
+        return -1;
+    }
+
+    z_owned_reply_t reply = z_reply_null();
+    z_call(channel.recv, &reply);
```
Mallets, line 101: I think we should use `for(z_call(..), z_check(..), z_call(..))`. The current code only reads one reply but it does not wait for all the replies to arrive.

Mallets (approved):

Mallets:
@jean-roland this PR looks good to me. However the CI seems to fail. I believe the failure is not related to the changes proposed in this PR but I'm not 100% sure. Could you please double check?

jean-roland:
There are issues that are fixed by #399 but there also seems to be issues related to this PR when some features are disabled.

sashacmc:
I fixed the compilation with disabled features and the memory leak (thanks @jean-roland for pointing out)
@Mallets you can merge it.

Mallets (approved):

PR #401, opened 2024-04-10
Fix concurrency issue on task_running value

jean-roland:
#306 introduced a concurrency issue on `task_running` which caused some OS (espidf) to stop their read & lease tasks at startup. This PR revert this issue.

@oteffahi could you do the review?


oteffahi (approved):
LGTM 👍

Mallets (approved):

PR #411, opened 2024-05-15
Fix windows socket timeout value

### LINKED ISSUE #409: [Bug] setsockopt SO_RCVTIMEO not working in Windows
### Describe the bug

In network.c, `_z_open_udp_unicast()` calls `setsockopt()` passing a `z_time_t` value, while in Windows it should be `DWORD` type.
Fixed by using:
```
#ifdef WIN32
        DWORD tv = tout;
#else
        z_time_t tv;
        tv.time = tout / (uint32_t)1000;
        tv.millitm = tout % (uint32_t)1000;
#endif
```

### To reproduce

Start either a publisher or a subscriber with no parameters in Windows without zenohd running.
The application will stuck in Opening session... and the expected timeout is not honored.

### System info

Windows 11

jean-roland:
This fixes #409, solves also some C4098 compilation warning.


gmartin82 (approved):
Changes look good to me. tv is now a DWORD rather than a z_time_t as required.

PR #438, opened 2024-06-24
Align version with zenoh-c

diogomatsubara:
As per https://github.com/eclipse-zenoh/zenoh-c/pull/466, align how we configure zenoh CMake based projects version.


Mallets (approved):

fuzzypixelz:
I think this should be reverted as `project_version` is not used by this repository's CMake script (unlike zenoh-c). 

[Instead we could benefit from a README section similar to zenoh-c's](https://github.com/eclipse-zenoh/zenoh-c/?tab=readme-ov-file#versioning).

PR #440, opened 2024-06-25
Remove refcounting from reply/sample.

jean-roland:
Oudated.

_For Zenoh-C compatibility issue, we have a single `z_loaned_sample_t` type which corresponds both to a sample received by a sub and the sample contained in a reply.

This type is refcounted as required for potential out-of-context usage by a subscriber. As replies are themselves refcounted that create a double refcount which is slightly wasteful memory-wise in a Query/reply scenario.

This PR attemps to offer a way to avoid this double refcounting by storing the reply sample without refcounting, provide a zenoh pico api to interact with it, while a call to the Zenoh-C api will still generate the recounted sample._

### DIFF
--- a/include/zenoh-pico/api/handlers.h
+++ b/include/zenoh-pico/api/handlers.h
@@ -51,9 +51,15 @@
             return;                                                                                                  \
         }                                                                                                            \
         if (elem == NULL) {                                                                                          \
-            internal_elem->_rc.in = NULL;                                                                            \
+            internal_elem->_val = NULL;                                                                              \
         } else {                                                                                                     \
-            elem_copy_f(&internal_elem->_rc, elem);                                                                  \
+            internal_elem->_val = (elem_loaned_type *)z_malloc(sizeof(elem_loaned_type));                            \
+            if (internal_elem->_val == NULL) {                                                                       \
+                z_free(internal_elem);                                                                               \
+                _Z_ERROR("Out of memory");                                                                           \
+                return;                                                                                              \
+            }                                                                                                        \
+            elem_copy_f(internal_elem->_val, (elem_loaned_type *)elem);                                              \
         }                                                                                                            \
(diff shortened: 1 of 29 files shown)


DenisBiryukov91 (approved):

sashacmc (approved):

include/zenoh-pico/api/types.h
```diff
@@ -184,8 +184,8 @@ _Z_LOANED_TYPE(_z_queryable_t, queryable)
  * Represents a Zenoh Query entity, received by Zenoh Queryable entities.
  *
  */
-_Z_OWNED_TYPE_RC(_z_query_rc_t, query)
-_Z_LOANED_TYPE(_z_query_rc_t, query)
+_Z_OWNED_TYPE_PTR(_z_query_t, query)
```
DenisBiryukov91, line 187: I just realized, that query likely should stay ref counted, because zenoh allows to make copies of queries and send multiple replies through them. The query is considered to be finalized when the last copy is dropped.

DenisBiryukov91 (changes requested):
I just realized, that query likely should stay ref counted, because zenoh allows to make copies of queries and send multiple replies through them. The query is considered to be finalized when the last copy is dropped.

include/zenoh-pico/api/handlers.h
```diff
@@ -51,9 +51,15 @@
             return;                                                                                                  \
         }                                                                                                            \
         if (elem == NULL) {                                                                                          \
-            internal_elem->_rc.in = NULL;                                                                            \
```
DenisBiryukov91, line 54: can we just use standard _clone methods that we define almost for every owned type, to avoid duplicating handlers code for rc and not-rc ?

DenisBiryukov91 (approved):

milyin (approved):

PR #441, opened 2024-06-25
Align version with zenoh-c (#438)

diogomatsubara:
Cherry-pick changes from `main` into `dev/1.0.0`

As per https://github.com/eclipse-zenoh/zenoh-c/pull/466, align how we configure zenoh CMake based projects version.


version.txt
```diff
@@ -0,0 +1 @@
+0.11.0.0
```
Mallets, line 1: Shouldn't be `1.0.0.0` since this PR targets `dev/1.0.0`?

diogomatsubara: Fixed

Mallets (approved):

Mallets (approved):

fuzzypixelz:
I think [f8a5238570dcf15666e03733711c0f48b84ac048 should be reverted as `project_version` is not used by this repository's CMake script (unlike zenoh-c). 

[Instead we could benefit from a README section similar to zenoh-c's](https://github.com/eclipse-zenoh/zenoh-c/?tab=readme-ov-file#versioning).

PR #487, opened 2024-07-02
Add z_view_keyexpr_from_str_autocanonize

sashacmc:
Add z_view_keyexpr_from_str_autocanonize method

### DIFF
--- a/include/zenoh-pico/api/primitives.h
+++ b/include/zenoh-pico/api/primitives.h
@@ -74,6 +74,19 @@ int8_t z_view_keyexpr_from_str(z_view_keyexpr_t *keyexpr, const char *name);
 int8_t z_view_keyexpr_from_str_unchecked(z_view_keyexpr_t *keyexpr, const char *name);
 
 /**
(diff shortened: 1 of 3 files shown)


include/zenoh-pico/api/primitives.h
```diff
@@ -73,6 +73,19 @@ int8_t z_view_keyexpr_from_str(z_view_keyexpr_t *keyexpr, const char *name);
  */
 int8_t z_view_keyexpr_from_str_unchecked(z_view_keyexpr_t *keyexpr, const char *name);
 
+/**
```
milyin, line 76: I think documentation should be updated a bit like in zenoh c, mentioning that canonization happens in place and it's result is always shortening the original string:

/// The string is canonized in-place before being passed to keyexpr, possibly shortening it by modifying `len`.
/// May SEGFAULT if `expr` is NULL or lies in read-only memory (as values initialized with string litterals do).
/// `expr` must outlive the constucted key expression.

milyin (changes requested):

milyin (approved):

PR #490, opened 2024-07-02
Move encoding value on receive side.

### LINKED ISSUE #482: Remove extra recopy when decoding messages.
### Describe the feature

For example, when decoding messages we allocate a `_z_encoding_t`, if needed, which is stored in a `_z_msg_put_t`. Which then will be copied and stored in a `_z_query_t` / `_z_reply_t` / `_z_sample_t`. Where it could have been simply moved.

The problem might exist for keyexpr/payload/other data containers.

jean-roland:
Fixes #482, encoding is now moved from the decoded message.

### DIFF
--- a/src/net/encoding.c
+++ b/src/net/encoding.c
@@ -39,11 +39,7 @@ _z_encoding_t _z_encoding_wrap(uint16_t id, const char *schema) {
 
 _z_encoding_t _z_encoding_null(void) { return _z_encoding_wrap(_Z_ENCODING_ID_DEFAULT, NULL); }
 
-void _z_encoding_clear(_z_encoding_t *encoding) {
-    if (_z_string_check(&encoding->schema)) {
-        _z_string_clear(&encoding->schema);
-    }
-}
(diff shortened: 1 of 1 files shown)


src/net/encoding.c
```diff
@@ -39,7 +39,11 @@ _z_encoding_t _z_encoding_wrap(uint16_t id, const char *schema) {
 
 _z_encoding_t _z_encoding_null(void) { return _z_encoding_wrap(_Z_ENCODING_ID_DEFAULT, NULL); }
 
-void _z_encoding_clear(_z_encoding_t *encoding) { _z_string_clear(&encoding->schema); }
+void _z_encoding_clear(_z_encoding_t *encoding) {
+    if (_z_string_check(&encoding->schema)) {
```
DenisBiryukov91, line 43: _z_string_check does not seem to be necessary here. If it is false, _z_string_clear will be a no-op

DenisBiryukov91 (approved):

milyin (approved):

Issue #470, opened 2024-06-27
Add missed z_string_* methods


sashacmc:
### Describe the release item

 - z_string_as_slice [include/zenoh_commons.h#L3345](https://github.com/eclipse-zenoh/zenoh-c/blob/6ba997f5a91c55ddc7f849c852d50bb220fcb408/include/zenoh_commons.h#L3345)
 - z_string_empty [include/zenoh_commons.h#L3365](https://github.com/eclipse-zenoh/zenoh-c/blob/6ba997f5a91c55ddc7f849c852d50bb220fcb408/include/zenoh_commons.h#L3365)
 - z_string_from_substring [include/zenoh_commons.h#L3372](https://github.com/eclipse-zenoh/zenoh-c/blob/6ba997f5a91c55ddc7f849c852d50bb220fcb408/include/zenoh_commons.h#L3372)
 - z_string_is_empty [include/zenoh_commons.h#L3378](https://github.com/eclipse-zenoh/zenoh-c/blob/6ba997f5a91c55ddc7f849c852d50bb220fcb408/include/zenoh_commons.h#L3378)
 - z_string_wrap [include/zenoh_commons.h#L3397](https://github.com/eclipse-zenoh/zenoh-c/blob/6ba997f5a91c55ddc7f849c852d50bb220fcb408/include/zenoh_commons.h#L3397)

sashacmc:
All except `z_string_as_slice`
done in https://github.com/eclipse-zenoh/zenoh-pico/pull/495

`z_string_as_slice` can't be implemented in the proper way now, because _z_slize_t and _z_string_t are different types and can't be safely converted.

PR #495, opened 2024-07-02
Add missed z_string_* methods

sashacmc:
Add methods:
 - z_string_empty
 - z_string_from_str
 - z_string_from_substr
 - z_string_is_empty

Closes: https://github.com/eclipse-zenoh/zenoh-pico/issues/470

### DIFF
--- a/include/zenoh-pico/api/primitives.h
+++ b/include/zenoh-pico/api/primitives.h
@@ -1133,6 +1133,49 @@ const char *z_string_data(const z_loaned_string_t *str);
  */
 size_t z_string_len(const z_loaned_string_t *str);
 
+/**
+ * Builds a :c:type:`z_string_t` by wrapping a ``const char *`` string.
+ *
+ * Parameters:
+ *   str: Pointer to an uninitialized :c:type:`z_string_t`.
+ *   value: Pointer to a null terminated string.
+ *
+ * Return:
+ *   ``0`` if creation successful, ``negative value`` otherwise.
+ */
+int8_t z_string_wrap(z_owned_string_t *str, const char *value);
+
+/**
+ * Builds an empty :c:type:`z_string_t`.
+ *
+ * Parameters:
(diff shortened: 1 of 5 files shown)


include/zenoh-pico/api/primitives.h
```diff
@@ -1133,6 +1133,49 @@ const char *z_string_data(const z_loaned_string_t *str);
...
+ * Builds a :c:type:`z_string_t` by wrapping a ``const char *`` string.
+ *
+ * Parameters:
+ *   str: Pointer to an uninitialized :c:type:`z_string_t`.
+ *   value: Pointer to a null terminated string.
+ *
+ * Return:
+ *   ``0`` if creation successful, ``negative value`` otherwise.
+ */
+int8_t z_string_wrap(z_owned_string_t *str, const char *value);
```
milyin, line 1146: there are some concerns about naming - see https://github.com/eclipse-zenoh/zenoh-c/issues/475
Maybe it's good moment to find some solution and then implement it on zenoh-c side

milyin (changes requested):

milyin (approved):

src/api/api.c
```diff
@@ -837,6 +837,20 @@ const z_loaned_encoding_t *z_reply_err_encoding(const z_loaned_reply_err_t *repl
...
 
+int8_t z_string_from_str(z_owned_string_t *str, const char *value) {
+    str->_val = _z_string_make(value);
+    return _Z_RES_OK;
+}
+
+void z_string_empty(z_owned_string_t *str) { str->_val = _z_string_null(); }
+
+int8_t z_string_from_substr(z_owned_string_t *str, const char *value, size_t len) {
+    str->_val = _z_string_n_make(value, len);
```
DenisBiryukov91, line 848: it might be unsafe to call _z_string_n_make with current implementation if you only allowed to touch len characters, since it will copy len+1 and thus may technically result in segfault.

DenisBiryukov91 (approved):

PR #499, opened 2024-07-03
Add z_timestamp_t methods

sashacmc:
- z_timestamp_new
 - z_timestamp_npt64_time
 - z_timestamp_id

Closes: https://github.com/eclipse-zenoh/zenoh-pico/pull/499


milyin:
This should be updated after finishing https://github.com/eclipse-zenoh/zenoh-c/issues/276
which in turn depends on https://github.com/eclipse-zenoh/zenoh/issues/1179

PR #510, opened 2024-07-05
Add weak references

### LINKED ISSUE #476: Replying to a query with a closed session should not yield an UB.
### Describe the feature

As we removed refcounting on queries, we no longer have the issue of cross-referencing rc that prevented us from refcounting the query session pointer.

A session rc is preferable to a pointer, especially if the query is used out of the callback context and more in line with the rest of the codebase.

jean-roland:
Trying to fix #476, I realised the session reference itself was dropped by the refcount which means the query has no way of knowing if the session is still valid.

A possible solution is to introduce weak references. The query having a session weak references, it would be able to check the session validity before the whole rc container is dropped.

Bonus: Working on this, I realized the `z_clone` was not working properly and there was a consolidation mode edge case when receiving replies.


include/zenoh-pico/collections/refcount.h
```diff
@@ -38,122 +41,292 @@
...
+        if (p.in != NULL) {                                                                     \
+            p.in->val = val;                                                                    \
+            _ZP_RC_OP_INIT_CNT                                                                  \
+        }                                                                                       \
+        return p;                                                                               \
+    }                                                                                           \
+    static inline name##_rc_t name##_rc_clone(const name##_rc_t *p) {                           \
+        name##_rc_t c;                                                                          \
+        unsigned int cmp_val = UINT32_MAX;                                                      \
+        if (_ZP_RC_OP_CHECK_WEAK_CNT) {                                                         \
```
DenisBiryukov91, line 195: this check will not work in mt scenario since in case of simultaneous clones of 2 orcs, both of them can pass this check before increasing weak count in the end of the function.

include/zenoh-pico/collections/refcount.h
```diff
@@ -38,122 +41,292 @@
...
+        _ZP_RC_OP_INCR_STRONG_CNT                                                               \
+        _ZP_RC_OP_INCR_WEAK_CNT                                                                 \
+        return c;                                                                               \
+    }                                                                                           \
+    static inline name##_rc_t *name##_rc_clone_as_ptr(const name##_rc_t *p) {                   \
+        unsigned int cmp_val = UINT32_MAX;                                                      \
+        if (_ZP_RC_OP_CHECK_WEAK_CNT) {                                                         \
+            return NULL;                                                                        \
+        }                                                                                       \
+        name##_rc_t *c = (name##_rc_t *)z_malloc(sizeof(name##_rc_t));                          \
```
DenisBiryukov91, line 209: should we account for malloc failure ?

include/zenoh-pico/collections/refcount.h
```diff
@@ -38,122 +41,292 @@
...
+#define _ZP_RC_OP_INIT_CNT                                     \
+    __sync_fetch_and_and(&p.in->_strong_cnt, (unsigned int)0); \
+    __sync_fetch_and_add(&p.in->_strong_cnt, (unsigned int)1); \
+    __sync_fetch_and_and(&p.in->_weak_cnt, (unsigned int)0);   \
+    __sync_fetch_and_add(&p.in->_weak_cnt, (unsigned int)1);
+#define _ZP_RC_OP_INCR_STRONG_CNT __sync_fetch_and_add(&p->in->_strong_cnt, (unsigned int)1);
+#define _ZP_RC_OP_INCR_WEAK_CNT __sync_fetch_and_add(&p->in->_weak_cnt, (unsigned int)1);
+#define _ZP_RC_OP_DECR_AND_CMP_STRONG __sync_fetch_and_sub(&p->in->_strong_cnt, (unsigned int)1) > (unsigned int)1
+#define _ZP_RC_OP_DECR_AND_CMP_WEAK __sync_fetch_and_sub(&p->in->_weak_cnt, (unsigned int)1) > (unsigned int)1
+#define _ZP_RC_OP_CHECK_STRONG_CNT __sync_bool_compare_and_swap(&p->in->_strong_cnt, cmp_val, cmp_val)
```
DenisBiryukov91, line 96: maybe it would be better to allow macros to accept cmp_val as an argument ? It would likely increase code readability.

include/zenoh-pico/protocol/definitions/network.h
```diff
@@ -78,7 +78,7 @@ typedef _z_qos_t _z_n_qos_t;
 static inline _z_qos_t _z_n_qos_create(_Bool express, z_congestion_control_t congestion_control,
                                        z_priority_t priority) {
     _z_n_qos_t ret;
-    _Bool nodrop = congestion_control == Z_CONGESTION_CONTROL_DROP ? 0 : 1;
+    _Bool nodrop = (congestion_control == Z_CONGESTION_CONTROL_DROP ? false : true);
```
DenisBiryukov91, line 81: congestion_control != Z_CONGESTION_CONTROL_DROP  ?

src/net/primitives.c
```diff
@@ -314,12 +314,13 @@ int8_t _z_send_reply(const _z_query_t *query, _z_keyexpr_t keyexpr, const _z_val
                      const z_sample_kind_t kind, const z_congestion_control_t cong_ctrl, z_priority_t priority,
                      _Bool is_express, const _z_timestamp_t *timestamp, const _z_bytes_t att) {
     int8_t ret = _Z_RES_OK;
+    _z_session_t *zn = &query->_zn.in->val;
```
DenisBiryukov91, line 317: should this one also use upgrade instead of getting the pointer directly

jean-roland: I think the pointer should be upgraded in the api layer, before this call so we can bail quickly if it fails.

DenisBiryukov91: then strong rc should be passed explicitly, otherwise when reading the code it is unclear what guarantees that reading ptr from weak is ok.

DenisBiryukov91 (approved):

PR #530, opened 2024-07-11
Use session to create timestamp

### LINKED ISSUE #502: timestamp should be taken from session
### Describe the release item

Update `new_timestamp` accordingly to https://github.com/eclipse-zenoh/zenoh-c/issues/276
which in turn depends on https://github.com/eclipse-zenoh/zenoh/issues/1179

jean-roland:
Closes #502.

This PR revealed an underlying issue regarding timestamps: https://github.com/eclipse-zenoh/zenoh/issues/1239

There might be more work required to make timestamps work on pico.

### DIFF
--- a/src/session/utils.c
+++ b/src/session/utils.c
@@ -39,8 +39,11 @@ void _z_timestamp_clear(_z_timestamp_t *tstamp) {
 }
 
 _Bool _z_timestamp_check(const _z_timestamp_t *stamp) {
-    for (uint8_t i = 0; i < sizeof(_z_id_t); ++i)
-        if (stamp->id.id[i]) return true;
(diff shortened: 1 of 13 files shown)


src/session/utils.c
```diff
@@ -39,8 +39,11 @@ void _z_timestamp_clear(_z_timestamp_t *tstamp) {
 }
 
 _Bool _z_timestamp_check(const _z_timestamp_t *stamp) {
-    for (uint8_t i = 0; i < sizeof(_z_id_t); ++i)
-        if (stamp->id.id[i]) return true;
+    for (uint8_t i = 0; i < sizeof(_z_id_t); ++i) {
```
DenisBiryukov91, line 42: I believe it would be better to use here dedicated _z_id_check.

Issue #546, opened 2024-07-16
[Bug] 0.11.0 example crashes Zephyr 


ypearson-bdai:
### Describe the bug

Using the zenoh-pico example`z_pub.c` crashes with the following 

```[00:00:02.011,000] <inf> net_config: Interface 1 (0x20001d74) coming up
[00:00:02.012,000] <inf> net_config: IPv4 address: 192.168.11.2
[00:00:02.112,000] <inf> net_config: IPv6 address: 2001:db8::2
[00:00:02.112,000] <inf> net_config: IPv6 address: 2001:db8::2
uart:~$ Opening Zenoh Session...Unable to open session!
[00:00:10.111,000] <err> pthread_mutex: Mutex is uninitialized (0)
[00:00:10.111,000] <err> pthread_mutex: Mutex is uninitialized (0)
[00:00:10.111,000] <err> pthread_mutex: Mutex is uninitialized (0)
[00:00:10.111,000] <err> pthread_mutex: Mutex is uninitialized (0)
[00:00:10.111,000] <err> pthread_mutex: Mutex is uninitialized (0)
[00:00:10.111,000] <err> pthread_mutex: Mutex is uninitialized (0)
[00:00:10.111,000] <err> pthread_mutex: Mutex is uninitialized (0)
[00:00:10.111,000] <err> pthread_mutex: Mutex is uninitialized (0)
[00:00:10.111,000] <err> pthread_mutex: Mutex is uninitialized (0)
[00:00:10.111,000] <err> pthread_mutex: Mutex is uninitialized (0)
[00:00:10.111,000] <err> pthread_mutex: Mutex is uninitialized (0)
[00:00:10.116,000] <err> os: ***** USAGE FAULT *****
[00:00:10.116,000] <err> os:   Illegal use of the EPSR
[00:00:10.116,000] <err> os: r0/a1:  0x00000000  r1/a2:  0x00000000  r2/a3:  0x2000798c
[00:00:10.116,000] <err> os: r3/a4:  0x00000000 r12/ip:  0x00000030 r14/lr:  0x0800636d
[00:00:10.116,000] <err> os:  xpsr:  0x000f0000
[00:00:10.116,000] <err> os: Faulting instruction address (r15/pc): 0x00000000
[00:00:10.116,000] <err> os: >>> ZEPHYR FATAL ERROR 35: Unknown error on CPU 0
[00:00:10.116,000] <err> os: Current thread: 0x20003208 (main)
[...]

PR #551, opened 2024-07-19
Add mutex_free value guard for Zephyr

jean-roland:
Should fix the hard fault part of #546.

### DIFF
--- a/src/system/zephyr/system.c
+++ b/src/system/zephyr/system.c
@@ -96,7 +96,12 @@ void z_task_free(z_task_t **task) {
 /*------------------ Mutex ------------------*/
 int8_t z_mutex_init(z_mutex_t *m) { return pthread_mutex_init(m, 0); }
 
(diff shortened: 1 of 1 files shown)


DenisBiryukov91 (approved):

src/system/zephyr/system.c
```diff
@@ -96,7 +96,12 @@ void z_task_free(z_task_t **task) {
 /*------------------ Mutex ------------------*/
 int8_t z_mutex_init(z_mutex_t *m) { return pthread_mutex_init(m, 0); }
 
-int8_t z_mutex_free(z_mutex_t *m) { return pthread_mutex_destroy(m); }
+int8_t z_mutex_free(z_mutex_t *m) {
+    if (*m == NULL) {
```
ypearson, line 100: Do you really mean `if(m == NULL)` ?

jean-roland: No in this case the pointer is fine as it's the address of a struct field so only the value should cause problem.

Although I could have add a (m==NULL) too.

PR #563, opened 2024-07-29
Allow providing custom delete function when serializing from pointers

DenisBiryukov91:
z_bytes_serialize_from_slice now takes z_loaned_slice_t;
introduce z_bytes_serialize_from_string taking z_loaned_string_t;
introduce z_bytes_from_slice which consumes z_owned_slice_t;
z_bytes_serialize_from_str is renamed into z_bytes_from_string and is now consuming z_owned_string_t;
introduce z_bytes_from_buf/str allowing to consume raw pointers by taking custom delete function;
introduce z_bytes_serialize_from_buf/str performing serialization of pointers by copying;
introduce z_bytes_from_static_buf/str performing serialization of pointers by aliasing;

implement _z_string_t in terms of _z_slice_t;
renamed z_string_from_str/substr into z_string_copy_from_str/substr;
renamed z_slice_from_buf into z_slice_copy_from_buf;
add z_string_from_str now accepts allocated null-terminated string with custom deleter;
add z_slice_from_buf now accepts allocated data with custom deleter;

### DIFF
--- a/examples/freertos_plus_tcp/z_put.c
+++ b/examples/freertos_plus_tcp/z_put.c
@@ -67,7 +67,7 @@ void app_main(void) {
 
     // Create payload
     z_owned_bytes_t payload;
-    z_bytes_serialize_from_str(&payload, VALUE);
+    z_bytes_serialize_from_str(&payload, VALUE, NULL, NULL);
 
     if (z_put(z_loan(s), z_loan(ke), z_move(payload), &options) < 0) {
         printf("Oh no! Put has failed...\n");
--- a/examples/freertos_plus_tcp/z_put.c
+++ b/examples/freertos_plus_tcp/z_put.c
@@ -67,7 +67,7 @@ void app_main(void) {
 
     // Create payload
     z_owned_bytes_t payload;
-    z_bytes_from_str(&payload, VALUE, (char *)NULL, NULL);
+    z_bytes_from_str(&payload, (char *)VALUE, NULL, NULL);
 
     if (z_put(z_loan(s), z_loan(ke), z_move(payload), &options) < 0) {
         printf("Oh no! Put has failed...\n");
--- a/examples/freertos_plus_tcp/z_put.c
+++ b/examples/freertos_plus_tcp/z_put.c
@@ -67,7 +67,7 @@ void app_main(void) {
 
     // Create payload
     z_owned_bytes_t payload;
-    z_bytes_serialize_from_str(&payload, VALUE, NULL, NULL);
+    z_bytes_from_str(&payload, VALUE, (char *)NULL, NULL);
 
     if (z_put(z_loan(s), z_loan(ke), z_move(payload), &options) < 0) {
         printf("Oh no! Put has failed...\n");
--- a/examples/freertos_plus_tcp/z_put.c
+++ b/examples/freertos_plus_tcp/z_put.c
@@ -67,7 +67,7 @@ void app_main(void) {
 
     // Create payload
     z_owned_bytes_t payload;
-    z_bytes_from_str(&payload, (char *)VALUE, NULL, NULL);
+    z_bytes_from_static_str(&payload, VALUE);
 
     if (z_put(z_loan(s), z_loan(ke), z_move(payload), &options) < 0) {
         printf("Oh no! Put has failed...\n");
--- a/include/zenoh-pico/api/primitives.h
+++ b/include/zenoh-pico/api/primitives.h
@@ -736,6 +736,18 @@ int8_t z_bytes_serialize_from_float(z_owned_bytes_t *bytes, float val);
 int8_t z_bytes_serialize_from_double(z_owned_bytes_t *bytes, double val);
 
 /**
+ * Encodes a slice into a :c:type:`z_owned_bytes_t` by copying.
+ *
+ * Parameters:
+ *   bytes: An uninitialized :c:type:`z_owned_bytes_t` to contain the encoded slice.
+ *   slice: Pointer to the slice to encode.
+ *
+ * Return:
+ *   ``0`` if encode successful, ``negative value`` otherwise.
+ */
+int8_t z_bytes_serialize_from_slice(z_owned_bytes_t *bytes, const z_loaned_slice_t *slice);
+
+/**
  * Encodes a slice into a :c:type:`z_owned_bytes_t`.
  *
  * Parameters:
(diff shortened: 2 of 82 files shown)


examples/freertos_plus_tcp/z_put.c
```diff
@@ -67,7 +67,7 @@ void app_main(void) {
 
     // Create payload
     z_owned_bytes_t payload;
-    z_bytes_serialize_from_str(&payload, VALUE);
+    z_bytes_from_str(&payload, (char *)VALUE, NULL, NULL);
```
milyin, line 70: I believe we can accept `const char*` here. This is very typical case when we need to send string without providing destructor (static one or allocated on command line parsing. Also payload is not supposed to be able to change the content of the string, it only takes ownership on it. So I see no reason to force user to write `(char*)`every time

milyin (changes requested):

examples/freertos_plus_tcp/z_put.c
milyin: It's a good point that "free" can't be called on const pointer.
Then it makes even more sense to have 2 functions:
```
z_bytes_from_static_str(z_owned_payload_t* payload, const char* str);
z_bytes_from_str(z_owned_payload_t* payload, char* str,  void (*deleter)(void *data, void *context),
 void *context);
```
and maybe even disallow to call second one with empty destructor

And do the same for sized buffer of course

Mallets: I prefer to be explicit in this case and piggy-backing on any warning the compiler would give about discarding any `const` qualifier.

include/zenoh-pico/api/primitives.h
```diff
@@ -769,19 +829,56 @@ int8_t z_bytes_serialize_from_slice_copy(z_owned_bytes_t *bytes, const uint8_t *
...
+ */
+int8_t z_bytes_from_str(z_owned_bytes_t *bytes, char *value, void (*deleter)(void *data, void *context), void *context);
 
 /**
- * Encodes a string into a :c:type:`z_owned_bytes_t` by copying
+ * Encodes a statically allocated constant null-terminated string into a :c:type:`z_owned_bytes_t` by aliasing.
+ *
+ * Parameters:
+ *   bytes: An uninitialized :c:type:`z_owned_bytes_t` to contain the encoded string.
+ *   value: Pointer to the the statically allocated constant string to encode.
```
milyin, line 865: This function doesn't require statically allocated string anymore

src/api/api.c
```diff
@@ -413,6 +406,21 @@ int8_t z_encoding_to_string(const z_loaned_encoding_t *encoding, z_owned_string_
     return _Z_RES_OK;
 }
 
+int8_t z_slice_from_buf(z_owned_slice_t *slice, const uint8_t *data, size_t len) {
```
milyin, line 409: function with similar name `z_bytes_from_str` takes ownersip of the buffer and accepts deleter and context. I think by analogy the `z_slice_wrap` should be named `z_slice_from_buf` and this `z_slice_from_buf` should be renamed to e.g. `z_slice_copy_from_buf` (of course we can't use word "serialize" here, it's reserved to payload).

I.e. all function names should follow the same logic:
- `z_xxx_from_yyy` takes ownership of it's parameter, providing deleter if necessary
- `z_xxx_copy_from_yyy` and `z_xxx_serialize_from_yyy` keeps parameter available to the caller

So all `z_xxx_wrap` frunctions should be renamed to `z_xxx_from_yyy`

milyin (changes requested):

include/zenoh-pico/api/primitives.h
```diff
@@ -436,6 +423,36 @@ const z_loaned_bytes_t *z_reply_err_payload(const z_loaned_reply_err_t *reply_er
...
+ *   data: Pointer to the data to be owned by `slice`.
+ *   len: Number of bytes in `data`.
+ *  deleter: A thread-safe delete function to free the `data`. Will be called once when `slice` is dropped. Can be
+ * NULL, in case if `data` is allocated in static memory.
+ *    context: An optional context to be passed to the `deleter`.
+ *
+ * Return:
+ *   ``0`` if creation successful, ``negative value`` otherwise.
+ */
+int8_t z_slice_wrap(z_owned_slice_t *slice, uint8_t *data, size_t len, void (*deleter)(void *data, void *context),
```
milyin, line 453: The parameter have type `uint8_t*' but deleter accepts `void*`. This makes not obvious that deleter accepts the same data which is passed to the function. Maybe deleters should accept the same type which is passed to function.

Counterargument may be that this makes harder to make common deleter for all functions. I agree with that. One solution, which may also improve the usability could be like this:
```C
z_payload_from_str(z_owned_payload_t* payload, char* str, void (*free)(void* str));
z_payload_from_str_with_ctx(z_owned_payload_t* payload, char* str, void (*free)(char* str, void* ctx), void* ctx);
z_payload_from_static_str(z_owned_payload_t* payload, const char* str);
```
This will allow user in most cases simply call
```C
z_owned_payload_t payload;
z_payload_from_str(&payload, strdup("foo"), free);
```
without creating any additonal release function.

And if user have some more complex allocators, he have "_ctx" functions for it accepting deleters, which should be implemented manually anyway.

milyin (changes requested):

include/zenoh-pico/api/primitives.h
DenisBiryukov91: If we accept typed deleter we will have to always wrap it in void* deleter (since strings are implemented in terms of slice in both zenoh-c and pico), which is an unnecessary extra cost.
If the user use strdup, they could simply call copy-making serialization function. Otherwise there is no common free function in C (even the one used in zenoh-c is likely different from standard free since it uses rust Box allocation/ deallocation), so I do not think it is makes much sense to bloat the api with functions that would likely lead to errors and a lot confusion and frustration (it is pretty difficult to debug the fact that wrong free is being called).

include/zenoh-pico/api/primitives.h
milyin: Sorry, seems I mistakenly thought that this is a comment for `z_bytes_from_str`

src/api/api.c
DenisBiryukov91: fixed

milyin (approved):

PR #568, opened 2024-08-01
allow specifying default encoding for publisher;

DenisBiryukov91:
allow specifying default encoding for publisher;
put clean up;

### DIFF
--- a/zenohpico.pc
+++ b/zenohpico.pc
@@ -3,6 +3,6 @@ prefix=/usr/local
 Name: zenohpico
 Description: 
 URL: 
-Version: 1.0.20240730dev
+Version: 1.0.20240801dev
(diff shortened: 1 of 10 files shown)


zenohpico.pc
```diff
@@ -1,8 +1,8 @@
-prefix=/usr/local
+prefix=/Users/denisbiryukov/repositories/work/local
```
sashacmc, line 1: Please revert this change

sashacmc (approved):
Small note, others LGTM

zenohpico.pc
DenisBiryukov91: fixed

milyin (approved):

PR #573, opened 2024-08-05
z_recv/z_try_recv without the need of z_check

DenisBiryukov91:
make z_recv/z_try_recv return z_result_t to allow differentiating between different cases without the need to call z_check on constructed object.


sashacmc (approved):
LGTM!

milyin (approved):

PR #574, opened 2024-08-06
Fix vle serialization

### LINKED ISSUE #564: [Bug] zenoh-pico dev/1.0.0 does not implement the same serializer as zenoh dev/1.0.0
### Describe the bug

I take as an example `uint64` serialization implemented [here](https://github.com/eclipse-zenoh/zenoh-pico/blob/a07ee806e809eb9cb69c630910fc98834a7c1782/src/api/api.c#L516C12-L516C32).
It seems that the actual serialization [implementation](https://github.com/eclipse-zenoh/zenoh-pico/blob/a07ee806e809eb9cb69c630910fc98834a7c1782/src/utils/endianness.c#L50) is simply copying bytes without stripping zeroed leading bytes as done in [Rust](https://github.com/eclipse-zenoh/zenoh/blob/86f0848b6fb5d950d50429644e95fe4b71b21a19/zenoh/src/api/bytes.rs#L1151).

While this still allows to have successful communication between zenoh-pico and zenoh, transmission on zenoh-pico is not as compact as it could be since meaningless bytes are transmitted on the network.

### To reproduce

N.A.

### System info

N.A.

jean-roland:
Closes #564

### DIFF
--- a/src/utils/endianness.c
+++ b/src/utils/endianness.c
@@ -35,19 +35,29 @@ uint64_t _z_le_load64(const uint8_t *src) {
            ((uint64_t)src[4] << 32) | ((uint64_t)src[5] << 40) | ((uint64_t)src[6] << 48) | ((uint64_t)src[7] << 56);
 }
 
-void _z_le_store16(const uint16_t val, uint8_t *dst) {
-    dst[0] = (uint8_t)(val >> 0);
-    dst[1] = (uint8_t)(val >> 8);
+static size_t _z_le_calc_vle_len(size_t size, uint8_t *buff) {
+    size_t vle_len = size;
+    while (vle_len > 1 && buff[vle_len - 1] == 0) {
+        --vle_len;
+    }
+    return vle_len;
 }
 
-void _z_le_store32(const uint32_t val, uint8_t *dst) {
+size_t _z_le_store16(const uint16_t val, uint8_t *dst) {
+    dst[0] = (uint8_t)(val >> 0);
+    dst[1] = (uint8_t)(val >> 8);
+    return _z_le_calc_vle_len(2, dst);
+}
+
+size_t _z_le_store32(const uint32_t val, uint8_t *dst) {
     dst[0] = (uint8_t)(val >> 0);
     dst[1] = (uint8_t)(val >> 8);
     dst[2] = (uint8_t)(val >> 16);
     dst[3] = (uint8_t)(val >> 24);
+    return _z_le_calc_vle_len(4, dst);
 }
 
-void _z_le_store64(const uint64_t val, uint8_t *dst) {
+size_t _z_le_store64(const uint64_t val, uint8_t *dst) {
     dst[0] = (uint8_t)(val >> 0);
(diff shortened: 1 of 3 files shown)


src/utils/endianness.c
```diff
@@ -35,19 +35,29 @@ uint64_t _z_le_load64(const uint8_t *src) {
...
+
+size_t _z_le_store16(const uint16_t val, uint8_t *dst) {
     dst[0] = (uint8_t)(val >> 0);
     dst[1] = (uint8_t)(val >> 8);
+    return _z_le_calc_vle_len(2, dst);
 }
 
-void _z_le_store32(const uint32_t val, uint8_t *dst) {
+size_t _z_le_store32(const uint32_t val, uint8_t *dst) {
     dst[0] = (uint8_t)(val >> 0);
```
DenisBiryukov91, line 53: if you are anyway looping over the bytes to count vle length. Maybe it is just simpler just too make bit shifts in a loop and terminate early once val turns to 0 ?
```
for (size_t i = 0; i < sizeof(val); ++i) {
    dst[i] = (uint8_t)val;
    val = val >> 8;
    if (val == 0) return i + 1;
}

This also can be done as a single macro and probably slightly reduce amount of code
```

jean-roland: I think a macro would make it less readable in this case. 

Now for combining the loop, I don't know, not that performance really matter but it might be a `strlcpy` situation (https://nrk.neocities.org/articles/cpu-vs-common-sense)

DenisBiryukov91: I'm not sure strlcpy is related since you'll have to loop once in both cases. And given that pico targets low-end controllers it is more likely that less code packed in a loop will have better performance than an unrolled loop given that vector registers might not be available at all.

DenisBiryukov91: I suppose to give equal chances we can still copy the whole uint, but check for last non-zero byte in the same loop. It will produce the similar code on smart-enough compiler (if it can optimize checking for last non-zero byte, it should likely be able to optimize a copy of uint), but still keeps the benefit of having smaller code size.

src/utils/endianness.c
```diff
@@ -35,27 +35,43 @@ uint64_t _z_le_load64(const uint8_t *src) {
...
-void _z_le_store64(const uint64_t val, uint8_t *dst) {
-    dst[0] = (uint8_t)(val >> 0);
-    dst[1] = (uint8_t)(val >> 8);
-    dst[2] = (uint8_t)(val >> 16);
-    dst[3] = (uint8_t)(val >> 24);
-    dst[4] = (uint8_t)(val >> 32);
-    dst[5] = (uint8_t)(val >> 40);
-    dst[6] = (uint8_t)(val >> 48);
-    dst[7] = (uint8_t)(val >> 56);
+size_t _z_le_store16(const uint16_t val, uint8_t *dst) {
```
DenisBiryukov91, line 38: there is no need to use const, the data is copied anyway

DenisBiryukov91 (approved):

Mallets (approved):

Issue #555, opened 2024-07-23
Add moved types


jean-roland:
### Describe the feature

The limitation with the current owned types and its move semantic is that there is no differentiation on function signature, and no obligation / no way to know when to use `z_move` unless you're copying from example code.

Moved types would fix this, see https://github.com/eclipse-zenoh/zenoh-c/pull/488

PR #575, opened 2024-08-06
Move protection

milyin:
added `z_move_xxx_t` types which should be used when owned object shoud be consumed by function
Same API as in https://github.com/eclipse-zenoh/zenoh-c/pull/488


jean-roland:
Closes #555

jean-roland (approved):

PR #582, opened 2024-08-08
Reduce default buffer sizes

jean-roland:
Default buffer sizes was causing session open errors on microcontrollers. Now, the goal of this PR is to set those values so that most MCU will be fine with them while still maintaining the bigger sizes for unix systems.

Also add a troubleshooting section to the README documenting the memory issue if it still occurs in the smaller systems.

Had to remove the values from the zenoh-pico.h as it wasn't working properly.


DenisBiryukov91:
It is strange that cmakedefine does not work. Alternatively  we can set Z_FRAG_MAX_SIZE to default value in cmake  and keep corresponding defines in zenoh-pico.h.in and allow user to override it by setting directly -DZ_FRAG_MAX_SIZE=MYVALUE when invoking cmake

DenisBiryukov91 (approved):

Mallets (approved):

Issue #589, opened 2024-08-13
Mark z_check/z_null as internal methods 


sashacmc:
### Describe the feature

z_check should be renamed to the _z_check
z_null should be renamed to the _z_null

We would like to use them internally (e.g. for zenoh-cpp) but don't expose to the API

Related to: https://github.com/eclipse-zenoh/zenoh-c/issues/492

PR #592, opened 2024-08-13
Make z_null and z_check as internal method

sashacmc:
- `z_null` renamed to `_z_null` and removed from public API
- `z_check` renamed to `_z_check` and removed from public API

Closes: https://github.com/eclipse-zenoh/zenoh-pico/issues/589


sashacmc:
It was not possible to just rename `z_type_null` to `_z_type_null` because of conflicts with the internal methods, so it was renamed to `_z_owned_type_null`

milyin (approved):

PR #596, opened 2024-08-15
Fix and simplify cmake package config

bjsowa:
The current package config template seems to be copied from zenoh-c and does not really work as the CMakeLists does not create STATICLIB or DYLIB variables. 

The solution I propose is much simpler. It just exports the `zenohpico` target which contains the public include directories and compile definitions, instead of creating an imported library target.

Dependent projects can just `find_package` this project and link `zenohpico::zenohpico` target to their executable.


CMakeLists.txt
```diff
@@ -66,11 +66,7 @@ elseif(CMAKE_SYSTEM_NAME MATCHES "Generic")
 endif()
 
 set(Libname "zenohpico")
-if(BUILD_SHARED_LIBS)
-  add_library(${Libname} SHARED)
-else()
-  add_library(${Libname} STATIC)
-endif()
+add_library(${Libname})
```
bjsowa, line 69: This was just redundant as `add_library` already does this:
> If no <type> is given the default is STATIC or SHARED based on the value of the [BUILD_SHARED_LIBS](https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html#variable:BUILD_SHARED_LIBS) variable.

CMakeLists.txt
```diff
@@ -237,7 +233,10 @@ file(GLOB_RECURSE PublicHeaders
   "include/zenoh-pico/utils/*.h"
   "include/zenoh-pico/config.h"
 )
-target_include_directories(${Libname} PUBLIC ${PROJECT_SOURCE_DIR}/include)
+target_include_directories(${Libname}
+  PUBLIC
+    $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>
+    $<INSTALL_INTERFACE:include>)
```
bjsowa, line 239: The included directory is different for installed target

examples/CMakeLists.txt
```diff
@@ -4,7 +4,7 @@ if(${CMAKE_SOURCE_DIR} STREQUAL ${CMAKE_CURRENT_SOURCE_DIR})
     project(zenohpico_examples LANGUAGES C)
     include(../cmake/helpers.cmake)
     set_default_build_type(Release)
-    configure_include_project(ZENOHPICO zenohpico zenohpico::static ".." zenohpico "https://github.com/eclipse-zenoh/zenoh-pico" "")
+    configure_include_project(ZENOHPICO zenohpico zenohpico ".." zenohpico "https://github.com/eclipse-zenoh/zenoh-pico" "")
```
bjsowa, line 7: This actually fixes the build if building the examples as the root project. When including the project using `add_subdirectory` or `FetchContent` the library target will be named `zenohpico`. 
The only small problem is that the target name for a `find_packaged` library will be `zenohpico::zenohpico` and this macro does not take the target namespace into account.

jean-roland:
Hello @bjsowa and thanks for your contribution, we'll look into it after the end of our release cycle so probably next week.

DenisBiryukov91 (approved):

DenisBiryukov91:
@bjsowa We still want to expose variables corresponding to enabled/disabled features of pico (this is needed mostly for cpp to decide which tests/examples to build). I made corresponding pr to your branch.

DenisBiryukov91 (changes requested):
Consider including changes from https://github.com/bjsowa/zenoh-pico/pull/1

examples/CMakeLists.txt
```diff
@@ -4,7 +4,7 @@ if(${CMAKE_SOURCE_DIR} STREQUAL ${CMAKE_CURRENT_SOURCE_DIR})
     project(zenohpico_examples LANGUAGES C)
     include(../cmake/helpers.cmake)
     set_default_build_type(Release)
-    configure_include_project(ZENOHPICO zenohpico zenohpico::static ".." zenohpico "https://github.com/eclipse-zenoh/zenoh-pico" "")
+    configure_include_project(ZENOHPICO zenohpico zenohpico::lib ".." zenohpico "https://github.com/eclipse-zenoh/zenoh-pico" "")
```
bjsowa, line 7: This should now work regardless of whether zenohpico is included using `find_package` or `add_subdirectory` as both define `zenohpico::lib` target

CMakeLists.txt
```diff
@@ -108,11 +108,8 @@ else()
 endif()
 
 set(Libname "zenohpico")
-if(BUILD_SHARED_LIBS)
-  add_library(${Libname} SHARED)
-else()
-  add_library(${Libname} STATIC)
-endif()
+add_library(${Libname})
+add_library(zenohpico::lib ALIAS ${Libname})
```
bjsowa, line 112: @DenisBiryukov91 I assume there was no reason to put the alias definition inside `add_definition` function. Let me know if I'm wrong.

DenisBiryukov91: Yes, sorry, this was a typo.

DenisBiryukov91 (approved):

PR #597, opened 2024-08-19
switch to ptr to z_xxx_moved_t

milyin:
update accordingly to https://github.com/eclipse-zenoh/zenoh-c/pull/598

### DIFF
--- a/include/zenoh-pico/config.h
+++ b/include/zenoh-pico/config.h
@@ -16,8 +16,8 @@
 #define INCLUDE_ZENOH_PICO_CONFIG_H
 
 /*--- CMake generated config; pass values to CMake to change the following tokens ---*/
-#define Z_FRAG_MAX_SIZE 0
-#define Z_BATCH_UNICAST_SIZE 0
+#define Z_FRAG_MAX_SIZE 4096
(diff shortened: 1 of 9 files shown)


include/zenoh-pico/config.h
```diff
@@ -16,9 +16,9 @@
 #define INCLUDE_ZENOH_PICO_CONFIG_H
 
 /*--- CMake generated config; pass values to CMake to change the following tokens ---*/
-#define Z_FRAG_MAX_SIZE 300000
-#define Z_BATCH_UNICAST_SIZE 65535
-#define Z_BATCH_MULTICAST_SIZE 8096
+#define Z_FRAG_MAX_SIZE 0
+#define Z_BATCH_UNICAST_SIZE 0
+#define Z_BATCH_MULTICAST_SIZE 2048
```
jean-roland, line 21: This is auto generated so it doesn't matter much but how did these values went to 0?

jean-roland (approved):

include/zenoh-pico/config.h
milyin: After clean rebuild the values were restored to adequate. Maybe this is caused by some previous unsuccessful builds

Issue #621, opened 2024-09-04
Implement liveliness token support


sashacmc:
### Describe the release item

Description 
https://github.com/eclipse-zenoh/roadmap/blob/main/rfcs/ALL/Liveliness.md

Related PR/Issues
https://github.com/eclipse-zenoh/zenoh-c/pull/189
https://github.com/eclipse-zenoh/zenoh/issues/705

PR #632, opened 2024-09-06
Implement liveliness support

sashacmc:
Implement liveliness support, can be enabled/disabled by means of `Z_FEATURE_LIVELINESS`

- Token:
  - z_liveliness_declare_token
  - z_liveliness_undeclare_token
 
- Subscriber: (require `Z_FEATURE_SUBSCRIPTION` enabled)
  - z_liveliness_declare_subscriber

- Query: (require `Z_FEATURE_QUERY` enabled)
  - z_liveliness_get

- Token interest query processing

Examples:
 - z_sub_liveliness
 - z_get_liveliness
 - z_liveliness

Additionally:
- Implemented iterator for _z_int_void_map
- Cleaned up (reused) stuff from pull subscribers ( _remote_subscriptions)

Closes: https://github.com/eclipse-zenoh/zenoh-pico/issues/621


include/zenoh-pico/net/primitives.h
```diff
@@ -76,6 +75,18 @@ uint16_t _z_declare_resource(_z_session_t *zn, _z_keyexpr_t keyexpr);
  */
 z_result_t _z_undeclare_resource(_z_session_t *zn, uint16_t rid);
 
+/**
+ * Decalre keyexpr if it is necessary and allowed.
```
jean-roland, line 79:
```suggestion
 * Declare keyexpr if it is necessary and allowed.
```

src/session/liveliness.c
```diff
@@ -0,0 +1,340 @@
...
+            _z_reply_t reply = _z_reply_create(expanded_ke, zn->_local_zid, _z_bytes_null(), timestamp, &encoding,
+                                               Z_SAMPLE_KIND_PUT, _z_bytes_null());
+
+            pq->_callback(&reply, pq->_arg);
+            _z_reply_clear(&reply);
+        }
+    }
+
+    if (ret == _Z_RES_OK) {
+        _z_liveliness_pending_query_intmap_remove(&zn->_liveliness_pending_queries, interest_id);
```
OlivierHecart, line 230: If we remove the pending query from the list here, we won't be able to process next DeclareTokens, right ?

OlivierHecart (changes requested):

OlivierHecart:
Another general comment:
As far as I understood, disconnection is not really managed yet run pico. But theoretically (and in Rust) when connectivity is lost, all remote tokens should be undeclared: an active liveliness subscriber should receive DEL messages for all alive remote tokens, and a liveliness get performed after the disconnection should get no replies.
Not sure if we want to handle that in this PR or another one.

examples/unix/c11/z_liveliness.c
```diff
@@ -0,0 +1,121 @@
...
+#include <stdio.h>
+#include <stdlib.h>
+#include <unistd.h>
+#include <zenoh-pico.h>
+
+#if Z_FEATURE_LIVELINESS == 1
+
+static volatile int keepRunning = 1;
+
+void intHandler(int dummy) {
```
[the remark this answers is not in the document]
sashacmc: Without this, neither session closing nor token undeclaration is called at all, the example with zenoh-pico is simply interrupted. Yes, after the session loss timeout, the router will report the loss of the token, but for example in multicast peer-to-peer this will not happen.

sashacmc:
> Another general comment: As far as I understood, disconnection is not really managed yet run pico. But theoretically (and in Rust) when connectivity is lost, all remote tokens should be undeclared: an active liveliness subscriber should receive DEL messages for all alive remote tokens, and a liveliness get performed after the disconnection should get no replies. Not sure if we want to handle that in this PR or another one.

Yes, make sense, I created an issue to don't forget do it after I rework disconnection handling in https://github.com/eclipse-zenoh/zenoh-pico/issues/767

src/net/liveliness.c
```diff
@@ -0,0 +1,162 @@
...
+    if (sub == NULL || _Z_RC_IS_NULL(&sub->_zn)) {
+        return _Z_ERR_ENTITY_UNKNOWN;
+    }
+
+    _z_subscription_rc_t *s =
+        _z_get_subscription_by_id(_Z_RC_IN_VAL(&sub->_zn), _Z_SUBSCRIBER_KIND_LIVELINESS_SUBSCRIBER, sub->_entity_id);
+    if (s == NULL) {
+        return _Z_ERR_ENTITY_UNKNOWN;
+    }
+
```
OlivierHecart, line 114: Shouldn't we send an Interest<Final> here ?

src/net/liveliness.c
```diff
@@ -0,0 +1,162 @@
...
+    _z_unregister_subscription(_Z_RC_IN_VAL(&sub->_zn), _Z_SUBSCRIBER_KIND_LIVELINESS_SUBSCRIBER, s);
```
OlivierHecart, line 115: Shouldn't we undeclared resource here (as we do for classic subscriber) ?

include/zenoh-pico/api/liveliness.h
```diff
@@ -0,0 +1,153 @@
...
+_Z_OWNED_FUNCTIONS_DEF(liveliness_token)
+
+#if Z_FEATURE_LIVELINESS == 1
+
+/**************** Liveliness Token ****************/
+
+/**
+ * The options for :c:func:`z_liveliness_declare_token()`.
+ */
+typedef struct z_liveliness_token_options_t_t {
```
OlivierHecart, line 44: I guess there is a query/replace error here:
Should be `z_liveliness_token_options_t`and not `z_liveliness_token_options_t_t`.

OlivierHecart, line 44: I fully agree with the rename here. But shouldn't we synchronize with zenoh-c for this?

OlivierHecart (changes requested):

include/zenoh-pico/api/liveliness.h
sashacmc: It was a typo, thanks!

sashacmc: In zenoh-c all liveliness methods currently zc_ (and unstable), we will do rename in zenoh-c after this PR merged.

jean-roland (approved):

include/zenoh-pico/api/liveliness.h
OlivierHecart: ok

src/net/liveliness.c
```diff
@@ -0,0 +1,173 @@
...
+        return _Z_ERR_ENTITY_UNKNOWN;
+    }
+
+    _z_subscription_rc_t *s =
+        _z_get_subscription_by_id(_Z_RC_IN_VAL(&sub->_zn), _Z_SUBSCRIBER_KIND_LIVELINESS_SUBSCRIBER, sub->_entity_id);
+    if (s == NULL) {
+        return _Z_ERR_ENTITY_UNKNOWN;
+    }
+
+    _z_interest_t interest = _z_make_interest(NULL, s->_val->_id, 0);
```
OlivierHecart, line 117: Since we have a `_z_make_interest_final` function, it would be good to use it here.

OlivierHecart (approved):

Issue #640, opened 2024-09-10
Remove legacy z_config_peer/z_config_client methods


sashacmc:
### Describe the release item

Nobody use them and they seems to redundant

PR #641, opened 2024-09-10
Remove legacy z_config_peer/z_config_client methods

sashacmc:
Closes: https://github.com/eclipse-zenoh/zenoh-pico/issues/640


tests/z_api_config_test.c
```diff
@@ -1,49 +0,0 @@
-//
```
DenisBiryukov91, line 1: Could we still have a test verifying that get/set operations work for the config ?

sashacmc: We already have it in z_api_alignment_test.c

milyin (approved):

Issue #639, opened 2024-09-10
Sync clone methods with zenoh-c


sashacmc:
### Describe the release item

Currently in zenoh-pico we have clone method for all owned/loaned types, but for some classes we don't need them and we can't implement them properly in zenoh-c, we should remove clone methods for types that don't have them in zenoh-c.

PR #645, opened 2024-09-10
Remove clone for types which do not involve copying

sashacmc:
Removes:
- z_subscriber_clone
- z_publisher_clone
- z_queryable_clone
- z_session_clone

Closes: https://github.com/eclipse-zenoh/zenoh-pico/issues/639


src/api/api.c
```diff
@@ -770,15 +770,15 @@ _Z_OWNED_FUNCTIONS_VALUE_IMPL(_z_string_t, string, _z_string_check, _z_string_nu
 _Bool _z_value_check(const _z_value_t *value) {
     return _z_encoding_check(&value->encoding) || _z_bytes_check(&value->payload);
 }
-_Z_OWNED_FUNCTIONS_VALUE_IMPL(_z_value_t, reply_err, _z_value_check, _z_value_null, _z_value_copy, _z_value_clear)
+_Z_OWNED_FUNCTIONS_VALUE_NO_COPY_IMPL(_z_value_t, reply_err, _z_value_check, _z_value_null, _z_value_clear)
```
DenisBiryukov91, line 773: Why it became a no copy type while reply err is still in z_clone?

include/zenoh-pico/api/macros.h
```diff
@@ -294,21 +294,14 @@
  *   Returns the cloned instance of `x`.
  */
 #define z_clone(x, y) _Generic((x), \
```
DenisBiryukov91, line 296: we also need to expose a z_slice clone

DenisBiryukov91:
There is a deep copy function _z_keyexpr_copy that can be used for proper z_keyexpr clone (zenoh-c has a keyexpr clone). 
User might also likely need a clone for reply_err aka value (zenoh-c has a reply err clone).
I would also remove z_session_clone since it is not a side-effects free operation (zenoh-c does not have a session clone).

DenisBiryukov91 (approved):

milyin (approved):

PR #650, opened 2024-09-12
Fix arduino examples

### LINKED ISSUE #649: [Bug] Arduino subscriber crashes on esp32
### Describe the bug

A user reported that running a subscriber example on its esp32 using Arduino framework lead to a crash when receiving a pub:

`guru meditation error core panic'ed (LoadProhibited). Exception was unhandled.`

### To reproduce

Load a subscriber on esp32 with arduino framework.
Try to publish some data.

### System info

Esp32 with Arduino framework

jean-roland:
Closes #649 and brushes up the arduino examples a little.

The issue itself was due to owned_entities declared as local variables that went out of scope at the end of `setup()` and were overwritten in the `loop()` phase, causing a segault.

### DIFF
--- a/examples/arduino/z_get.ino
+++ b/examples/arduino/z_get.ino
@@ -21,16 +21,12 @@
 #define SSID "SSID"
 #define PASS "PASS"
 
-#define CLIENT_OR_PEER 0  // 0: Client mode; 1: Peer mode
-#if CLIENT_OR_PEER == 0
+// Client mode values (comment/uncomment as needed)
 #define MODE "client"
 #define CONNECT ""  // If empty, it will scout
-#elif CLIENT_OR_PEER == 1
-#define MODE "peer"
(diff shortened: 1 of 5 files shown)


examples/arduino/z_get.ino
```diff
@@ -21,16 +21,12 @@
 #define SSID "SSID"
 #define PASS "PASS"
 
-#define CLIENT_OR_PEER 0  // 0: Client mode; 1: Peer mode
-#if CLIENT_OR_PEER == 0
```
sashacmc, line 25: What's the reason for not using the CLIENT_OR_PEER approach?

sashacmc (approved):

examples/arduino/z_get.ino
jean-roland: Basically the #if was ignored and compiler complained that #mode and #connect were redefined.

PR #659, opened 2024-09-13
switch Z_FEATURE_UNSTABLE_API to #ifdef

### LINKED ISSUE #633: Align unstable feature support with zenoh-c
### Describe the release item

Currently to check unstable feature support in zenoh-c we use macro UNSTABLE (not very good naming though), while in zenoh-pico we use Z_FEATURE_UNSTABLE = 1. It will simplify zenoh-cpp checks if both implementations would provide the same mechanism for checking availability of unstable features. 

The proposal might to be remove = 1 for pico and just check for features using ifdef/ifndef which seems to be a more common way for binary enabling of features. And then align zenoh-c naming.

jean-roland:
Closes #633.


DenisBiryukov91 (approved):

DenisBiryukov91:
LGTM. Although we should probably also switch to using similar Z_FEATURE_xxx defines in zenoh-c, so I would like to postpone merging till this is resolved.

PR #673, opened 2024-09-16
Add z_bytes example

### LINKED ISSUE #662: Add z_bytes example
### Describe the release item

See https://github.com/eclipse-zenoh/zenoh/issues/1418

jean-roland:
Closes #662.

Adds missing encoding constants.


src/api/api.c
```diff
@@ -197,6 +197,337 @@ int8_t zp_config_insert(z_loaned_config_t *config, uint8_t key, const char *valu
 #if Z_FEATURE_ENCODING_VALUES == 1
 #define ENCODING_SCHEMA_SEPARATOR ';'
 
+const z_owned_encoding_t ENCODING_ZENOH_BYTES = {
```
sashacmc, line 200: It would be nice to define some macro to avoid repeating the definition code so many times.
And also use a separate file to avoid overloading api.c

sashacmc (changes requested):

include/zenoh-pico/api/encoding.h
```diff
@@ -0,0 +1,408 @@
...
+ */
+
+#if Z_FEATURE_ENCODING_VALUES == 1
+// - Below are Primitives types, supported in all Zenoh bindings
+// Just some bytes.
+//
+// Constant alias for string: `"zenoh/bytes"`.
+//
+// Usually used for types: `uint8_t[]`.
+extern const z_owned_encoding_t ENCODING_ZENOH_BYTES;
```
DenisBiryukov91, line 50: please see how encoding constants are implemented in zenoh-c: https://github.com/eclipse-zenoh/zenoh-c/blob/610a4b9057a703ecba5f098e084bdc40e4e786f3/include/zenoh_commons.h#L2254

DenisBiryukov91:
We will also need to add the docs, and probably an example of using slice iterator similar to https://github.com/eclipse-zenoh/zenoh/blob/db755e3e75733b10cd9b9fa63acb624fd4e57a39/examples/examples/z_bytes.rs#L107C1-L112C6

examples/unix/c11/z_bytes.c
```diff
@@ -131,6 +131,21 @@ int main(void) {
...
+    // Slice iterator
+    kv_pair_t input_val[1];
+    input_val[0] = (kv_pair_t){.key = "test_key", .value = "test_value"};
+    ctx = (kv_pairs_tx_t){.data = input_val, .current_idx = 0, .len = 1};
+    z_bytes_from_iter(&payload, hashmap_iter, (void *)&ctx);
+    z_bytes_slice_iterator_t slice_iter = z_bytes_get_slice_iterator(z_bytes_loan(&payload));
+    z_view_slice_t curr_slice;
+    while (z_bytes_slice_iterator_next(&slice_iter, &curr_slice)) {
+        printf("slice len: %d, slice data: '%.*s'\n", (int)z_slice_len(z_view_slice_loan(&curr_slice)),
+               (int)z_slice_len(z_view_slice_loan(&curr_slice)),
```
DenisBiryukov91, line 144: this is not good since it will print garbage, could we rather print slice content as list of hexadecimal values ?

sashacmc (approved):

PR #675, opened 2024-09-17
Replace _Bool with bool to make c++ compiler happy

DenisBiryukov91:
bool is available since C99.
This still does not address the issue of potentially different representation of bool in C vs C++
Also replaces asm argument name which is a reserved keyword in __asm__ windows stub.


DenisBiryukov91:
Superseded by #679

PR #679, opened 2024-09-17
Fix (C++) build for windows

DenisBiryukov91:
Fix C++ build for windows:
Enable shared libs build for windows;
rename __asm__ stub argument from "asm" to "instruction" since asm is reserved keyword;
force exporting all symbols for dll (TODO: mark only public api for export);
replace _Bool with bool, since the later is supported starting from C99 (_Bool is not supported in C++ under windows)


include/zenoh-pico/config.h
```diff
@@ -21,7 +21,7 @@
 #define Z_BATCH_MULTICAST_SIZE 2048
 #define Z_CONFIG_SOCKET_TIMEOUT 100
 
-#define Z_FEATURE_UNSTABLE_API 0
+#define Z_FEATURE_UNSTABLE_API 1
```
jean-roland, line 24: I'd revert this value as PIO doesn't seem to autogenerate this file properly for now. Otherwise, LGTM

jean-roland (approved):

PR #711, opened 2024-10-04
Fix c type formatting in documentation

sashacmc:
`:c:types:` prefix should be used for cross document links, using this prefix for system types make no sense and produce sphinx warnings.
This PR fix it.

### DIFF
--- a/include/zenoh-pico/api/serialization.h
+++ b/include/zenoh-pico/api/serialization.h
@@ -92,7 +92,7 @@ ze_deserializer_t ze_deserializer_from_bytes(const z_loaned_bytes_t *bytes);
 bool ze_deserializer_is_done(const ze_deserializer_t *deserializer);
 
 /**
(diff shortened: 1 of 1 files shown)


include/zenoh-pico/api/serialization.h
```diff
@@ -606,7 +606,7 @@ z_result_t ze_serialize_str(z_owned_bytes_t *bytes, const char *value);
 z_result_t ze_serialize_slice(z_owned_bytes_t *bytes, const z_loaned_slice_t *slice);
 
 /**
- * Serializes a :c:type:`int8_t` into a :c:type:`z_owned_bytes_t` .
+ * Serializes a `int8_t` into a :c:type:`z_owned_bytes_t` .
```
DenisBiryukov91, line 609: an 'int8_t'  (and same for other ints) ?

PR #720, opened 2024-10-08
Subscriber/Queryable api alignment

DenisBiryukov91:
remove z_undeclare_subscriber/publisher/queryable;
make z_subscriber/publisher/queryable_drop automatically undeclare corresponding entities;
add z_declare_background_subscriber/queryable functions, that do not return an entity, but rather register the corresponding callbacks until the session is closed/dropped;
fix memory leaks in publisher due to non-cleared encoding and write_filter;

### DIFF
--- a/include/zenoh-pico/api/primitives.h
+++ b/include/zenoh-pico/api/primitives.h
@@ -1752,7 +1752,7 @@ z_result_t z_declare_queryable(z_owned_queryable_t *queryable, const z_loaned_se
                                const z_queryable_options_t *options);
 
 /**
- * Declares a background queryable for a given keyexpr. The queryable callback will be be called
+ * Declares a background queryable for a given keyexpr. The queryable callback will be called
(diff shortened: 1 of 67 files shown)


include/zenoh-pico/api/primitives.h
```diff
@@ -1765,15 +1752,20 @@ z_result_t z_declare_queryable(z_owned_queryable_t *queryable, const z_loaned_se
                                const z_queryable_options_t *options);
 
 /**
- * Undeclares and clears the queryable.
+ * Declares a background queryable for a given keyexpr. The queryable callback will be be called
```
sashacmc, line 1755:
```suggestion
 * Declares a background queryable for a given keyexpr. The queryable callback will be called
```

jean-roland (approved):

sashacmc (approved):

milyin (approved):

PR #730, opened 2024-10-10
Improve throughput performance (trigger local subscriber focus)

jean-roland:
There is a small performance dip in non-fragment peer throughput compared with 0.11. Profiling helped identify a much longer time spent in `trigger_local_subscriber`and this PR aims to fix this.

* Static inline functions:
Functions that should have been inlined by the compiler (like `z_sample_null`) weren't. This is solved by explicitly setting all the `_null`, `_check` and `_alias` functions to `static inline`.

* Null functions:
A lot of time was spent initializing types to 0 compared to just using `memset(0)`, switching all the `_null` functions to a `{0}` initialization closed this gap and is less prone to error.

* Slice init / string preallocate:
This functions had many superfluous checks / init. They got simplified.

* get_expanded_key_by_key:
Added a shorter path for already expanded keys.

* trigger_local_subcriber:
Added a shorter path when there is 0 subscribers to trigger.
Added a config token, `off` by default, to execute this function.

*The fastest code is the one we don't execute*

### DIFF
--- a/src/net/reply.c
+++ b/src/net/reply.c
@@ -40,13 +40,13 @@ void _z_reply_data_free(_z_reply_data_t **reply_data) {
 
 z_result_t _z_reply_data_copy(_z_reply_data_t *dst, const _z_reply_data_t *src) {
     *dst = _z_reply_data_null();
+    dst->_tag = src->_tag;
     if (src->_tag == _Z_REPLY_TAG_DATA) {
         _Z_RETURN_IF_ERR(_z_sample_copy(&dst->_result.sample, &src->_result.sample));
     } else if (src->_tag == _Z_REPLY_TAG_ERROR) {
         _Z_RETURN_IF_ERR(_z_value_copy(&dst->_result.error, &src->_result.error));
     }
     dst->replier_id = src->replier_id;
-    dst->_tag = src->_tag;
     return _Z_RES_OK;
 }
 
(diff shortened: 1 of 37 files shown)


src/session/resource.c
```diff
@@ -136,6 +139,7 @@ _z_keyexpr_t __z_get_expanded_key_from_key(_z_resource_list_t *xs, const _z_keye
             }
         }
     }
+    // Warning: list must be release with _z_list_free(&strs, _z_noop_free) or will release the suffix as well
```
DenisBiryukov91, line 142: list must be releaseD

DenisBiryukov91:
Looks good, but I do not really like the idea of memsetting everything to 0, especially for dependent structs. In case we modify null-constructors for their fields, we risk to face some difficult to detect bugs.
I'm wondering if memsetting to 0 really provides a speed boost that is worth the cost.

jean-roland:
I understand the issue, but then maybe we should make a distinction between `_null` which would be some kind of *memory* initialization and `_init` that would be a *type* initialization, in the case where those two operations are different.

If we distinguish the two then, I believe the performance increase won't cause problems.

src/net/reply.c
```diff
@@ -49,13 +40,13 @@ void _z_reply_data_free(_z_reply_data_t **reply_data) {
 
 z_result_t _z_reply_data_copy(_z_reply_data_t *dst, const _z_reply_data_t *src) {
     *dst = _z_reply_data_null();
+    dst->_tag = src->_tag;
```
sashacmc, line 43: What the reason of this change?
In case of error will be nice to avoid unnecessary dst changes.

jean-roland: Yeah so "no data" value of tag is `3` and not `0` so I was trying to avoid a `reply_data_init` function, but I guess that won't be avoided.

sashacmc (approved):

PR #735, opened 2024-10-11
Fix undefined behavior on udp config

jean-roland:
`_z_string_t` treated as null-terminated strings in config were causing undefined behavior picked up by valgrind

### DIFF
--- a/src/protocol/config.c
+++ b/src/protocol/config.c
@@ -16,6 +16,7 @@
 
 #include <stdbool.h>
 #include <stddef.h>
+#include <stdio.h>
 #include <string.h>
 
 #include "zenoh-pico/utils/pointers.h"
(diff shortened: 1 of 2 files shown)


src/protocol/config.c
```diff
@@ -75,20 +77,25 @@ z_result_t _z_str_intmap_from_strn(_z_str_intmap_t *strint, const char *s, uint8
...
+                p_value_len = value_max_size;
+            } else {
+                p_value_len = _z_ptr_char_diff(p_value_end, p_value_start);
             }
-
-            size_t p_value_len = _z_ptr_char_diff(p_value_end, p_value_start) + (size_t)1;
-            char *p_value = (char *)z_malloc(p_value_len);
+            char *p_value = (char *)z_malloc(p_value_len + 1);
             if (p_value != NULL) {
-                _z_str_n_copy(p_value, p_value_start, p_value_len);
```
sashacmc, line 86: If we still need `\0` at the end it will be better to use strncmp (or _z_str_n_copy like before with adjusted length)

sashacmc (approved):

PR #737, opened 2024-10-12
Improve throughput performance #2 (refcount focus)

### LINKED ISSUE #520: Add back a refcount implementation without weak feature
### Describe the feature

With #510, rc now has weak capabilities. This is currently only needed for sessions, it might be worth to add a non-weak rc for all other types.

Also, it seems queries could hold directly a strong reference to the session.

jean-roland:
This PR continues on the work of #730. Profiling helped identify multiple causes of performance issue, including our refcount implementation with weak pointers.

* More static inline rework:
Similar work to what was done previously.

* Lazier & faster clear / initialization:
Some types were taking time initializing or clearing things that weren't necessary. They got simplified.
Chief of this was timestamps. A validity boolean flag was added to the struct to simplify checks & clears of the type.

* "Simple rc" - name subject to change:
Added back a rc implementation without weak, used only for arc_slices. Closes #520 

* Add `Z_FEATURE_PUBLISHER_SESSION_CHECK` config token
On by default, this can be turned off to avoid upgrading the publisher's session weak in apps where session is not dropped before it stops publishing.

* Reworked logs from the transport/codec layer.

* Multicast read task more tolerant:
As pico multicast is on a `best-effort` transport only, a missed fragment shouldn't close the connection (it was causing tests to halt).


src/protocol/codec/message.c
```diff
@@ -102,6 +102,7 @@ z_result_t _z_timestamp_decode(_z_timestamp_t *ts, _z_zbuf_t *zbf) {
     _Z_DEBUG("Decoding _TIMESTAMP");
     z_result_t ret = _Z_RES_OK;
 
+    ts->valid = true;
```
DenisBiryukov91, line 105: What if decoding below fails, should timestamp be still marked as valid. I think it is better to set at the end to (res == _Z_RES_OK)

DenisBiryukov91 (approved):

src/link/link.c
```diff
@@ -153,7 +153,7 @@ size_t _z_link_recv_exact_zbuf(const _z_link_t *link, _z_zbuf_t *zbf, size_t len
 
 z_result_t _z_link_send_wbuf(const _z_link_t *link, const _z_wbuf_t *wbf) {
     z_result_t ret = _Z_RES_OK;
-    bool link_is_streamed = false;
+    bool link_is_streamed;
```
sashacmc, line 156: nit: it will be better to have 
```
bool link_is_streamed = link->_cap._flow == Z_LINK_CAP_FLOW_STREAM;
```

src/protocol/codec/declarations.c
```diff
@@ -234,6 +235,7 @@ z_result_t _z_decl_commons_decode(_z_zbuf_t *zbf, uint8_t header, bool *has_exte
         _z_zint_t len;
         _Z_RETURN_IF_ERR(_z_zsize_decode(&len, zbf));
         if (_z_zbuf_len(zbf) < len) {
+            _Z_INFO("Not enough bytes to read");
```
sashacmc, line 238: Seems we need to add _Z_LOG_LVL_WARN (not in this PR, but will be nice to have it for cases like it)

src/session/utils.c
```diff
@@ -28,18 +28,24 @@
...
+    ts.valid = tstamp->valid;
     ts.id = tstamp->id;
     ts.time = tstamp->time;
     return ts;
 }
 
+void _z_timestamp_move(_z_timestamp_t *dst, _z_timestamp_t *src) {
+    dst->valid = src->valid;
+    dst->id = src->id;
+    dst->time = src->time;
```
sashacmc, line 40:
```suggestion
    *dst = *src;
```

src/session/utils.c
```diff
@@ -28,18 +28,24 @@
 /*------------------ clone helpers ------------------*/
 _z_timestamp_t _z_timestamp_duplicate(const _z_timestamp_t *tstamp) {
     _z_timestamp_t ts;
+    ts.valid = tstamp->valid;
     ts.id = tstamp->id;
     ts.time = tstamp->time;
     return ts;
```
sashacmc, line 34:
```suggestion
    return *ts;
```

src/protocol/codec/declarations.c
jean-roland: Yes that's what I was thinking about. Will open an issue

sashacmc (approved):

PR #749, opened 2024-10-17
Improve throughput performance #3 (rx buffer focus) 

jean-roland:
In order to improve rx performance we need to reduce the amount of copies that are made. By using a refcount and by allocating dynamically the rx buffers we can avoid copies at the cost of memory allocation, only in the case where data is processed out of context [note: that is the case of query replies in latest consolidation mode].

* The rx buffers (zbuf_t) now have a refcounted slice. If after processing there are more than 1 refcount, the transport will allocate a new buffer.
* Fixed a few undefined behaviors in tests and examples.

### DIFF
--- a/include/zenoh-pico/collections/refcount.h
+++ b/include/zenoh-pico/collections/refcount.h
@@ -244,8 +244,8 @@ size_t _z_simple_rc_strong_count(void *cnt);
         *p = name##_simple_rc_null();                                                                         \
         return res;                                                                                           \
     }                                                                                                         \
-    static inline bool name##_simple_rc_is_last_ref(const name##_simple_rc_t *p) {                            \
-        return (_z_simple_rc_strong_count(p->_cnt) == 1);                                                     \
+    static inline size_t name##_simple_rc_count(const name##_simple_rc_t *p) {                                \
+        return _z_simple_rc_strong_count(p->_cnt);                                                            \
     }                                                                                                         \
     static inline size_t name##_simple_rc_size(name##_simple_rc_t *p) {                                       \
         _ZP_UNUSED(p);                                                                                        \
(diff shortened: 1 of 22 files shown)


src/transport/raweth/rx.c
```diff
@@ -117,6 +117,26 @@ z_result_t _z_raweth_recv_t_msg(_z_transport_multicast_t *ztm, _z_transport_mess
     return _z_raweth_recv_t_msg_na(ztm, t_msg, addr);
 }
 
+z_result_t _z_raweth_update_rx_buff(_z_transport_multicast_t *ztm) {
+    // Check if user or defragment buffer took ownership of buffer
+    if (!_z_zbuf_is_last_ref(&ztm->_zbuf)) {
+        // Allocate a new buffer
+        _z_zbuf_t new_zbuf = _z_zbuf_make(Z_BATCH_MULTICAST_SIZE);
```
sashacmc, line 124: I don't know the raweth very well, but should we use Z_BATCH_MULTICAST_SIZE there?

jean-roland: Yes because raweth is a mix between a link and a multicast transport so it should behave as closely as possible as such.

sashacmc (approved):

include/zenoh-pico/collections/refcount.h
```diff
@@ -244,6 +244,9 @@ size_t _z_simple_rc_strong_count(void *cnt);
         *p = name##_simple_rc_null();                                                                         \
         return res;                                                                                           \
     }                                                                                                         \
+    static inline bool name##_simple_rc_is_last_ref(const name##_simple_rc_t *p) {                            \
```
DenisBiryukov91, line 247: I think it would be more useful to have a function that returns the reference count.

PR #750, opened 2024-10-18
Fix badly named encoding constants

jean-roland:
Seems `encoding.c` and `encoding.h` did not agree on the encoding constants name which made them unusable by user.


sashacmc (commented):
Are we proposing to use constants ZP_ENCODING_XXX rather than methods z_encoding_xxx like in zenoh-c?
If not will be better to name _Z_ENCODING_XXX

sashacmc (approved):

PR #752, opened 2024-10-19
fix: update debian packaging

diogomatsubara:
- include libzenohpico.so in debian package
- change the package name to libzenohpico in line with libzenohc
- set correct debian version for pre releases
- fix wrong version used in Release mode


CMakeLists.txt
```diff
@@ -541,9 +549,11 @@ if(PACKAGING)
     set(CPACK_DEBIAN_PACKAGE_ARCHITECTURE ${DEBARCH})
     set(CPACK_DEB_COMPONENT_INSTALL ON)
     set(CPACK_DEBIAN_FILE_NAME DEB-DEFAULT)
-    set(CPACK_DEBIAN_LIB_PACKAGE_NAME ${PROJECT_NAME}) # avoid "-lib" suffix for "lib" package
+    set(CPACK_DEBIAN_PACKAGE_VERSION ${debian_version})
+    set(CPACK_DEBIAN_LIB_PACKAGE_NAME "lib${PROJECT_NAME}")
     set(CPACK_DEBIAN_LIB_PACKAGE_DEPENDS "libc6 (>=2.12)")
-    set(CPACK_DEBIAN_DEV_PACKAGE_DEPENDS "${CPACK_DEBIAN_LIB_PACKAGE_NAME} (=${PROJECT_VERSION})")
+    set(CPACK_DEBIAN_DEV_PACKAGE_NAME "lib${PROJECT_NAME}_dev")
```
sashacmc, line 555:
```suggestion
    set(CPACK_DEBIAN_DEV_PACKAGE_NAME "lib${PROJECT_NAME}-dev")
```
Debian dev package should have "-dev" suffix

CMakeLists.txt
```diff
@@ -43,8 +43,10 @@ set(project_version "${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR}.${PROJECT
 if(NOT DEFINED PROJECT_VERSION_TWEAK)
 	set(project_version "${project_version}")
 elseif(PROJECT_VERSION_TWEAK EQUAL 0)
+	set(debian_version "${project_version}~dev-1")
 	set(project_version "${project_version}-dev")
 elseif(PROJECT_VERSION_TWEAK GREATER 1)
+	set(debian_version "${project_version}~pre.${PROJECT_VERSION_TWEAK}-1")
```
sashacmc, line 49: It is not entirely clear for me what this `-1` means and why it is hardcoded.

sashacmc (changes requested):

CMakeLists.txt
diogomatsubara: To comply with the debian version policy.

CMakeLists.txt
diogomatsubara: Fixed in fba07a9e

CMakeLists.txt
sashacmc: Can it be 2, 3 etc in future?

diogomatsubara: Unlikely, so far we haven't had any reason to change it. In case we need to change in a further date I can make it default to `-1` and allow the user to set a value.

diogomatsubara:
I rebased on top of Denis's PR #751 which handles the `libzenohpico.so` better. Once that one is merged, we can merge this one.

sashacmc (approved):

Mallets (approved):

PR #755, opened 2024-10-21
Remove default parameters from z_closure

sashacmc:
The default parameters for `z_closure` not implemented in zenoh-c. It's possible to add it (done in PR https://github.com/eclipse-zenoh/zenoh-c/pull/782) , but in fact this functionality is useless in practice: no one will use callbacks without context in real applications. So for simplicity it seems better to remove this, despite that this is slight API change.


DenisBiryukov91 (approved):

kydos (approved):
I am fine with the change. We can merge.

PR #758, opened 2024-10-22
Add manual batching mechanism

### LINKED ISSUE #298: Add a way to manually batch messages
### Describe the feature

Zenoh pico does not support the emission of batch messages even if it is pretty useful when sending large quantity of small messages.

We should add a mechanism to batch message from the API, as evoked in PR #203

jean-roland:
Closes #298. This adds an API to do manual batching which yields improved throughput performance for smaller packets.

User can start batching with `zp_start_batching`, any subsequent api call that would have put a network message (`z_put`, `z_get`, `z_declare`...) is bufferized to be later sent as a batch (or multiple batches) when user calls `zp_stop_batching`.

While I was modifying the transports I decided to refactor a little because the functions were too complicated to parse.

### DIFF
--- a/include/zenoh-pico/api/primitives.h
+++ b/include/zenoh-pico/api/primitives.h
@@ -2057,8 +2057,29 @@ const z_loaned_keyexpr_t *z_subscriber_keyexpr(const z_loaned_subscriber_t *subs
 #endif
 
 #if Z_FEATURE_BATCHING == 1
-z_result_t zp_batching_start(const z_loaned_session_t *zs);
-z_result_t zp_batching_stop(const z_loaned_session_t *zs);
+/**
+ * Activate the batching mechanism.
+ * Any message that would have been sent on the network by a subsequent api call (e.g z_put, z_get)
+ * will be instead stored until the batch is flushed with :c:func:`zp_batch_flush`.
+ *
+ * Parameters:
+ *   zs: Pointer to a :c:type:`z_loaned_session_t` that will start batching messages.
+ *
+ * Return:
+ *   ``0`` if batching started, ``negative value`` otherwise.
+ */
+z_result_t zp_batch_start(const z_loaned_session_t *zs);
+
+/**
+ * Deactivate the batching mechanism and flush the messages that were stored.
+ *
+ * Parameters:
+ *   zs: Pointer to a :c:type:`z_loaned_session_t` that will stop batching messages.
+ *
+ * Return:
+ *   ``0`` if batching stopped and batch successfully sent, ``negative value`` otherwise.
+ */
+z_result_t zp_batch_flush(const z_loaned_session_t *zs);
 #endif
 
 /************* Multi Thread Tasks helpers **************/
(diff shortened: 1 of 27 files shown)


include/zenoh-pico/api/primitives.h
```diff
@@ -2056,6 +2056,11 @@ z_result_t z_declare_background_subscriber(const z_loaned_session_t *zs, const z
 const z_loaned_keyexpr_t *z_subscriber_keyexpr(const z_loaned_subscriber_t *subscriber);
 #endif
 
+#if Z_FEATURE_BATCHING == 1
+z_result_t zp_batching_start(const z_loaned_session_t *zs);
+z_result_t zp_batching_stop(const z_loaned_session_t *zs);
```
sashacmc, line 2061: I don't think that `stop` good word for it, we should choose something which will show that this method is sending data.

src/transport/multicast/transport.c
```diff
@@ -206,6 +219,44 @@ void _z_multicast_transport_clear(_z_transport_t *zt) {
...
+
+void _z_multicast_peer_mutex_unlock(_z_transport_multicast_t *ztm) { _z_mutex_unlock(&ztm->_mutex_peer); }
+
+#else
+z_result_t _z_multicast_tx_mutex_lock(_z_transport_multicast_t *ztm, bool block) {
+    _ZP_UNUSED(ztm);
+    _ZP_UNUSED(block);
+    return _Z_RES_OK;
+}
+void _z_multicast_tx_mutex_unlock(_z_transport_multicast_t *ztm) { _ZP_UNUSED(ztm); }
```
sashacmc, line 248: Will compiler optimize empty method call, maybe will be better to use macros? They will be definitely optimized.
And even in multithreaded mode we can avoid unnecessary call. Other option will be use `inline` methods.

include/zenoh-pico/api/primitives.h
sashacmc: Maybe something like 
```
zp_batch_start
zp_batch_flush
```

src/transport/multicast/tx.c
```diff
@@ -40,126 +41,224 @@ _z_zint_t __unsafe_z_multicast_get_sn(_z_transport_multicast_t *ztm, z_reliabili
...
+                                                     const _z_network_message_t *n_msg, z_reliability_t reliability,
+                                                     _z_zint_t first_sn) {
+    bool is_first = true;
+    _z_zint_t sn = first_sn;
+    // Encode message on temp buffer
+    _Z_RETURN_IF_ERR(_z_network_message_encode(fbf, n_msg));
+    // Fragment message
+    while (_z_wbuf_len(fbf) > 0) {
+        // Get fragment sequence number
+        if (is_first == false) {
```
sashacmc, line 55:
```suggestion
        if (!is_first) {
```

sashacmc (commented):
Last nit, others LGTM

PR #759, opened 2024-10-22
Fix z_task cleanup for platforms with pthread support

sashacmc:
pthread don't require free on pthread_t


jean-roland (approved):

milyin (approved):

milyin:
fix caused memory leak, followup fix in https://github.com/eclipse-zenoh/zenoh-pico/pull/760

PR #777, opened 2024-11-06
Improve sub frame decode performance

jean-roland:
The introduction of batching revealed lot of room to improve performance in frame decoding. The will be split in a subscriber focused PR (this one) and a query focused one (TBD).

* Introduce Memoization (off by default):

One of the costliest operation in subscriber triggering was the construction of the key-expression and the callback lookup. Memoizing the results improved significantly performance. Although the current implementation is a bit trivial as it only holds one entry. Probably worth adding a true lru cache implementation in next PR (open to suggestion).

* Frame decoding buffer switch from vec to svec;

With frames containing 4000+ frames, the use of reference vectors were causing a lot of memory fragmentation and inefficiencies. The use of svec is much better but require to minimize reallocation, thus a configurable `Z_CONFIG_FRAME_AVG_MSG_SIZE` has been introduced to strike a balance between memory consumption and time spent on reallocation.

Svec/elem were a bit improved as well (added the elem_move, ability to choose default svec move/copy behaviour with boolean, return result instead of bool...)

* Removed redundant initialization and drop:

There were a few cases of superfluous or redundant type init and clear. Specifically on the network messages and they were removed.


include/zenoh-pico/config.h
```diff
@@ -45,6 +45,7 @@
 #define Z_FEATURE_LOCAL_SUBSCRIBER 0
 #define Z_FEATURE_PUBLISHER_SESSION_CHECK 1
 #define Z_FEATURE_BATCHING 1
+#define Z_FEATURE_MEMOIZATION 0
```
sashacmc, line 48:
```suggestion
#define Z_FEATURE_MEMORIZATION 0
```

src/protocol/codec/message.c
```diff
@@ -332,7 +332,6 @@ z_result_t _z_push_body_decode(_z_push_body_t *pshb, _z_zbuf_t *zbf, uint8_t hea
     switch (_Z_MID(header)) {
         case _Z_MID_Z_PUT: {
             pshb->_is_put = true;
-            pshb->_body._put = (_z_msg_put_t){0};
```
sashacmc, line 335: What the reason to remove this initialization?

sashacmc: Can we guarantee that decode will overwrite all fields?

src/protocol/codec/transport.c
```diff
@@ -28,6 +28,10 @@
 #include "zenoh-pico/utils/logging.h"
 #include "zenoh-pico/utils/result.h"
 
+#define _Z_FRAME_VEC_BASE_SIZE 8  // Abritrary small value
+#define _Z_FRAME_VEC_SIZE_FROM_ZBUF_LEN(len) \
+    _Z_FRAME_VEC_BASE_SIZE + (len) / Z_CONFIG_FRAME_AVG_MSG_SIZE  // Approximate number of messages in frame
```
sashacmc, line 33:
```suggestion
    (_Z_FRAME_VEC_BASE_SIZE + (len) / Z_CONFIG_FRAME_AVG_MSG_SIZE)  // Approximate number of messages in frame
```
To avoid potential incorrect usage in expressions

include/zenoh-pico/config.h
jean-roland: The technical term is Memoization though, https://en.wikipedia.org/wiki/Memoization. But yeah that might not be transparent to the user, in that case we might as well go to something like `Z_FEATURE_RX_CACHE`

src/protocol/codec/message.c
jean-roland: Decode will not overwrite all fields (for instance it's not going to write a timestamp if the messaged doesn't have one) but the point of the change is that put message is part of a bigger generic container,`z_network_message_t` which itself is already initialized.

sashacmc (approved):

PR #781, opened 2024-11-11
Improve query/reply perf

jean-roland:
The last perf PR before 1.1 merge into main. Include:

* Use of resource pool for decoding:

For every message of a frame we used to allocate and free a network message and a `Vec<Arc<Subslice>>` of size 1 for every payload. Instead of doing that, a transport now has a `Vec<network message>` and a `Vec<Arc<Subslice>>` that can be reused from a frame to another to avoid the time lost in allocating/freeing and the fragmentation.

* Svec use_elem_f

Svec allows to chose if we want to use the element function when we copy/move per call instead of at define.

* Query/reply alignment with sample changes:

Query, reply and samples functions take ownership of the decoded data and free them systematically.
Queryable have a cache system similar to subscriber (that will also be moved to a lru cache in the future).
Queries are not longer an rc and holds a reference to the session instead of a weak.

### DIFF
--- a/examples/unix/c11/z_get_attachment.c
+++ b/examples/unix/c11/z_get_attachment.c
@@ -70,7 +70,7 @@ void reply_handler(z_loaned_reply_t *reply, void *ctx) {
         const z_loaned_bytes_t *attachment = z_sample_attachment(sample);
         ze_deserializer_t deserializer = ze_deserializer_from_bytes(attachment);
         size_t attachment_len;
-        if (ze_deserializer_deserialize_sequence_length(&deserializer, &attachment_len) != 0) {
+        if (ze_deserializer_deserialize_sequence_length(&deserializer, &attachment_len) < 0) {
             return;
         }
         kv_pair_t *kvp = (kv_pair_t *)malloc(sizeof(kv_pair_t) * attachment_len);
--- a/examples/unix/c11/z_get_attachment.c
+++ b/examples/unix/c11/z_get_attachment.c
@@ -68,12 +68,11 @@ void reply_handler(z_loaned_reply_t *reply, void *ctx) {
 
         // Check attachment
         const z_loaned_bytes_t *attachment = z_sample_attachment(sample);
-        if (attachment == NULL) {
-            return;
-        }
         ze_deserializer_t deserializer = ze_deserializer_from_bytes(attachment);
         size_t attachment_len;
-        ze_deserializer_deserialize_sequence_length(&deserializer, &attachment_len);
+        if (ze_deserializer_deserialize_sequence_length(&deserializer, &attachment_len) != 0) {
+            return;
+        }
         kv_pair_t *kvp = (kv_pair_t *)malloc(sizeof(kv_pair_t) * attachment_len);
         for (size_t i = 0; i < attachment_len; ++i) {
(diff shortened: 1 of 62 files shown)


examples/unix/c11/z_get_attachment.c
```diff
@@ -68,12 +68,11 @@ void reply_handler(z_loaned_reply_t *reply, void *ctx) {
 
         // Check attachment
         const z_loaned_bytes_t *attachment = z_sample_attachment(sample);
-        if (attachment == NULL) {
-            return;
-        }
         ze_deserializer_t deserializer = ze_deserializer_from_bytes(attachment);
         size_t attachment_len;
-        ze_deserializer_deserialize_sequence_length(&deserializer, &attachment_len);
+        if (ze_deserializer_deserialize_sequence_length(&deserializer, &attachment_len) != 0) {
```
sashacmc, line 73:
```suggestion
        if (ze_deserializer_deserialize_sequence_length(&deserializer, &attachment_len) < 0) {
```
According to our API, an error is a value less than zero and in others examples we check "< 0"

examples/unix/c11/z_queryable_attachment.c
```diff
@@ -68,10 +68,9 @@ void query_handler(z_loaned_query_t *query, void *ctx) {
 
     // Check attachment
     const z_loaned_bytes_t *attachment = z_query_attachment(query);
-    if (attachment != NULL) {
-        ze_deserializer_t deserializer = ze_deserializer_from_bytes(attachment);
-        size_t attachment_len;
-        ze_deserializer_deserialize_sequence_length(&deserializer, &attachment_len);
+    ze_deserializer_t deserializer = ze_deserializer_from_bytes(attachment);
+    size_t attachment_len;
+    if (ze_deserializer_deserialize_sequence_length(&deserializer, &attachment_len) == 0) {
```
sashacmc, line 73:
```suggestion
    if (ze_deserializer_deserialize_sequence_length(&deserializer, &attachment_len) == Z_OK) {
```
Good to don't have hard coded numbers where it possible

src/protocol/keyexpr.c
```diff
@@ -103,7 +102,10 @@ bool _z_keyexpr_equals(const _z_keyexpr_t *left, const _z_keyexpr_t *right) {
     if (_z_keyexpr_mapping_id(left) != _z_keyexpr_mapping_id(right)) {
         return false;
     }
-    return _z_string_equals(&left->_suffix, &right->_suffix);
+    if (_z_keyexpr_has_suffix(left) || _z_keyexpr_has_suffix(right)) {
```
sashacmc, line 105: Will better to add a case for _z_keyexpr_has_suffix(left) && _z_keyexpr_has_suffix(right), because as far as I see, _z_string_equals will have UB in case of keyexpr with NULL suffix

jean-roland: Then to take all possible cases it should be something like:
```C
if (_z_keyexpr_has_suffix(left) != _z_keyexpr_has_suffix(right)) {
    return false;
}
if (_z_keyexpr_has_suffix(left) && _z_keyexpr_has_suffix(right)) {
    return _z_string_equals(&left->_suffix, &right->_suffix);
}
```

PR #812, opened 2024-11-22
Add AddressSanitizer to CI tests

sashacmc:
Add AddressSanitizer to CI tests and fix memory issues:
 - weak reference counting
 - keyexpr memory access
 - unitest memory leaks


jean-roland (approved):

CMakeLists.txt
```diff
@@ -453,6 +454,11 @@ if(BUILD_EXAMPLES)
   add_subdirectory(examples)
 endif()
 
+if(ASAN)
+  add_compile_options(-fsanitize=address)
```
DenisBiryukov91, line 458: this is not correct, since -fsanitize is a clang feature only and thus will fail with other compilers

sashacmc: I didn't tested it with MSVC, but as far as I see there
https://learn.microsoft.com/en-us/cpp/build/reference/compiler-options?view=msvc-170
and there
https://learn.microsoft.com/en-us/cpp/sanitizers/asan?view=msvc-170
this type of option is also correct.
In addition, this is an optional option that does not have to be enabled for all compilers.

DenisBiryukov91 (approved):

PR #822, opened 2024-12-08
FreeRTOS-Plus-TCP port fixes

bjsowa:
1. Fixed mutex drop by using vSemaphoreDelete instead of z_free.
2. Added static allocation support to mutexes.
3. Fixed _z_task_detach by pointing it to _z_task_cancel.
4. Fixed freeing task resources. The current implementation does not make much sense.
5. Fixed issues with task cancelling and joining.
6. Used gettimeofday for z_time functions. It doesn't really make sense to use tick count for this. If compiled with newlib for a target that does not support this function, user can just implement `_gettimeofday` if needed.

### DIFF
--- a/include/zenoh-pico/system/platform/freertos_plus_tcp.h
+++ b/include/zenoh-pico/system/platform/freertos_plus_tcp.h
@@ -37,6 +37,11 @@ typedef struct {
 typedef struct {
     TaskHandle_t handle;
     EventGroupHandle_t join_event;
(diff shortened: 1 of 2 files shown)


bjsowa:
I'm still testing some things, will mark this as ready for review once I'm done.

jean-roland (approved):

sashacmc (approved):

PR #841, opened 2024-12-19
Add lru cache

jean-roland:
1.1 introduced a placeholder rx cache feature that will be filled with this PR. Cache is still off by default because it's aimed at high throughput scenarios.

Cache itself is a double linked list (ideally suited for lru operations)  ~~and has an optional binary tree hybridation to improve search time on bigger caches. Although given the memory requirements and embedded constraints it's off by default.~~ associated to a sorted array for lookup.

Changed some functions to accept keyexpr pointers instead of value.


src/session/queryable.c
```diff
@@ -32,56 +32,62 @@
...
-    return;
+#if Z_FEATURE_RX_CACHE == 1
+int _z_queryable_cache_data_compare(const void *first, const void *second) {
+    _z_queryable_cache_data_t *first_data = (_z_queryable_cache_data_t *)first;
+    _z_queryable_cache_data_t *second_data = (_z_queryable_cache_data_t *)second;
+    return _z_keyexpr_compare(&first_data->ke_in, &second_data->ke_in);
 }
 #endif  // Z_FEATURE_RX_CACHE == 1
 
+// #if Z_FEATURE_RX_CACHE == 1
```
sashacmc, line 48: Do we need this code commented?

src/collections/lru_cache.c
```diff
@@ -0,0 +1,325 @@
...
+    }
+    // Compare keys
+    int res = compare(_z_lru_cache_node_value(node), value);
+    // Node has key
+    if (res == 0) {
+        return node;
+    }
+    _z_lru_cache_node_data_t *node_data = _z_lru_cache_node_data(node);
+    // Key is greater than node's key
+    if (res > 0) {
```
sashacmc, line 148: How big is the cache expected to be in real life?
With the recursive approach there is a danger of stack overflow.

src/collections/string.c
```diff
@@ -101,6 +101,13 @@ void _z_string_free(_z_string_t **str) {
     }
 }
 
+int _z_string_compare(const _z_string_t *left, const _z_string_t *right) {
+    if (_z_string_len(left) <= _z_string_len(right)) {
+        return strncmp(_z_string_data(left), _z_string_data(right), _z_string_len(left));
+    }
+    return strncmp(_z_string_data(left), _z_string_data(right), _z_string_len(right));
+}
+
```
sashacmc, line 110: Use case for equal prefix with different length was missed.
For instance, "abc" and "abcd" would compare as equal.
This will be better:

```suggestion
int _z_string_compare(const _z_string_t *left, const _z_string_t *right) {
    size_t len_left = _z_string_len(left);
    size_t len_right = _z_string_len(right);

    int result = strncmp(_z_string_data(left), _z_string_data(right), len_left < len_right ? len_left : len_right);

    if (result == 0) {
        if (len_left < len_right) {
            return -1;
        } else if (len_left > len_right) {
            return 1;
        }
    }

    return result;
}

```

src/collections/lru_cache.c
jean-roland: Yes this is why the tree can be deactivated. Since it's one entry / keyexpr used on the network I'd say 10-100 entries. 

The tree is more aiming at potential PC/MPU using pico with high throughput and lots of keyexpr.

src/session/queryable.c
jean-roland: Nope, that's an oversight.

src/collections/lru_cache.c
jean-roland: After thinking about it, we probably should replace the tree with a separate sorted svec with an insert sort and a binary search, that'd be slightly more costly in allocated memory but it won't have the recursive issue. 
Also the nature of the data might not be random (declared keyexpr are given id in a monotonic fashion) which means we're might easily fall in an unbalanced tree scenario.

sashacmc (approved):

PR #847, opened 2024-12-31
CMakeLists.txt: don't override the optimization level

tpetazzoni:
zenoh-pico's CMakeLists.txt has some logic to provide its own compiler optimization level: -O3 for release builds, -O0 for debug builds.

Unfortunately, using add_compile_options() means that those settings take precedence over what the user can pass as custom compiler flags. And this causes issues for example when doing a debug build with _FORTIFY_SOURCE enabled, as _FORTIFY_SOURCE support in glibc is incompatible with unoptimized builds causing this build failure:

/home/thomas/projets/buildroot/output/host/arm-buildroot-linux-gnueabihf/sysroot/usr/include/features.h:414:4: error: #warning _FORTIFY_SOURCE requires compiling with optimization (-O) [-Werror=cpp]
  414 | #  warning _FORTIFY_SOURCE requires compiling with optimization (-O)
      |    ^~~~~~~

which is a warning, but as zenoh-pico builds with -Werror, it turns into a build failure.

As it is unclear how CMakeLists.txt can pass a default -O level, while allowing it to be overridden by the user, we simply remove those optimization level options.


tpetazzoni:
The CI asks me to add a label... but I don't have the rights to do so. The "bug" label should be added.

PR #850, opened 2025-01-06
Update batching to stable.

jean-roland:
This PR fixes the few issues that prevented manual batching to be stable:
* is_express qos should trigger an immediate batch send.
* keep alive, close and other transport messages should be sent immediately.

### DIFF
--- a/src/protocol/codec/network.c
+++ b/src/protocol/codec/network.c
@@ -38,14 +38,11 @@
 
 /*------------------ Push Message ------------------*/
 
-z_result_t _z_push_encode(_z_wbuf_t *wbf, const _z_n_msg_push_t *msg, bool *is_express) {
+z_result_t _z_push_encode(_z_wbuf_t *wbf, const _z_n_msg_push_t *msg) {
     uint8_t header = _Z_MID_N_PUSH | (_z_keyexpr_is_local(&msg->_key) ? _Z_FLAG_N_REQUEST_M : 0);
     bool has_suffix = _z_keyexpr_has_suffix(&msg->_key);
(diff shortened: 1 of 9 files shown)


src/protocol/codec/network.c
```diff
@@ -38,11 +38,14 @@
 
 /*------------------ Push Message ------------------*/
 
-z_result_t _z_push_encode(_z_wbuf_t *wbf, const _z_n_msg_push_t *msg) {
+z_result_t _z_push_encode(_z_wbuf_t *wbf, const _z_n_msg_push_t *msg, bool *is_express) {
     uint8_t header = _Z_MID_N_PUSH | (_z_keyexpr_is_local(&msg->_key) ? _Z_FLAG_N_REQUEST_M : 0);
     bool has_suffix = _z_keyexpr_has_suffix(&msg->_key);
     bool has_qos_ext = msg->_qos._val != _Z_N_QOS_DEFAULT._val;
     bool has_timestamp_ext = _z_timestamp_check(&msg->_timestamp);
+    if (is_express != NULL) {
```
sashacmc, line 46: To be honest, I don't really like the approach of calculating is_express value as a side effect of the encoding method.
It would be better to make a separate method for this calculation. Yes, it's a bit more code and maybe some performance loss because of the extra call (but maybe it will be optimized and there will be no loss at all)
But I think this is the price to pay for clean code.

sashacmc (changes requested):

sashacmc (approved):
Small syntax sugar, others LGTM

gabrik (approved):

PR #880, opened 2025-02-03
Fix queryables memory leak

### LINKED ISSUE #879: [Bug] _z_trigger_queryables_inner() does not release infos for unhandled queries
### Describe the bug

When `Z_FEATURE_RX_CACHE = 0` and an unhandled query is received, the function `_z_trigger_queryables_inner()` in session/queryable.c drops the keyexpr, but leaks the infos vector.

I think it should be:
```
    ....
    if (qle_infos.qle_nb == 0) {
        _z_keyexpr_clear(&qle_infos.ke_out);
#if Z_FEATURE_RX_CACHE != 1
        _z_queryable_infos_svec_release(&qle_infos.infos);  // Otherwise it's released with cache
#endif
        return _Z_RES_OK;
    }
    ....
```

Is that right?
Thanks

### To reproduce

.

### System info

.

jean-roland:
Fixes #879 

Fix a raised assertion introduced with #877 on `z_bytes_to_string` api call.


CMakeLists.txt
```diff
@@ -243,7 +243,7 @@ set(Z_FEATURE_TCP_NODELAY 1 CACHE STRING "Toggle TCP_NODELAY")
 set(Z_FEATURE_LOCAL_SUBSCRIBER 0 CACHE STRING "Toggle local subscriptions")
 set(Z_FEATURE_SESSION_CHECK 1 CACHE STRING "Toggle publisher/querier session check")
 set(Z_FEATURE_BATCHING 1 CACHE STRING "Toggle batching")
-set(Z_FEATURE_MATCHING 1 CACHE STRING "Toggle matching feature")
+set(Z_FEATURE_MATCHING 0 CACHE STRING "Toggle matching feature")
```
sashacmc, line 246:
```suggestion
set(Z_FEATURE_MATCHING 1 CACHE STRING "Toggle matching feature")
```

include/zenoh-pico/config.h
```diff
@@ -47,7 +47,7 @@
 #define Z_FEATURE_LOCAL_SUBSCRIBER 0
 #define Z_FEATURE_PUBLISHER_SESSION_CHECK 1
 #define Z_FEATURE_BATCHING 1
-#define Z_FEATURE_MATCHING 1
+#define Z_FEATURE_MATCHING 0
```
sashacmc, line 50:
```suggestion
#define Z_FEATURE_MATCHING 1
```

sashacmc (approved):
Removed some unexpected changes, other LGTM

CMakeLists.txt
jean-roland: Since unstable is OFF by default we fall under:

```

if(Z_FEATURE_MATCHING AND NOT Z_FEATURE_UNSTABLE_API)
  message(WARNING "Z_FEATURE_MATCHING can only be enabled when Z_FEATURE_UNSTABLE_API is also enabled. Disabling Z_FEATURE_MATCHING.")
  set(Z_FEATURE_MATCHING 0 CACHE STRING "Toggle matching feature" FORCE)
endif()
```
So Matching ON by default triggers a visible warning to the user and it get turned off anyway.

PR #895, opened 2025-02-12
Fix z_collections_test

sashacmc:
Fix int_map_iterator_deletion_test in z_collections_test


gmartin82 (approved):
Looks good. Tests pass without segfault with fix applied.

PR #900, opened 2025-02-21
Refactor unicast

jean-roland:
Intermediary PR, it contains groundwork to break the assumption that unicast => client and that every data comes from the link socket.

Does not introduce new feature, doesn't impact measurably performance of unicast client mode.


gmartin82 (approved):
Looks good to me.

PR #902, opened 2025-02-26
Add tcp listen capabilities

jean-roland:
Add a possibility to listen to a tcp socket with zenoh pico.

Still rough around the edges (doesn't handle high volume traffic) and tx side isn't working yet.

### DIFF
--- a/src/collections/list.c
+++ b/src/collections/list.c
@@ -14,6 +14,7 @@
 
 #include "zenoh-pico/collections/list.h"
 
+#include <assert.h>
 #include <stddef.h>
 
 /*-------- Inner single-linked list --------*/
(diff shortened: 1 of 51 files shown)


gmartin82 (approved):
Looks good.

src/collections/list.c
```diff
@@ -90,6 +91,18 @@ _z_list_t *_z_list_find(const _z_list_t *xs, z_element_eq_f c_f, const void *e)
     return ret;
 }
 
+_z_list_t *_z_list_drop_element(_z_list_t *list, _z_list_t *elem, _z_list_t *prev, z_element_free_f f_f) {
+    if (list == elem) {  // Head removal
```
kogratte, line 95: Dumb reader question.
Looks like we try to remove an element from a list, and in order to achieve, we call this function with the element we would like to remove, but also the previous element in list.
My dumb question is: why should we use this method instead of a double chained list?

I think I understand from the prev->tail = elem->tail that the list in itself seems to be persistent in memory (maybe for performance purposes), but I don't see the point using such signature.

Would you mind enlighten the old C practicer I was a long time ago?

kogratte: @JEnoch maybe?

kogratte: Or better, @jean-roland :)

jean-roland: 1) That's the type I have on hand so it's easier for me to use this and it suits the use case, but yeah for this particular operation a double linked list is easier to use.
2) Memory constraint as double linked list has one more pointer per element.

PR #905, opened 2025-02-28
Fix null allocation in msgcodec

jean-roland:
* Unit tests were not run as DEBUG on CI.
* When run as DEBUG, issues related to the null allocation change (#877) appeared and are fixed with this PR.


gmartin82 (approved):
Looks good. Fixes the test failure and ensures tests are run in debug mode through the CI which means assert's in the main codebase will be triggered if violated.

PR #906, opened 2025-02-28
Fix config issue

jean-roland:
A config token had an obsolete name in the header files.

Add a config token for testing purposes.


gmartin82 (approved):
Looks good.

PR #908, opened 2025-03-05
Update unicast tx

jean-roland:
Rework lease task and tx layer to handle list of peers.

Functional but misses implementation of recursive mutex to insure thread safety that will come in next PR. Write flter doesn't work well also.


gmartin82 (approved):
Looks good to me.

PR #913, opened 2025-03-11
fix _z_open_serial_from_dev in zephyr

XxChang:
The `ret` variable is not returned, which will cause undefined behavior.


jean-roland (approved):

jean-roland:
Yeah it was pretty clear that the code format wasn't correct, you'll need to run clang-format on your PR for it to be approved.

XxChang:
Great! It's my oversight, tks.

PR #914, opened 2025-03-11
Allow multiple connect

jean-roland:
Allow a list of connect endpoints and deactivate write_filters for peer mode as they're not working properly.


gmartin82 (approved):
Looks good.

PR #925, opened 2025-04-01
Fix rpi pico listen config

jean-roland:
lwip config was missing options to allow for listen capabilities on rpi pico


jean-roland:
Yeah, it's getting old...

Issue #919, opened 2025-03-21
Make zenoh-pico compilation open-ended on platforms


vortex314:
### Describe the feature

The zenoh-pico source tree now contains certain combinations of platforms and frameworks that are supported.
When trying to port this to other platforms , I needed to fork the source tree and change some files
Suggestion : if the platform combination is not known or supported just warn this and don't block 
Needed changes : 
- extra_script.py -> should ignore all platform specific source files if unknown ZENOH_GENERIC platform is encountered
- platform.h should give a "#warn" instead of a "#error"
- platform.h should refer to a "zenoh_generic.h" include 
- library.json should list all platforms known to platformio ( maybe there is a way to override this )

Doing this permits to keep system.cpp, network.cpp, zenoh_generic.h at the developer side 
I used this in the fork :  https://github.com/vortex314/zenoh-pico/tree/main

The result  is that a pio project can include zenoh-pico and make the effort of the porting without a need of forks and local copies. The config.h settings remain a challenge however

Example project : https://github.com/vortex314/zenoh-projects/tree/main/zenoh-stm32

Example platformio.ini
```ini
[env:stm32_stm32cube]
platform = ststm32
board = nucleo_f401re
framework = stm32cube
monitor_speed = 115200

lib_deps = 
	lib/zenoh-pico
	lib/stm32_stm32cube

build_flags = 
	-D__GLIBCXX_ASSERTIONS
	-DZENOH_GENERIC
	-DZENOH_DEBUG=3
	-std=gnu++17
	-Wno-error=deprecated-declarations
	-Ilib/stm32_stm32cube
	-g
;	-Os


build_unflags = 
	-std=gnu++11
```

vortex314:
See also : #835

vortex314:
See also : #818

vortex314:
See also : #241

jean-roland:
I agree to the need to do something on this front. Feel free to open a PR with your work on this.

vortex314:
See pull request. Worked at least for the Texas instruments Tiva lm4f120 and strm32cube. 
Let me know what you think about the approach.

jean-roland:
Fixed with #933

PR #927, opened 2025-04-02
Zenoh generic

vortex314:
An attempt to make the build of zenoh-pico open-ended on platforms,

See : https://github.com/vortex314/zenoh-pico/blob/zenoh-generic/generic_platform.md

on the approach. 

See issue : https://github.com/eclipse-zenoh/zenoh-pico/issues/919


vortex314:
Not clear to me , how to add a label to this pull request.

vortex314:
@bot add enhancement

include/zenoh-pico/config.h
```diff
@@ -15,29 +15,33 @@
...
+
 /*--- CMake generated config; pass values to CMake to change the following tokens ---*/
 #define Z_FRAG_MAX_SIZE 4096
 #define Z_BATCH_UNICAST_SIZE 2048
 #define Z_BATCH_MULTICAST_SIZE 2048
 #define Z_CONFIG_SOCKET_TIMEOUT 100
 
 /* #undef Z_FEATURE_UNSTABLE_API */
-#define Z_FEATURE_MULTI_THREAD 1
+#define Z_FEATURE_MULTI_THREAD 0
```
jean-roland, line 29: All the following values should not be committed

jean-roland (changes requested):

include/zenoh-pico/config.h
vortex314: Indeed, as I tried different approaches, it still contains some traces of that, as the new approach overrides externally these settings, They can remain as such. Other things ? Looks like a valid approach ?

vortex314:
Why are all these checks failing ?

include/zenoh-pico/config.h
jean-roland: Yeah I think the approach is sound.

jean-roland:
ECA is something you need to do on your own if you've done it already then it is a bug from their end and just need to retart the process. 

For the other processes looks like your change had an impact on the includes for these platforms and now they are missing some definitions and stuff

jean-roland (approved):

Issue #909, opened 2025-03-06
[Bug] Write filters only works with a single entity


jean-roland:
### Describe the bug

The way they are implemented, they works well in client mode but in p2p, if I have one publisher and two subscribers, as soon as one subscriber leaves (undeclare). It triggers the write filter.

Needs to add some sort of counter to keep track of number of declare/undeclare.

### To reproduce

Start a publisher on peer mode, then two subscribers, then stop one subscriber.

### System info

NA

PR #929, opened 2025-04-04
Listener refuse client connections.

jean-roland:
Because of #909 clients with automatic resource declaration causes memory issues on a peer unicast listener so for now a listener will refuse connection coming from client nodes.


gmartin82 (approved):
Looks good.

PR #934, opened 2025-04-07
deprecate ZENOH_DEBUG for ZENOH_LOG + add WARN/TRACE levels

wyfo:


sashacmc:
I think it would be better to introduce text logging level (info, error, debug, etc.) by CMake variable ZENOH_LOG.
And translate it to numeric C define ZENOH_LOG_LEVEL. Also it is better to set it at intervals to have potential for expansion.
10 - error
20 - warning
30 - info
40 - debug
50 - trace

sashacmc:
Another wish is to move printf as a macro so that it can be easily replaced when debugging on embedded platforms (printf is not available everywhere) by changing only one line of code.

.github/workflows/arduino_esp32.yaml
```diff
@@ -51,7 +51,7 @@ jobs:
 
           mkdir -p $ARDUINO_BASE
           cd $ARDUINO_BASE
-          platformio init -b esp32thing_plus -O "board_build.cmake_extra_args=-DZ_FEATURE_LINK_BLUETOOTH=1 -DZ_FEATURE_LINK_SERIAL=1" -O "build_flags=-DZENOH_DEBUG=3 -DZENOH_COMPILER_GCC" -O "lib_ldf_mode=deep+"
+          platformio init -b esp32thing_plus -O "board_build.cmake_extra_args=-DZ_FEATURE_LINK_BLUETOOTH=1 -DZ_FEATURE_LINK_SERIAL=1" -O "build_flags=-DZENOH_LOG_LEVEL=40 -DZENOH_COMPILER_GCC" -O "lib_ldf_mode=deep+"
```
sashacmc, line 54: IMHO, we should totally avoid expose this numeric constants outside of the code. If it is not possible to use ZENOH_LOG there, let's leave it as is with legacy ZENOH_DEBUG.

docs/api.rst
```diff
@@ -1542,14 +1542,25 @@ Zenoh-Pico supports three logging levels:
 Enabling Logging
 ----------------
 
-To enable logging, you can adjust the logging level by defining the ``ZENOH_DEBUG`` macro at compile time.
+To enable logging, you can adjust the logging level by defining the ``ZENOH_LOG_LEVEL`` macro at compile time.
 
-- Set ``ZENOH_DEBUG`` to ``1`` to enable error messages.
-- Set ``ZENOH_DEBUG`` to ``2`` to enable informational messages.
-- Set ``ZENOH_DEBUG`` to ``3`` to enable debug messages (includes info and error).
+- Set ``ZENOH_LOG_LEVEL`` to ``10`` or greater to enable error messages.
```
sashacmc, line 1547: See previous comment on it

.github/workflows/arduino_esp32.yaml
wyfo: It would be possible to use `ZENOH_LOG`, but it would rely on the compiler to optimize lines like `"DEBUG" == "DEBUG" || "DEBUG" == "debug"`, so I'm not sure that we want to do so.
Regarding `ZENOH_LOG_LEVEL` hiding, then you force to use the legacy system, which don't become legacy this way, so I would go against your opinion here.

jean-roland: I think the point is just to avoid digits and have more something like `-DZENOH_LOG_LEVEL=LOG_LVL_TRACE`?

wyfo: And from where would come `LOG_LVL_TRACE`? Why do we want to avoid digits?

jean-roland: I can't speak for Sasha rationale but if I am in user's shoes, it's easier for me, beyond the myriad of other things in my mind, to remember something like "TRACE, DEBUG, WARNING" values that are generic rather than arbitrary numbers that are specific to this lib only.

Now how to make it work, this is another problem.

sashacmc: Yes, I would like to avoid any "magic numbers" and especially at the API level
I thought we had the opportunity to transfer everything to CMake. In what case should users define C definition directly?
In this particular case we can do like:
```
platformio init -b esp32thing_plus -O "board_build.cmake_extra_args=-DZ_FEATURE_LINK_BLUETOOTH=1 -DZ_FEATURE_LINK_SERIAL=1 -DZENOH_LOG=debug" -O "build_flags=-DZENOH_COMPILER_GCC" -O "lib_ldf_mode=deep+"
```

wyfo: `api.rst` gives examples with `gcc` command, so they use `ZENOH_LOG_LEVEL`; which command should I put in them if it's hidden?
I'm fine with `cmake_extra_args`, I will update the PR accordingly.

sashacmc: If we need to support compilation without CMake, I can offer a definition option via
-DZENOH_LOG_LEVEL_TRACE, -DZENOH_LOG_LEVEL_DEBUG in place of numbers
In this case we also hide numbers in CMake, and it will be like:

```
if (DEFINED ZENOH_DEBUG AND NOT (ZENOH_DEBUG STREQUAL ""))
  pico_add_compile_definition(ZENOH_DEBUG=${ZENOH_DEBUG})
elseif (DEFINED ZENOH_LOG_LEVEL AND NOT (ZENOH_LOG_LEVEL STREQUAL ""))
  pico_add_compile_definition(ZENOH_LOG_LEVEL=${ZENOH_LOG_LEVEL})
elseif (ZENOH_LOG STREQUAL ERROR OR ZENOH_LOG STREQUAL error)
  pico_add_compile_definition(ZENOH_LOG_LEVEL_ERROR=1)
elseif (ZENOH_LOG STREQUAL WARN OR ZENOH_LOG STREQUAL warn)
  pico_add_compile_definition(ZENOH_LOG_LEVEL_WARN=1)
...
endif()
```

This will slightly complicate the logic of checking variables in macros, but it will still happen during compilation.

wyfo: It should be better now

include/zenoh-pico/utils/logging.h
```diff
@@ -23,58 +23,83 @@
...
+#elif ZENOH_DEBUG == 3
+#define #ZENOH_LOG_DEBUG
+#endif
+#endif
 
-// Logging macros
-#define _Z_LOG_PREFIX(prefix) \
-    __z_print_timestamp();    \
-    printf(#prefix " ::%s] ", __func__);
+#define _Z_CHECK_LOG(...) do { if (false) ZENOH_LOG_PRINT(__VA_ARGS__); } while (false)
```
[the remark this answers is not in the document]
wyfo: Oups, I forget to put back the comment explaining it. The goal is to check the arguments you pass to `_Z_TRACE` even when *trace* level is disabled, to not have surprise the day you activate trace on release mode someday. If you look at the diff, there was a different way of doing before, but I prefer this one as it doesn't impact the code expanded on release build.

sashacmc (approved):
Ok, LGTM
Last note, could you please fill PR description

@jean-roland could you please also take a look and test this PR with some embedded device?

sashacmc (approved):

jean-roland:
PR LGTM, still need to run it on embedded platforms before approving.

jean-roland:
Ok so it seems `board_build.cmake_extra_args=` arguments are not taken into account. More specifically, our CMakeLists is not run which means the pre compiler value is not set. Going through ZENOH_DEBUG works because it sets the precompiler value directly.
 I don't think this is an issue related to this PR though so I suggest merging it and fixing this issue afterwards.

jean-roland (approved):

sashacmc:
@jean-roland did you check this PR with the real embedded platform?

PR #946, opened 2025-04-23
Rework lru cache insertion

jean-roland:
Change to make it O(log n) instead of O(n).

### DIFF
--- a/src/collections/lru_cache.c
+++ b/src/collections/lru_cache.c
@@ -149,7 +149,29 @@ static int _z_lru_cache_find_position(_z_lru_cache_node_t **slist, _z_lru_val_cm
     return start;
 }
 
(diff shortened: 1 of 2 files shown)


wyfo:
I'm starting to review the full implementation, but it seems the insertion could be made more efficient. Indeed, when inserting an element, a first memmove is done after removing the last item, and then a second memmove is done to insert the item.
By looking for the indexes of both the item to remove and the one to insert first, you could then analyse what would be the most efficient moves to do.
For example
```
| 0 | 1 | 2 | 5 | 6 | 7 | 9 | 10 | 11 |
# to remove 7 and insert 3, 
# you only need to move 5-6 to the right, not the rest of the array
```

gmartin82 (approved):
Looks good.

PR #953, opened 2025-04-29
Use lowest lease value

jean-roland:
Currently we are taking the received lease value to calculate keep alive period, we should take the minimum of the two instead.


gmartin82 (approved):
Looks good.

PR #958, opened 2025-05-15
Add missing malloc type cast

jean-roland:


gmartin82 (approved):
Looks good, types match declared types.

PR #959, opened 2025-05-19
Refactor resources & write filter for peer to peer unicast

jean-roland:
The initial implementation of resource declaration and write filter could only work in the simpler client to router mode, it had to be reworked entirely to make it work with peer to peer unicast (multicast requires even more work on top of that).


gmartin82 (approved):
Looks good.

PR #960, opened 2025-05-19
Fix multicast write filter segfault

jean-roland:
Since we don't create a write filter in multlicast for now, we shouldn't use or drop it.


gmartin82 (approved):
Looks good.

PR #961, opened 2025-05-22
Remove frame eager deserialization

jean-roland:
Original idea of @wyfo who implemented this on Zenoh. Reduce memory consumption on large frames.

Also remove the refusal of CLIENT connections to a PEER now that resource declaration work on p2pu


gmartin82 (commented):
Changes look good other than 1 CI job failure. It looks like the same test is passing on other runs though so could just be a random failure?

gmartin82 (approved):
Looks good.

PR #966, opened 2025-05-28
Add multicast resource declaration / write filter

jean-roland:
* Add a config token that enables automatic resource declarations and write filters.

To allow this, a late joining node will pull the declarations with an interest, this will cause a lot of traffic as each previous node will send all its declarations so the scaling is O(n) hence the config token that is off by default.

* Reworked multicast testing and added a no liveliness build test.
* Interest protocol now handles the restricted flag.


gmartin82 (commented):
Looks good, just a few questions on whether guards are needed in a few places.

PR #967, opened 2025-05-30
New "batch" of performance optimization

jean-roland:
Included in this PR:
* A lot of helper function explicit inlining
* Optimization in the codec/serialization layer
* Pass by reference instead of value on critical path for "bigger" types
* Remove the "fix_mapping" operation
* The introduction of batch compile options to allow a batch to hold the mutexes (has side effects !)


gmartin82 (commented):
Looks good. My only concern is the effect on sizing from the increased inlining. Would be good to understand what effect this has for more constrained platforms.

jean-roland:
Good point, I'll check the binary size difference with and without this PR on stm32 & esp-idf

PR #969, opened 2025-06-02
STM32 ThreadX target platform

g4sp3r:
Hi, 
At [Ubiquity robotics](https://github.com/UbiquityRobotics) we are happy users of zenoh-pico running on multiple of our stm32 based embedded projects :heart_eyes:

We would like to contribute back this pull request that adds new STM32 platform with ThreadX system implementation. 

Currently only supports serial transport. Implementation is with circular DMA for data reception.

### DIFF
--- a/examples/threadx_stm32/z_pub.c
+++ b/examples/threadx_stm32/z_pub.c
@@ -61,7 +61,7 @@ VOID start_example_thread(ULONG initial_input) {
     }
 
     // Publish data
-    while (true) {
+    while (1) {
         char buf[256];
         for (int idx = 0; 1; ++idx) {
             z_sleep_s(1);
--- a/examples/threadx_stm32/z_sub.c
+++ b/examples/threadx_stm32/z_sub.c
@@ -75,7 +75,7 @@ VOID start_example_thread(ULONG initial_input) {
         return;
     }
 
-    while (true) {
+    while (1) {
         z_sleep_s(1);
     }
 


jean-roland:
Hello @g4sp3r, thanks for your contribution, it seems the PR has two issues currently:
* Formatting is not correct, please run `find include/ src/ tests/ examples/ tools/ -iname "*.ino" -o -iname "*.h" -o -iname "*.c" | xargs clang-format -i` in your zenoh-pico repo, you'll need to install clang-format if you haven't.
* Your threadX files seem to be visible from other systems, make sure the includes are guarded against that.

jean-roland:
Highly possible the cpp CI failure is due to a flaky test and not related to your changes, so don't worry about it for now.

README.md
```diff
@@ -436,6 +436,44 @@ e.g.
 zenohd -l serial//dev/ttyACM1#baudrate=112500
 ```
 
+#### 2.2.7. STM32 Threadx
```
sashacmc:
```suggestion
#### 2.2.7. STM32 ThreadX
```

README.md
```diff
@@ -436,6 +436,44 @@ e.g.
...
+2. In CubeMX project configuration tool:
+    - Add and enable AZURE-RTOS(threadx) Middleware.
+    - Enable UART periphery, set up RX DMA in circular mode and enable UART global interrupt.
+    - Move HAL_Tick to TIM peripheral for threadx to work with HAL. (SYS - Timebase source configuration)
+
+3. Generate initialization code with CubeMX.
+
+4. Clone zenoh-pico repository to your project folder (or submodule, copy files)
+
+5. Add zenoh-pico/src to project source folders. Exclude any folders in platforms/* except  common and threadx/stm32.
```
sashacmc:
```suggestion
5. Add zenoh-pico/src to project source folders. Exclude any folders in platforms/* except common and threadx/stm32.
```

examples/threadx_stm32/z_pub.c
```diff
@@ -0,0 +1,101 @@
+/**
+ ******************************************************************************
+ * @file    z_pub.c
+ * @brief   Example zenoh-pico publisher on STM32 running Threadx
```
sashacmc:
```suggestion
 * @brief   Example zenoh-pico publisher on STM32 running ThreadX

```

examples/threadx_stm32/z_pub.c
```diff
@@ -0,0 +1,101 @@
...
+    z_owned_publisher_t pub;
+    z_view_keyexpr_t ke;
+    z_view_keyexpr_from_str_unchecked(&ke, KEYEXPR);
+    if (z_declare_publisher(z_loan(s), &pub, z_loan(ke), NULL) < 0) {
+        _LOG("Unable to declare publisher for key expression!\n");
+        return;
+    }
+
+    // Publish data
+    for (;;) {
```
sashacmc:
```suggestion
    while (1) {
```

examples/threadx_stm32/z_sub.c
```diff
@@ -0,0 +1,104 @@
+/**
+ ******************************************************************************
+ * @file    z_sub.c
+ * @brief   Example zenoh-pico subscriber on STM32 running Threadx
```
sashacmc:
```suggestion
 * @brief   Example zenoh-pico subscriber on STM32 running ThreadX
```

examples/threadx_stm32/z_pub.c
jean-roland: I'm going to counter nitpick that with `while(true)` because of MISRA Rule 14.4

The controlling expression of an if statement and the controlling expression of an iteration-statement shall have essentially Boolean type

sashacmc: Yes, but some old compilers still don't support this, plus in other examples we have it like it, and better when it's the same everywhere for the code consistency.

g4sp3r:
Hey @sashamc! Thanks for your fixes. They are accepted and PR is updated.

About the project setup, I am also not a fan and understand your concerns.

In my experience, most of the users will probably work with STMCubeIDE (eclipse) using CDT managed build and gui tools to configure their project and peripherals. This is also how our projects was configured from start and is also the way most of the examples for stm32 you can find online are done.

You can see the examples in this PR don't do any peripheral or kernel initialization so user needs to provide those layers by themselves. But this means examples could be run on any target that the manufacturer toolchain supports.

The main issue I see for CMake support is that the landscape is wider than for other platforms. To give some examples:
- Correct hal drivers would need to be downloaded from ST for which license agreements need to be accepted.
- threadx low level initialization is done in assembly and generated by the project configuration tool.

I am not saying its impossible to do a CMake build for stm32 target, I just think this way will be better suited for majority of users.

At some point support for libopencm3 can be added but it will surely support less targets.

Maybe the instructions are not clear, but zenoh-pico could also be located outside your project folder. What is required is adding correct source and include paths to the project configuration. The exclusion pattern is used because it requires less operations from user than adding each folder in src/ individually. If all the system.c and network.c files from every platform would have #ifdef guards, then this exclusion pattern would not be needed.

PR #978, opened 2025-06-27
Use svec in wbuf

jean-roland:
Using svec in wbuf reduces memory allocation around fragmentation buffers.


gmartin82 (approved):
Looks good.

PR #986, opened 2025-07-10
Fix compilation issues

jean-roland:
When compiling in debug with features deactivated.


include/zenoh-pico/utils/string.h
```diff
@@ -63,6 +64,12 @@ size_t _z_strcnt(char const *haystack_start, const char *harstack_end, const cha
 
 size_t _z_str_startswith(const char *s, const char *needle);
 
+static inline void _z_str_append(char *str, const char c) {
```
gmartin82, line 68: This might be worth commenting due to its requirement that str[strlen(str) + 1] is valid memory.

I can see it is used correctly but I think a note for future function users may be useful.

gmartin82 (approved):
Looks good.

PR #987, opened 2025-07-15
Add configuration options documentation

jean-roland:


docs/config.rst
```diff
@@ -0,0 +1,145 @@
...
+-----------
+
+Defines a single endpoint a node will listen on.
+
+* `Z_CONFIG_LISTEN_KEY`: The index of the option in the config table.
+
+Scouting
+-----------
+
+Scouting is used when a node isn't provided without connect or listen endpoint.
```
gmartin82: Should be with rather than without?

jean-roland: Yeah I did a double negative it seems.

PR #1004, opened 2025-07-29
*IMPORTANT* Change default behavior of zp_read

jean-roland:
Default behavior of zp_read is now to read all messages in the buffer instead of a single one, added a "single_read" option to keep original behavior possible.

### DIFF
--- a/include/zenoh-pico/api/types.h
+++ b/include/zenoh-pico/api/types.h
@@ -450,7 +450,7 @@ typedef struct {
  * Represents the configuration used to configure a read operation started via :c:func:`zp_read`.
  */
 typedef struct {
-    uint8_t __dummy;  // Just to avoid empty structures that might cause undefined behavior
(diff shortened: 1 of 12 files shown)


sashacmc:
If we change default behavior of zp_read, we need to make sure that this will not cause negative side effects. It seems that it will not and for many it will be just a performance boost, but we need to check at least by running standard performance tests.
And also we need to add enough information to the PR description so that we don't forget to write about it in a blogpost.

gmartin82 (approved):
Looks good to me, pending the check on performance impact as @sashacmc suggested.

jean-roland:
So throughput cannot be tested because this method is slow and prevents protection of tcp buffer overflow which means we get corrupted data. If anything the single read being even slower, it gets corrupted data right from the start while the multiple read successfully get some throughput measurement before shutting down.

Latency is unchanged but that's because we send the packets one by one.

PR #1008, opened 2025-08-04
Add error log functions

jean-roland:
Activated with trace log level, very useful when tracing the origin of an error code and will allows to add function call in the future for notifications or keeping history of errors.


gmartin82 (approved):
Looks good.

PR #1009, opened 2025-08-05
Repeat the ZENOH_COMPILER_GCC flag for espressif32

bmagyar:
Fixes "Multi-thread refcount in C99 only exists for GCC, use GCC or C11 or deactivate multi-thread" when compiling for an ESP32


bmagyar:
The ECA thing is definitely not making it easy to contribute. I've created an account, signed it and linked my github to it as well.

jean-roland (approved):

PR #1016, opened 2025-08-19
ci: reorg tests to run from ci.yml

diogomatsubara:


milyin:
The test fails are unrelated to changes, the dummy pr in zenoh-cpp have the same problems
https://github.com/eclipse-zenoh/zenoh-cpp/actions/runs/17098619348/job/48488989677?pr=582

PR #1024, opened 2025-09-01
Allow subscriptions to key expressions beginning with a wildcard.

### LINKED ISSUE #1023: [Bug] Unable to subscribe to a key expression starting with a wildcard
### Describe the bug

It's not currently possible to subscribe to valid key expressions which begin with a wildcard character. When trying to create a subscriber with such a key expression create fails.

### To reproduce

1. Run the subscriber:

$ ./examples/z_sub -k **/example/zenoh-pico-pub

or

$ ./examples/z_sub -k */example/zenoh-pico-pub

### System info

- Platform: Ubuntu 24.04
- Zenoh version: 1.3.4+.

gmartin82:
Closes #1023.


sashacmc (changes requested):
Could you please add an unittest to cover this case?

sashacmc (approved):

PR #1025, opened 2025-09-01
Add Mbed TLS support

steils:
Enable TLS transport so users can connect via `tls/<host>:<port>` locators.

TLS options set through the configuration with the following keys:

- Z_CONFIG_TLS_ROOT_CA_CERTIFICATE_KEY: path to the CA bundle used to validate the remote certificate (required).
- Z_CONFIG_TLS_ROOT_CA_CERTIFICATE_BASE64_KEY: inline/base64 CA bundle (alternative to the file path).
- Z_CONFIG_TLS_LISTEN_PRIVATE_KEY_{KEY,BASE64_KEY}: listening side private key (file path or base64) required when listening for TLS peers.
- Z_CONFIG_TLS_LISTEN_CERTIFICATE_{KEY,BASE64_KEY}: listening side certificate presented to clients (file path or base64).
- Z_CONFIG_TLS_ENABLE_MTLS_KEY: set to true to require mutual TLS (client authentication); disabled by default.
- Z_CONFIG_TLS_CONNECT_PRIVATE_KEY_{KEY,BASE64_KEY}: client private key (file path or base64) sent when mTLS is enabled.
- Z_CONFIG_TLS_CONNECT_CERTIFICATE_{KEY,BASE64_KEY}: client certificate (file path or base64) sent when mTLS is enabled.
- Z_CONFIG_TLS_VERIFY_NAME_ON_CONNECT_KEY: set to false to skip CN/SAN hostname checks; enabled by default.

Example:

```
zp_config_insert(z_loan_mut(cfg), Z_CONFIG_CONNECT_KEY, "tls/127.0.0.1:7447");
zp_config_insert(z_loan_mut(cfg), Z_CONFIG_TLS_ROOT_CA_CERTIFICATE_KEY, "/etc/ssl/certs/ca.pem");
zp_config_insert(z_loan_mut(cfg), Z_CONFIG_TLS_ENABLE_MTLS_KEY, "true");
zp_config_insert(z_loan_mut(cfg), Z_CONFIG_TLS_CONNECT_CERTIFICATE_KEY, "client.pem");
zp_config_insert(z_loan_mut(cfg), Z_CONFIG_TLS_CONNECT_PRIVATE_KEY_KEY, "client.key");
```

TLS build is enabled with Z_FEATURE_LINK_TLS=1 (disabled by default).

Currently available for Unix platform.


diogomatsubara:
internally tracked as [ZEN-655]

include/zenoh-pico/system/platform/unix.h
```diff
@@ -47,6 +47,9 @@ typedef struct {
         int _fd;
 #endif
     };
+#if Z_FEATURE_LINK_TLS == 1
+    void *_tls_sock;  // Pointer to _z_tls_socket_t
```
sashacmc, line 51: Can `_fd` and `_tls_sock` be used together? Shouldn't it be moved to union?

src/link/unicast/tls.c
```diff
@@ -0,0 +1,280 @@
...
+#include "zenoh-pico/config.h"
+#include "zenoh-pico/link/link.h"
+#include "zenoh-pico/link/manager.h"
+#include "zenoh-pico/system/link/tls.h"
+#include "zenoh-pico/utils/logging.h"
+#include "zenoh-pico/utils/pointers.h"
+
+#if Z_FEATURE_LINK_TLS == 1
+
+static char *__z_parse_port_segment_tls(_z_string_t *address) {
```
sashacmc: I see that for TCP and UDP, address parsing was simply copy/paste, but I don't think it's worth doing it a third time. It seems time to move these methods out and reuse the code.

src/system/unix/tls.c
```diff
@@ -0,0 +1,547 @@
...
+    int ret = mbedtls_ssl_write(&sock->_tls_ctx->_ssl, ptr, len);
+    if (ret > 0) {
+        return (size_t)ret;
+    }
+
+    if (ret == MBEDTLS_ERR_SSL_WANT_READ || ret == MBEDTLS_ERR_SSL_WANT_WRITE) {
+        return 0;
+    }
+
+    _Z_ERROR("40 %s TLS write error: -0x%04x", __func__, -ret);
```
sashacmc: `40` looks like some artifact.

src/system/unix/tls.c
```diff
@@ -0,0 +1,547 @@
...
+    if (verify_opt != NULL) {
+        if (verify_opt[0] == '0' || verify_opt[0] == 'n' || verify_opt[0] == 'N' || verify_opt[0] == 'f') {
+            verify_name = false;
+        }
+    }
+
+    bool enable_mtls = false;
+    const char *mtls_opt = _z_str_intmap_get(config, TLS_CONFIG_ENABLE_MTLS_KEY);
+    if (mtls_opt != NULL) {
+        if (!(mtls_opt[0] == '0' || mtls_opt[0] == 'n' || mtls_opt[0] == 'N' || mtls_opt[0] == 'f')) {
```
sashacmc: String parsing for boolean config values is repeated. Please consider helper function.

src/link/unicast/tls.c
```diff
@@ -0,0 +1,280 @@
...
+        }
+    }
+
+    if (ret == _Z_RES_OK) {
+        char *s_port = __z_parse_port_segment_tls(&endpoint->_locator._address);
+        if (s_port == NULL) {
+            _Z_ERROR_LOG(_Z_ERR_CONFIG_LOCATOR_INVALID);
+            ret = _Z_ERR_CONFIG_LOCATOR_INVALID;
+        } else {
+            uint32_t port = (uint32_t)strtoul(s_port, NULL, 10);
```
sashacmc: Please add proper handling for non-numeric input, like
```
char *endptr = NULL;
errno = 0;
uint32_t port = (uint32_t)strtoul(s_port, &endptr, 10);
if ((errno != 0) || (endptr == s_port) || (*endptr != '\0') || (port < 1) || (port > 65535)) 
```

src/system/unix/tls.c
```diff
@@ -0,0 +1,547 @@
...
+#include "zenoh-pico/config.h"
+#include "zenoh-pico/link/config/tls.h"
+#include "zenoh-pico/system/link/tcp.h"
+#include "zenoh-pico/utils/logging.h"
+#include "zenoh-pico/utils/pointers.h"
+
+#ifdef ZENOH_LOG_TRACE
+static void _z_tls_debug(void *ctx, int level, const char *file, int line, const char *str) {
+    _ZP_UNUSED(ctx);
+    _Z_DEBUG("mbed TLS [%d] %s:%04d: %s", level, file, line, str);
```
sashacmc: Looks like _z_tls_debug generate additional EOL, please remove it to have more clear log

src/transport/raweth/rx.c
```diff
@@ -82,12 +82,12 @@ z_result_t _z_raweth_recv_t_msg_na(_z_transport_multicast_t *ztm, _z_transport_m
     // Prepare the buffer
     _z_zbuf_reset(&ztm->_common._zbuf);
 
-    switch (ztm->_common._link._cap._flow) {
+    switch (ztm->_common._link->_cap._flow) {
```
sashacmc, line 85: Ok, for now, but for the future this kind of refactoring better to do as a separate PR.

include/zenoh-pico/system/platform/unix.h
steils: I think we need them both at the same time. For example _fd is used _z_tls_bio_send/_recv (the context is _fd, taken from _z_sys_net_socket_t), in src/system/unix/network.c (_z_socket_wait_event() is called during TLS connections in peer mode, for example), and many other places.

milyin:
Agree with @sashacmc . In the main zenoh these certificate options are part of .json config file. It would be nice to unify zenoh-pico and zenoh-c config APIs in the future: for example to make zenoh-c support options from zenoh-pico and translate them to the zenoh's json options. So any change which makes **values** of options incompatible makes this further unification harder.

PR #1030, opened 2025-09-02
Ensure interest declarations are resent on session reconnect.

### LINKED ISSUE #1029: [Bug] Interests are not re-propagated on reconnect
### Describe the bug

zenoh-pico clients don't maintain a store of session interests, so interests are not sent to their peer/router upon reconnection. 

This means that if a subscriber for example arrives after a zenoh-pico client connects and reconnects, the write filter on the zenoh-pico client will filter out publications since the peer/router wouldn't have propagated any subscription to it.

Interests need to be re-propagated and the write filters need to be managed accordingly (see https://github.com/eclipse-zenoh/zenoh/blob/main/zenoh/src/net/routing/hat/client/interests.rs#L40 for the relevant zenoh-rust code).

### To reproduce

N/A

### System info

- ref: 81a61e4293fcc9ccedd2c0f16195b51543833f96

gmartin82:
Fixes #1029


tests/connection_restore.py
```diff
@@ -232,6 +233,137 @@
...
+
+        # Restart router
+        print("Restarting router...")
+        terminate_processes(router_ps)
+        time.sleep(timeout)
+
+        # Both client logs should eventually show disconnect; don't hard-fail if only one shows it
+        try:
+            wait_disconnect(pub_output);   pub_output.clear()
+        except Exception:
```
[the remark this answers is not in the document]
gmartin82: wait_disconnect throws a general exception, hence why the code catches one.

tests/connection_restore.py
```diff
@@ -232,6 +233,137 @@
...
+            pass
+        try:
+            wait_disconnect(sub_output);   sub_output.clear()
+        except Exception:
```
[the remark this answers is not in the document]
gmartin82: wait_disconnect throws a general exception, hence why the code catches one.

oteffahi (approved):
LGTM.

If possible, it would be good to add a test that checks that writer-side filtering is re-enabled in the case where a subscriber was present in the network before the router restarts, but is not present after the restart. Although I'm not sure how this can be observed in the tests without using a packet sniffer or adding a log on the router on reception of a push message that does not have a matching destination.

src/net/primitives.c
```diff
@@ -563,6 +566,9 @@ z_result_t _z_remove_interest(_z_session_t *zn, uint32_t interest_id) {
         if (_z_send_n_msg(zn, &n_msg, Z_RELIABILITY_RELIABLE, Z_CONGESTION_CONTROL_BLOCK, NULL) != _Z_RES_OK) {
             _Z_ERROR_RETURN(_Z_ERR_TRANSPORT_TX_FAILED);
         }
+#if Z_FEATURE_AUTO_RECONNECT == 1
+        _z_prune_declaration(zn, &n_msg);
```
oteffahi, line 570: Seems like I missed this in my previous review:
Current implementation of `_z_prune_declaration` needs updating to support removing Interest messages from the declarations buffer.

gmartin82: It is actually handled in `_z_prune_declaration()` under `case _Z_N_INTEREST:`. I've just confirmed this with the `z_pub` example and some quickly inserted print statements:

```
Calling _z_prune_declaration()
_z_prune_declaration() calling _z_network_message_slist_drop_filter()
```

oteffahi: My bad, I was on `release/1.4.0` branch due to the issue being noticed first on there by the user 😅

oteffahi (approved):

sashacmc (approved):

PR #1046, opened 2025-09-22
Hardened closing of network sockets.

### LINKED ISSUE #1014: [Bug] Socket may be closed multiple times on connection failure
### Describe the bug

Open functions in system/xx/network.c do close the socket on failure before exit.
For example `_z_open_tcp()` executes the following code on exit:
 ```
if (ret != _Z_RES_OK) {
    close(sock->_fd);
 }
```
At higher level, `_z_close_tcp()` may also be called, for example from `_z_link_clear()`.

This may be a problem in some platform like Android (are there others?) where mechanism lice [fdsan](https://android.googlesource.com/platform/bionic/+/master/docs/fdsan.md) are used.
In Android API 30 and later, this may lead in crashes if fdsan isn't disabled.

### To reproduce

N/A

### System info

- Android 13
- Zenoh 1.5.0

gmartin82:
Closes #1014.


gmartin82:
> It seems something wrong with the Windows, could you please check?

@sashacmc it looks like it was just a flaky test.

PR #1056, opened 2025-10-07
chore: update dev debian version

diogomatsubara:
This way daily dev release will compare higher than the released version

Fix https://github.com/eclipse-zenoh/zenoh-pico/actions/runs/18298240280/job/52101067160


gmartin82 (approved):
Looks good.

PR #1066, opened 2025-10-20
Local messages optimisation

steils:
- Deliver PUT/DELETE, query, reply, and reply-final messages for a key expression that is defined in the same local session locally, without causing network activity.
- Extend tests to validate deliveries both to remote and local subscribers.
- Support locality of samples to be received by subscribers or targeted by publishers.
- Enable the locality-specific CMake options in CI.

User can now set allowed_origin / allowed_destination in z_put_options_t, z_publisher_put_options_t, z_subscriber_options_t, z_queryable_options_t, etc. with the following values:
- Z_LOCALITY_SESSION_LOCAL - stay inside the session (don't send
  anything to network)
- Z_LOCALITY_REMOTE - send/accept only over the transport (no
  same-session samples)
- Z_LOCALITY_ANY - allow both, exactly as with zenoh-c's zc_locality_t.

Co-developed-by: @DenisBiryukov91

### DIFF
--- a/src/session/subscription.c
+++ b/src/session/subscription.c
@@ -301,18 +301,9 @@ z_result_t _z_trigger_subscriptions_impl(_z_session_t *zn, _z_subscriber_kind_t 
 
 void _z_unregister_subscription(_z_session_t *zn, _z_subscriber_kind_t kind, _z_subscription_rc_t *sub) {
 #if Z_FEATURE_LOCAL_SUBSCRIBER == 1
-    _z_keyexpr_t key_copy = _z_keyexpr_null();
-    z_locality_t origin = Z_LOCALITY_ANY;
-    bool notify = false;
     if (kind == _Z_SUBSCRIBER_KIND_SUBSCRIBER) {
         _z_subscription_t *sub_val = _Z_RC_IN_VAL(sub);
-        if (_z_locality_allows_local(sub_val->_allowed_origin)) {
-            key_copy = _z_keyexpr_duplicate(&sub_val->_key);
-            if (_z_keyexpr_has_suffix(&key_copy)) {
-                origin = sub_val->_allowed_origin;
-                notify = true;
-            }
-        }
+        _z_write_filter_notify_subscriber(zn, &sub_val->_key, sub_val->_allowed_origin, false);
     }
 #endif
     _z_session_mutex_lock(zn);
@@ -325,13 +316,6 @@ void _z_unregister_subscription(_z_session_t *zn, _z_subscriber_kind_t kind, _z_
     }
 
     _z_session_mutex_unlock(zn);
-
-#if Z_FEATURE_LOCAL_SUBSCRIBER == 1
-    if (notify) {
-        _z_write_filter_notify_subscriber(zn, &key_copy, origin, false);
-        _z_keyexpr_clear(&key_copy);
-    }
-#endif
 }
 
 void _z_flush_subscriptions(_z_session_t *zn) {


src/session/loopback.c
```diff
@@ -0,0 +1,471 @@
...
+            if (_z_encoding_copy(&encoding_copy, encoding) != _Z_RES_OK) {
+                _z_bytes_drop(&payload_copy);
+                _z_bytes_drop(&attachment_copy);
+                return false;
+            }
+        }
+    }
+
+    switch (kind) {
+        case Z_SAMPLE_KIND_PUT:
```
sashacmc: If `*_copy` is only used for `Z_SAMPLE_KIND_PUT`, wouldn't it be better to put them under case? To avoid unnecessary cleanup and potential memory leak.

src/session/loopback.c
```diff
@@ -0,0 +1,471 @@
...
+            _z_slice_clear(&params_copy);
+            _z_bytes_drop(&payload_copy);
+            _z_bytes_drop(&attachment_copy);
+            return false;
+        }
+    }
+
+    _z_zenoh_message_t msg;
+    _z_n_msg_make_query(&msg, keyexpr, &params_copy, qid, Z_RELIABILITY_DEFAULT, consolidation,
+                        (payload == NULL) ? NULL : &payload_copy, (encoding == NULL) ? NULL : &encoding_copy, 0,
```
sashacmc: Any reason to have timeout hardcoded to `0` in place of original one?

tests/z_local_loopback_test.c
```diff
@@ -0,0 +1,245 @@
+//
```
sashacmc, line 1: The tests covers basic scenarios:
  - Single local subscriber
  - Single local queryable

It would be nice to have also:
  - Multiple local subscribers/queriables
  - Mixed local + remote subscribers/queriables

src/session/loopback.c
steils: By mistake. Didn't forward it from the original query. I'll fix it.

src/session/loopback.c
steils: I see. I'll reorganize the code in `_z_session_deliver_push_locally()`.

src/net/primitives.c
```diff
@@ -195,8 +196,15 @@ z_result_t _z_write(_z_session_t *zn, const _z_keyexpr_t *keyexpr, const _z_byte
                     z_priority_t priority, bool is_express, const _z_timestamp_t *timestamp,
                     const _z_bytes_t *attachment, z_reliability_t reliability, const _z_source_info_t *source_info) {
     z_result_t ret = _Z_RES_OK;
-    _z_network_message_t msg;
     _z_qos_t qos = _z_n_qos_make(is_express, cong_ctrl == Z_CONGESTION_CONTROL_BLOCK, priority);
+
+    bool handled_locally = _z_session_deliver_push_locally(zn, keyexpr, payload, encoding, kind, qos, timestamp,
+                                                           attachment, reliability, source_info);
+    if (handled_locally && !_z_session_has_remote_targets(zn)) {
```
gmartin82: I don't understand this condition. If there are no remote targets, then why does `handled_locally` matter when deciding whether to publish to the network or not?

src/net/primitives.c
```diff
@@ -378,6 +386,12 @@ z_result_t _z_send_reply(const _z_query_t *query, const _z_session_rc_t *zsrc, c
     }
     // Build the reply context decorator. This is NOT the final reply.
     _z_n_qos_t qos = _z_n_qos_make(is_express, cong_ctrl == Z_CONGESTION_CONTROL_BLOCK, priority);
+    bool handled_locally = _z_session_deliver_reply_locally(query, zsrc, keyexpr, payload, encoding, kind, qos,
+                                                            timestamp, att, source_info);
+    if (handled_locally && !_z_session_has_remote_targets(zn)) {
```
gmartin82: Same question as before.

src/net/primitives.c
```diff
@@ -408,6 +422,11 @@ z_result_t _z_send_reply_err(const _z_query_t *query, const _z_session_rc_t *zsr
     // Build the reply context decorator. This is NOT the final reply.
     _z_n_qos_t qos = _z_n_qos_make(false, true, Z_PRIORITY_DEFAULT);
     _z_source_info_t source_info = _z_source_info_null();
+    bool handled_locally = _z_session_deliver_reply_err_locally(query, zsrc, payload, encoding, qos);
+    if (handled_locally && !_z_session_has_remote_targets(zn)) {
```
gmartin82: Same question as before.

src/net/primitives.c
```diff
@@ -501,6 +520,13 @@ z_result_t _z_query(_z_session_t *zn, const _z_keyexpr_t *keyexpr, const char *p
     _z_n_msg_make_query(&z_msg, keyexpr, &params, pq->_id, Z_RELIABILITY_DEFAULT, pq->_consolidation, payload, encoding,
                         timeout_ms, attachment, qos, &source_info);
 
+    bool handled_locally = _z_session_deliver_query_locally(zn, keyexpr, &params, pq->_consolidation, payload, encoding,
+                                                            attachment, &source_info, pq->_id, timeout_ms, qos);
+    if (handled_locally && !_z_session_has_remote_targets(zn)) {
```
gmartin82: Same question as before.

src/net/query.c
```diff
@@ -31,10 +32,16 @@ z_result_t _z_query_send_reply_final(_z_query_t *q) {
     if (_Z_RC_IS_NULL(&sess_rc)) {
         _Z_ERROR_RETURN(_Z_ERR_TRANSPORT_TX_FAILED);
     }
+
+    _z_session_t *session = _Z_RC_IN_VAL(&sess_rc);
+    if (_z_session_deliver_reply_final_locally(session, q->_request_id)) {
```
gmartin82: I'm not sure if this code path checks that reply final was handled locally. It looks like it gets to a function that returns `_Z_RES_OK` regardless.

src/net/primitives.c
gmartin82: Actually, I think this is okay with it being a reply.

src/net/primitives.c
gmartin82: Actually, I think this is okay with it being a reply.

src/net/primitives.c
steils: Well, if `handled_locally == false` we fall back to the old behaviour, so it's unchanged: every `_z_write` went through `_z_send_n_msg`, even if nobody was listening (or we don't know they are listening, for example targets built without interest support). If `handled_locally == true` we skip the network path, otherwise there would be no sense in these changes. Is it okay?

src/net/query.c
steils: What do you mean? It checks if there is `pending_query` with this `request_id`, and then gets to `_z_handle_network_message()` with `msg->tag == _Z_N_RESPONSE_FINAL`. Isn't it what we want it to do?

src/net/primitives.c
gmartin82: I understand why the changes were made, we don't want to write to the network unnecessarily when responding to a local entity.

My point though is that we either need to send the data to the network or we don't. I think the check should simply be `if (!_z_session_has_remote_targets(zn)) {`.

src/net/primitives.c
gmartin82: So here we're replying to a query which involves a single querier so I think the check should simply be: `if (handled_locally) {`

src/net/primitives.c
gmartin82: As above I think this should just be: `if (handled_locally) {`

src/net/query.c
gmartin82: Yeah, sorry this is fine. I was  looking at `_z_trigger_reply_final()` and noticed it aleays returned `_Z_RES_OK` but I can see the check for a local pending query handles this case.

src/net/primitives.c
steils: `_z_session_has_remote_targets` only knows what it can get through interest/declarations. If for some reason we don't get this information, we won't send the message at all, and it will silently disappear. Do you think it's acceptable?
The old behaviour was always to call `_z_send_n_msg()`. In my version, when we didn't handle the message locally, we preserve the old behaviour.

gmartin82: The behaviour has been changed by this PR, though. Previously, the implementation attempted to deliver messages to both local and remote subscribers, which is the correct behaviour for pub/sub middleware since (except for queries) there can be multiple subscribers, and local and remote delivery are independent concerns.

This PR makes remote delivery partly dependent on whether messages have been delivered locally, which seems incorrect. The decision to send messages over the network should be independent of local delivery.

So either we can rely on `_z_session_has_remote_targets` to determine whether messages need to be sent based on interest/declarations, or we can’t. If we can’t, then messages should always be sent to the network. But if we can, then the optimisation should occur independently of local message delivery.

src/net/primitives.c
steils: z_api_alignment_test showed that we cannot rely on interest to decide whether to send the message through the transport. So I had to rework it to use "has peers" condition.

include/zenoh-pico/api/constants.h
```diff
@@ -54,6 +56,44 @@ typedef enum z_whatami_t {
...
 
+/**
+ * The locality of samples to be received by subscribers or targeted by publishers.
+ *
+ * Enumerators:
+ *   ZP_LOCALITY_ANY: Allow both session-local and remote traffic.
+ *   ZP_LOCALITY_SESSION_LOCAL: Allow session-local traffic only.
+ *   ZP_LOCALITY_REMOTE: Allow remote traffic only.
+ */
+typedef enum zp_locality_t {
```
DenisBiryukov91: I think it would make sense to use z_ instead of zp_ prefix for locality (and do the same for zenoh-c) - this will simplify C++ implementation.

include/zenoh-pico/api/constants.h
```diff
@@ -54,6 +56,44 @@ typedef enum z_whatami_t {
...
+    ZP_LOCALITY_ANY = 0,
+    ZP_LOCALITY_SESSION_LOCAL = 1,
+    ZP_LOCALITY_REMOTE = 2,
+} zp_locality_t;
+
+static inline zp_locality_t zp_locality_default(void) { return ZP_LOCALITY_ANY; }
+
+static inline bool zp_locality_allows_local(zp_locality_t locality) {
```
DenisBiryukov91: I would suggest to prefix this and `zp_locality_allows_remote` with underscore, to indicate that they are for internal use only and maybe also move them out of public header.

include/zenoh-pico/net/primitives.h
```diff
@@ -307,8 +311,8 @@ z_result_t _z_undeclare_querier(_z_querier_t *querier);
 z_result_t _z_query(_z_session_t *zn, const _z_keyexpr_t *keyexpr, const char *parameters, size_t parameters_len,
                     z_query_target_t target, z_consolidation_mode_t consolidation, const _z_bytes_t *payload,
                     const _z_encoding_t *encoding, _z_closure_reply_callback_t callback, _z_drop_handler_t dropper,
-                    void *arg, uint64_t timeout_ms, const _z_bytes_t *attachment, _z_n_qos_t qos, _z_zint_t *out_id);
-
+                    void *arg, uint64_t timeout_ms, const _z_bytes_t *attachment, _z_n_qos_t qos, _z_zint_t *out_id,
```
DenisBiryukov91: I would suggest to put out_id out parameter at the end of the list

src/session/loopback.c
```diff
@@ -0,0 +1,492 @@
...
+    _z_encoding_t encoding_copy = _z_encoding_null();
+
+    if (parameters != NULL) {
+        if (parameters->len > 0) {
+            if (_z_slice_copy(&params_copy, parameters) != _Z_RES_OK) {
+                return false;
+            }
+        } else if (parameters->start != NULL) {
+            // Preserve zero-length string so downstream handlers see the original buffer
+            params_copy = _z_slice_alias_buf(parameters->start, 0);
```
DenisBiryukov91: I am not sure that it is safe to alias "", since it still could be a non-const string.

src/session/loopback.c
```diff
@@ -0,0 +1,492 @@
...
+    }
+
+    _z_slice_t params_copy = _z_slice_null();
+    _z_bytes_t payload_copy = _z_bytes_null();
+    _z_bytes_t attachment_copy = _z_bytes_null();
+    _z_encoding_t encoding_copy = _z_encoding_null();
+
+    if (parameters != NULL) {
+        if (parameters->len > 0) {
+            if (_z_slice_copy(&params_copy, parameters) != _Z_RES_OK) {
```
DenisBiryukov91: instead of multiple early returns in case of error, maybe it will be more readeable to chain errors using _Z_SET_IF_OK macro, and just call all the clean up code once  before _z_n_msg_make_query.

src/session/loopback.c
```diff
@@ -0,0 +1,492 @@
...
+        return false;
+    }
+
+    _z_transport_common_t *transport = _z_session_get_transport_common(dst);
+    if (transport == NULL) {
+        _z_session_rc_drop(&dst_rc);
+        return false;
+    }
+
+    _z_bytes_t payload_copy = _z_bytes_null();
```
DenisBiryukov91: instead of multiple early returns in case of error, maybe it will be more readeable to chain errors using _Z_SET_IF_OK macro, and just call all the clean up code once  before _z_n_msg_make_reply_...

src/session/loopback.c
```diff
@@ -0,0 +1,492 @@
...
```
DenisBiryukov91: instead of multiple early returns in case of error, maybe it will be more readeable to chain errors using _Z_SET_IF_OK macro, and just call all the clean up code once  before _z_n_msg_make_reply_err.

src/session/subscription.c
```diff
@@ -212,15 +227,21 @@ static z_result_t _z_subscription_get_infos(_z_session_t *zn, _z_subscriber_kind
...
+                     __unsafe_z_get_subscriptions_rc_by_key(zn, kind, &infos->ke_out, infos->is_remote, &infos->infos));
 #if Z_FEATURE_RX_CACHE == 1
-        _z_subscription_cache_data_t cache_storage = _z_subscription_cache_data_null();
-        cache_storage.infos = _z_subscription_rc_svec_rc_clone(&infos->infos);
-        _Z_SET_IF_OK(ret, _z_keyexpr_copy(&cache_storage.ke_out, &infos->ke_out));
-        _Z_SET_IF_OK(ret, _z_keyexpr_copy(&cache_storage.ke_in, &infos->ke_in));
-        _Z_SET_IF_OK(ret, _z_subscription_lru_cache_insert(&zn->_subscription_cache, &cache_storage));
-        if (ret != _Z_RES_OK) {
-            _z_subscription_cache_data_clear(&cache_storage);
+        if (ret == _Z_RES_OK) {
```
DenisBiryukov91: I'd rather not attempt any early exits in case of error, since this is considered to be unlikely scenario and thus it is not worth to optimize for it in favor of readability

src/session/queryable.c
```diff
@@ -116,7 +119,14 @@ static z_result_t __unsafe_z_get_session_queryables_by_key(_z_session_t *zn, con
     while (xs != NULL) {
         // Parse queryable list
         _z_session_queryable_rc_t *qle = _z_session_queryable_rc_slist_value(xs);
-        if (_z_keyexpr_suffix_intersects(&_Z_RC_IN_VAL(qle)->_key, key)) {
+        const _z_session_queryable_t *qle_val = _Z_RC_IN_VAL(qle);
+        bool origin_allowed = is_remote ? zp_locality_allows_remote(qle_val->_allowed_origin)
+                                        : zp_locality_allows_local(qle_val->_allowed_origin);
+        if (!origin_allowed) {
```
DenisBiryukov91: I would suggest combine this check with the next one i.e `if (origin_allowed && _z_keyexpr_suffix_intersects(&qle_val->_key, key))`

src/session/subscription.c
steils: Like this?
```
            _Z_SET_IF_OK(ret, _z_keyexpr_copy(&cache_storage.ke_out, &infos->ke_out));
            _Z_SET_IF_OK(ret, _z_keyexpr_copy(&cache_storage.ke_in, &infos->ke_in));
            _Z_SET_IF_OK(ret, _z_subscription_lru_cache_insert(&zn->_subscription_cache, &cache_storage));
```

src/session/loopback.c
steils: Well, `_z_slice_alias_buf()` always stores the pointer as const uint8_t * and sets a null-deleter, and we treat params_copy as const during the whole call chain. But if you want to be extra safe, we can call _z_slice_copy() always, but I think that just adds allocations without changing behaviour.

src/session/subscription.c
DenisBiryukov91: yes, maybe also add a macro like `_Z_DO_IF_OK`, to put stuff like `cache_storage.infos = _z_subscription_rc_svec_rc_clone(&infos->infos);` and `cache_storage.is_remote = infos->is_remote;` under it for better clarity i.e. like `_Z_DO_IF_OK(ret, cache_storage.infos = _z_subscription_rc_svec_rc_clone(&infos->infos); ...)` to hide boilerplate if checks from the reader.

src/session/loopback.c
DenisBiryukov91: The issue is that these parameters go directly into query callback, and thus if user move them out (via fifo channel for example) and at the same time frees parameters after z_get he will end up with a dangling pointer.

DenisBiryukov91: I think that we should just leave it as _z_slice_null in the case of 0 len.

include/zenoh-pico/api/constants.h
```diff
@@ -14,6 +14,8 @@
 #ifndef ZENOH_PICO_API_CONSTANTS_H
 #define ZENOH_PICO_API_CONSTANTS_H
 
+#include <stdbool.h>
```
gmartin82: Is this actually used in this file? I don't see any boolean values.

steils: You are right. That's a leftover from the previous version of the patch, I moved the helpers which used boolean values to another header.

include/zenoh-pico/api/types.h
```diff
@@ -218,6 +218,7 @@ typedef struct {
 #ifdef Z_FEATURE_UNSTABLE_API
     z_reliability_t reliability;
 #endif
+    z_locality_t allowed_destination;
```
DenisBiryukov91, line 226: Should not this and similar fields  be under Z_FEATURE_LOCAL_SUBSCRIBER/Z_FEATURE_LOCAL_QUERYABLE - otherwise they do not seem to make any sense.

src/session/loopback.c
```diff
@@ -0,0 +1,484 @@
...
+    }
+
+    _z_slice_t params_copy = _z_slice_null();
+    _z_bytes_t payload_copy = _z_bytes_null();
+    _z_bytes_t attachment_copy = _z_bytes_null();
+    _z_encoding_t encoding_copy = _z_encoding_null();
+    z_result_t ret = _Z_RES_OK;
+
+    if (parameters != NULL) {
+        if (parameters->len > 0) {
```
DenisBiryukov91: no need for the inner if/else branch just `_Z_SET_IF_OK(ret, _z_slice_copy(&params_copy, parameters));` should be ok independently of the length

src/session/loopback.c
```diff
@@ -0,0 +1,484 @@
...
+    bool found = false;
+
+    _z_session_mutex_lock(zn);
+    _z_keyexpr_t expanded = __unsafe_z_get_expanded_key_from_key(zn, keyexpr, true, NULL);
+    if (_z_keyexpr_has_suffix(&expanded)) {
+        _z_session_queryable_rc_slist_t *xs = zn->_local_queryable;
+        while (xs != NULL) {
+            _z_session_queryable_rc_t *qle = _z_session_queryable_rc_slist_value(xs);
+            const _z_session_queryable_t *qle_val = _Z_RC_IN_VAL(qle);
+            if (!_z_locality_allows_local(qle_val->_allowed_origin)) {
```
DenisBiryukov91: I would suggest to combine 2 ifs into one to simplify the code

src/session/loopback.c
```diff
@@ -0,0 +1,484 @@
...
+    bool found = false;
+
+    _z_session_mutex_lock(zn);
+    _z_keyexpr_t expanded = __unsafe_z_get_expanded_key_from_key(zn, keyexpr, true, NULL);
+    if (_z_keyexpr_has_suffix(&expanded)) {
+        _z_subscription_rc_slist_t *xs = zn->_subscriptions;
+        while (xs != NULL) {
+            _z_subscription_rc_t *sub = _z_subscription_rc_slist_value(xs);
+            const _z_subscription_t *sub_val = _Z_RC_IN_VAL(sub);
+            if (!_z_locality_allows_local(sub_val->_allowed_origin)) {
```
DenisBiryukov91: I would suggest to combine 2 ifs into 1 to simplify the code

src/session/loopback.c
```diff
@@ -0,0 +1,484 @@
...
+            _z_encoding_t encoding_copy = _z_encoding_null();
+
+            if (payload != NULL) {
+                if (_z_bytes_copy(&payload_copy, payload) != _Z_RES_OK) {
+                    return false;
+                }
+            }
+            if (attachment != NULL) {
+                if (_z_bytes_copy(&attachment_copy, attachment) != _Z_RES_OK) {
+                    _z_bytes_drop(&payload_copy);
```
DenisBiryukov91: Similarly to other places I would suggest to chain _Z_SET_IF_OK and call clean up functions only once before `_z_n_msg_make_put`

src/session/loopback.c
```diff
@@ -0,0 +1,484 @@
...
+                Z_CONSOLIDATION_MODE_DEFAULT, qos, timestamp, source_info, (payload == NULL) ? NULL : &payload_copy,
+                (encoding == NULL) ? NULL : &encoding_copy, (attachment == NULL) ? NULL : &attachment_copy);
+            break;
+        case Z_SAMPLE_KIND_DELETE:
+            _z_n_msg_make_reply_ok_del(&msg, &src->_local_zid, query->_request_id, keyexpr, Z_RELIABILITY_DEFAULT,
+                                       Z_CONSOLIDATION_MODE_DEFAULT, qos, timestamp, source_info,
+                                       (attachment == NULL) ? NULL : &attachment_copy);
+            break;
+        default:
+            _z_bytes_drop(&payload_copy);
```
DenisBiryukov91: maybe it makes sense to just return early if kind is not a valid Z_SAMPLE_KIND (although why should it be ?) to avoid copy pasting same clean up code ?

src/session/subscription.c
```diff
@@ -113,7 +118,14 @@ static z_result_t __unsafe_z_get_subscriptions_by_key(_z_session_t *zn, _z_subsc
     while (xs != NULL) {
         // Parse subscription list
         _z_subscription_rc_t *sub = _z_subscription_rc_slist_value(xs);
-        if (_z_keyexpr_suffix_intersects(&_Z_RC_IN_VAL(sub)->_key, key)) {
+        const _z_subscription_t *sub_val = _Z_RC_IN_VAL(sub);
+        bool origin_allowed = is_remote ? _z_locality_allows_remote(sub_val->_allowed_origin)
+                                        : _z_locality_allows_local(sub_val->_allowed_origin);
+        if (!origin_allowed) {
```
DenisBiryukov91: I would suggest to combine 2 ifs into one

src/session/subscription.c
```diff
@@ -212,15 +228,19 @@ static z_result_t _z_subscription_get_infos(_z_session_t *zn, _z_subscriber_kind
...
+                     __unsafe_z_get_subscriptions_rc_by_key(zn, kind, &infos->ke_out, infos->is_remote, &infos->infos));
 #if Z_FEATURE_RX_CACHE == 1
-        _z_subscription_cache_data_t cache_storage = _z_subscription_cache_data_null();
-        cache_storage.infos = _z_subscription_rc_svec_rc_clone(&infos->infos);
-        _Z_SET_IF_OK(ret, _z_keyexpr_copy(&cache_storage.ke_out, &infos->ke_out));
-        _Z_SET_IF_OK(ret, _z_keyexpr_copy(&cache_storage.ke_in, &infos->ke_in));
-        _Z_SET_IF_OK(ret, _z_subscription_lru_cache_insert(&zn->_subscription_cache, &cache_storage));
-        if (ret != _Z_RES_OK) {
-            _z_subscription_cache_data_clear(&cache_storage);
+        if (ret == _Z_RES_OK) {
```
DenisBiryukov91: this if is not necessary

src/session/queryable.c
```diff
@@ -182,16 +193,22 @@ static z_result_t _z_session_queryable_get_infos(_z_session_t *zn, _z_queryable_
...
 #if Z_FEATURE_RX_CACHE == 1
-        // Update cache
-        _z_queryable_cache_data_t cache_storage = _z_queryable_cache_data_null();
-        cache_storage.infos = _z_session_queryable_rc_svec_rc_clone(&infos->infos);
-        _Z_SET_IF_OK(ret, _z_keyexpr_copy(&cache_storage.ke_in, &infos->ke_in));
-        _Z_SET_IF_OK(ret, _z_keyexpr_copy(&cache_storage.ke_out, &infos->ke_out));
-        _Z_SET_IF_OK(ret, _z_queryable_lru_cache_insert(&zn->_subscription_cache, &cache_storage));
-        if (ret != _Z_RES_OK) {
-            _z_queryable_cache_data_clear(&cache_storage);
+        if (ret == _Z_RES_OK) {
```
DenisBiryukov91: this if is not necessary

include/zenoh-pico/api/types.h
steils: Well, with `Z_FEATURE_LOCAL_* == 0` we just ignore them safely. The default value is `Z_LOCALITY_ANY`. The main benefit of keeping those fields without ifdefs is avoiding ABI splits between different build options, which may be useful for user apps. It would also add complexity to the initialisations (more ifdefs there).
Do you think it's worth it?

DenisBiryukov91: the issue is that if user makes a put with locality= Local he would expect the messages to be delivered only to local subscribers, while it would not be the case with Z_FEATURE_LOCAL_SUBSCRIBER=0

src/api/api.c
```diff
@@ -1085,9 +1099,9 @@ z_result_t z_declare_publisher(const z_loaned_session_t *zs, z_owned_publisher_t
...
 
     // Set publisher
-    _z_publisher_t int_pub =
-        _z_declare_publisher(zs, final_key, opt.encoding == NULL ? NULL : &opt.encoding->_this._val,
-                             opt.congestion_control, opt.priority, opt.is_express, reliability);
+    _z_publisher_t int_pub = _z_declare_publisher(
+        zs, final_key, opt.encoding == NULL ? NULL : &opt.encoding->_this._val, opt.congestion_control, opt.priority,
+        opt.is_express, reliability, opt.allowed_destination);
     // Create write filter
     z_result_t res = _z_write_filter_create(zs, &int_pub._filter, keyexpr_aliased, _Z_INTEREST_FLAG_SUBSCRIBERS, false);
```
DenisBiryukov91: Write filter should also take into account locality (i.e. it should be active if locality of publisher does not match any of subscribers locality), same for queryables. 
This is required because write_filter is used by matching listener/status and thus if not accounted for correctly - the values they return would be wrong. I.e. matching status for local only publisher should not return true if there are only remote subscribers with matching keyexpr.

include/zenoh-pico/api/types.h
steils: Aaah, sure! I totally missed the elephant in the room :-D
I'll fix it, thanks.

src/session/loopback.c
steils: It makes `_z_slice_init(dst, src->len)` inside `_z_slice_copy`, and then `assert(capacity != 0)` (where capacity is `src->len`), so I don't think it would be ok.

src/session/loopback.c
steils: I don't quite understand how to write it with _Z_SET_IF_OK, so that we wouldn't call _z_n_msg_make_push_put() in case of failure. I would leave it like this in this particular place, it looks clean enough for me.

src/session/queryable.c
steils: Fixed. Also there was a typo (?), the wrong cache was updated. I made a separate PR for it: https://github.com/eclipse-zenoh/zenoh-pico/pull/1090

src/session/loopback.c
DenisBiryukov91: same as in other places put it under `if (ret == _Z_RES_OK) { ... }` or introduce a macro do `_Z_DO_IF_OK` as I suggested earlier

src/session/loopback.c
DenisBiryukov91: then we would need to fix _z_slice_copy to account for zero length slices instead of exposing such details everywhere, because this definitely has nothing to do with `_z_session_deliver_query_locally`

steils: @DenisBiryukov91 done, made a separate PR for this: https://github.com/eclipse-zenoh/zenoh-pico/pull/1095

src/api/api.c
steils: That was a big one, but done.

src/session/loopback.c
steils: We cannot call clean up functions in one place here. I afraid I don't understand how to rewrite this particular place, and I don't think we should.

src/net/filtering.c
```diff
@@ -205,14 +374,94 @@ z_result_t _z_write_filter_ctx_add_callback(_z_write_filter_ctx_t *ctx, size_t i
...
+        _z_write_filter_registry_unlock();
+        return;
+    }
+
+    // Count matching write filters so we can allocate a buffer once.
+    size_t match_count = 0;
+    for (_z_write_filter_registration_t *registration = entry->registrations; registration != NULL;
+         registration = registration->next) {
+        _z_write_filter_ctx_t *registration_ctx = _Z_RC_IN_VAL(&registration->ctx_rc);
+        if (registration_ctx->allow_local && _z_keyexpr_suffix_intersects(&registration_ctx->key, key)) {
```
DenisBiryukov91: We should account here for completness of queryables: i.e. 
```suggestion
        if (registration_ctx->allow_local && (registration_ctx->is_complete ? (is_complete && _z_keyexpr_suffix_includes(key, &registration_ctx->key)) : _z_keyexpr_suffix_intersects(&registration_ctx->key, key) )) {
```

src/session/queryable.c
```diff
@@ -163,16 +171,27 @@ _z_session_queryable_rc_t *_z_register_session_queryable(_z_session_t *zn, _z_se
     *ret = _z_session_queryable_rc_new_from_val(q);
     _z_session_mutex_unlock(zn);
 
+#if Z_FEATURE_LOCAL_QUERYABLE == 1
+    if (ret != NULL && _z_locality_allows_local(q->_allowed_origin)) {
+        _z_session_queryable_t *qle_val = _Z_RC_IN_VAL(ret);
+        _z_write_filter_notify_local_entity(zn, &qle_val->_key, qle_val->_allowed_origin, true);
```
DenisBiryukov91: we need to pass `complete` flag of queryable here

src/net/filtering.c
```diff
@@ -205,14 +374,94 @@ z_result_t _z_write_filter_ctx_add_callback(_z_write_filter_ctx_t *ctx, size_t i
 }
 #endif
 
+void _z_write_filter_notify_local_entity(_z_session_t *session, const _z_keyexpr_t *key, z_locality_t allowed_origin,
+                                         bool add) {
```
DenisBiryukov91: We need to pass completeness of new entity here (or do the check prior to the call):
```suggestion
                                         bool is_complete, bool add) {
```

src/net/filtering.c
```diff
@@ -205,14 +374,94 @@ z_result_t _z_write_filter_ctx_add_callback(_z_write_filter_ctx_t *ctx, size_t i
...
+    }
+
+    for (_z_write_filter_registration_t *registration = entry->registrations; registration != NULL;
+         registration = registration->next) {
+        _z_write_filter_ctx_t *registration_ctx = _Z_RC_IN_VAL(&registration->ctx_rc);
+        if (!registration_ctx->allow_local || !_z_keyexpr_suffix_intersects(&registration_ctx->key, key)) {
+            continue;
+        }
+        matches[match_len] = _z_write_filter_ctx_rc_clone(&registration->ctx_rc);
+        if (_Z_RC_IS_NULL(&matches[match_len])) {
```
DenisBiryukov91: I do not think this check is necessary here, since rc_clone normally should not fail (except for counter overflow) - which is highly unlikely in this context

src/net/filtering.c
```diff
@@ -205,14 +374,94 @@ z_result_t _z_write_filter_ctx_add_callback(_z_write_filter_ctx_t *ctx, size_t i
...
+    matches = (_z_write_filter_ctx_rc_t *)z_malloc(sizeof(_z_write_filter_ctx_rc_t) * match_count);
+    if (matches == NULL) {
+        _z_write_filter_registry_unlock();
+        return;
+    }
+
+    for (_z_write_filter_registration_t *registration = entry->registrations; registration != NULL;
+         registration = registration->next) {
+        _z_write_filter_ctx_t *registration_ctx = _Z_RC_IN_VAL(&registration->ctx_rc);
+        if (!registration_ctx->allow_local || !_z_keyexpr_suffix_intersects(&registration_ctx->key, key)) {
```
DenisBiryukov91: Should account for completeness here similarly to what suggested above. 
Overall I do not think this is a good idea to to spilt count of matching filters and their gathering, while you probably save a couple of allocations - you will have to call `_z_keyexpr_suffix_intersects` / `_z_keyexpr_suffix_includes` twice which are quite expensive.

src/session/loopback.c
```diff
@@ -0,0 +1,471 @@
...
+                }
+            }
+            if (encoding != NULL) {
+                if (_z_encoding_copy(&encoding_copy, encoding) != _Z_RES_OK) {
+                    _z_bytes_drop(&payload_copy);
+                    _z_bytes_drop(&attachment_copy);
+                    return false;
+                }
+            }
+
```
DenisBiryukov91: Why can't we call all clean up functions just here and return `false` if one of the copy operations above fails ?

steils: Ah. I was being dumb. Done.

src/net/filtering.c
steils: Done, made it with z_list_t instead of allocations, and in one loop.

src/net/filtering.c
```diff
@@ -205,14 +295,81 @@ z_result_t _z_write_filter_ctx_add_callback(_z_write_filter_ctx_t *ctx, size_t i
...
+        return;
+    }
+
+    _z_session_mutex_lock(session);
+
+    _z_list_t *matches = NULL;
+    for (_z_write_filter_registration_t *registration = session->_write_filters; registration != NULL;
+         registration = registration->next) {
+        _z_write_filter_ctx_t *registration_ctx = _Z_RC_IN_VAL(&registration->ctx_rc);
+        if (!registration_ctx->allow_local ||
```
DenisBiryukov91: maybe just `!(registration_ctx->allow_local && (registration_ctx->is_complete ? (is_complete && _z_keyexpr_suffix_includes(key, &registration_ctx->key)) : _z_keyexpr_suffix_intersects(&registration_ctx->key, key)))` to reduce number of '!' `?

src/net/filtering.c
```diff
@@ -24,9 +24,16 @@
 #include "zenoh-pico/session/resource.h"
 #include "zenoh-pico/session/session.h"
 #include "zenoh-pico/session/utils.h"
+#include "zenoh-pico/utils/locality.h"
 
 #if Z_FEATURE_INTEREST == 1
 
+typedef struct _z_write_filter_registration_t {
+    _z_write_filter_ctx_rc_t ctx_rc;
+    _z_session_t *session;
```
DenisBiryukov91: you do not need session field here, since _z_write_filter_ctx_rc_t already contains a weak pointer to its session

src/net/query.c
```diff
@@ -31,10 +34,30 @@ z_result_t _z_query_send_reply_final(_z_query_t *q) {
...
+        allow_remote = _z_locality_allows_remote(pending->_allowed_destination);
+    }
+    _Z_DEBUG("send_reply_final: rid=%jd allow_remote=%d", (intmax_t)q->_request_id, allow_remote);
+
+    if (_z_session_deliver_reply_final_locally(session, q->_request_id, allow_remote)) {
+        _z_session_rc_drop(&sess_rc);
+        return _Z_RES_OK;
+    }
+#else
+    if (_z_session_deliver_reply_final_locally(session, q->_request_id, true)) {
```
DenisBiryukov91: this copy pasted code can be avoided if you set `allow_remote = true;` for `Z_FEATURE_QUERY != 1`

src/net/primitives.c
```diff
@@ -535,6 +574,43 @@ z_result_t _z_query(_z_session_t *zn, const _z_keyexpr_t *keyexpr, const char *p
...
+    _ZP_UNUSED(encoding);
+    _ZP_UNUSED(attachment);
+    _ZP_UNUSED(source_info);
+#endif
+
+    _z_transport_common_t *common = _z_transport_get_common(&zn->_tp);
+    bool remote_possible = allow_remote && (common != NULL && common->_link != NULL);
+    _Z_DEBUG("_z_query: handled_locally=%d remote_possible=%d", handled_locally, remote_possible);
+
+    if (handled_locally && remote_possible) {
```
DenisBiryukov91: This does not look right for Z_CONSOLIDATION_MODE_LATEST, since it will potentially trigger reply 2 times instead of once (once for local queryable and again for remote one).

PR #1068, opened 2025-10-21
ci: increase timeout for integration tests

steils:
After adding Mbed TLS integration test that takes  quite a long time, CI started failing too often due to timeouts.


gmartin82 (approved):
LGTM

PR #1069, opened 2025-10-21
Fix matching

### LINKED ISSUE #1054: [Bug] Querier matching status listener do not account for querier 'complete' field
### Describe the bug

Matching status/listener logic uses interests which do not propagate queryable completeness - which means that querier's 'complete' field is ignored. This is different from zenoh-rust behavior where querier's 'complete' field is taken into account by matching status/listener.

### To reproduce

declare incomplete Queryable
declare complete Querier
Querier's matching status matching field will be true.

### System info

Any

DenisBiryukov91:
Fix matching logic to account for completeness of queryable
Move matching listener callbacks into _z_write_filter_callback
Check only exact ke equality in case of aggregated interests instead of intersection.
Resolves #1054 
Depends on https://github.com/eclipse-zenoh/zenoh/pull/2210

### DIFF
--- a/tests/api.sh
+++ b/tests/api.sh
@@ -29,9 +29,11 @@ echo "------------------ Running test $TESTBIN -------------------"
 sleep 5
 
 if [ ! -f zenohd ]; then
-    git clone https://github.com/DenisBiryukov91/zenoh.git zenoh-git
+    git clone https://github.com/eclipse-zenoh/zenoh.git zenoh-git
     cd zenoh-git || exit
-    git switch fix-aggregate
+    if [ -n "$ZENOH_BRANCH" ]; then
+        git switch "$ZENOH_BRANCH"
+    fi
     rustup show
     cargo build --lib --bin zenohd
     cp ./target/debug/zenohd "$TESTDIR"/
--- a/tests/routed.sh
+++ b/tests/routed.sh
@@ -29,9 +29,11 @@ echo "------------------ Running test $TESTBIN -------------------"
 sleep 5
 
 if [ ! -f zenohd ]; then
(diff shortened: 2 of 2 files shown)


tests/api.sh
```diff
@@ -29,11 +29,9 @@ echo "------------------ Running test $TESTBIN -------------------"
 sleep 5
 
 if [ ! -f zenohd ]; then
-    git clone https://github.com/eclipse-zenoh/zenoh.git zenoh-git
+    git clone https://github.com/DenisBiryukov91/zenoh.git zenoh-git
```
gmartin82: I guess this will be reverted to eclipse-zenoh before being committed?

tests/routed.sh
```diff
@@ -29,11 +29,9 @@ echo "------------------ Running test $TESTBIN -------------------"
 sleep 5
 
 if [ ! -f zenohd ]; then
-    git clone https://github.com/eclipse-zenoh/zenoh.git zenoh-git
+    git clone https://github.com/DenisBiryukov91/zenoh.git zenoh-git
```
gmartin82: I guess this will be reverted to eclipse-zenoh before being committed?

gmartin82 (approved):
Looks good. I'm assuming the git references will be reverted once the corresponding Zenoh PR is merged.

tests/api.sh
DenisBiryukov91: Yes, I temporary modified the branch to ensure that ci will pass

tests/routed.sh
DenisBiryukov91: Yes, I temporary modified the branch to ensure that ci will pass

PR #1073, opened 2025-10-27
Fix queryable and subscriber data races

DenisBiryukov91:
Fix data races when processing callbacks of subscriber or queryable
Fix data races in z_subscriber_keyexpr and z_queryable_keyexpr


DenisBiryukov91:
Maybe it is better to just keep session lock while we execute subscriber/queryable callabacks (which we already do for get). This should significantly simplify the code at the cost of slightly worth user experience (and inability to call zenoh entity functions from the same session inside callbacks - but this is already something we strongly discourage). @gmartin82 @steils do you have any opinions on this ?

gmartin82 (approved):
Looks good.

I think the CI job failure can be concessioned and handled separately, because it happens on main (and other PR's) and is related to disk space.

PR #1079, opened 2025-10-31
Fix _z_condvar_wait_until and z_task_join/detach

### LINKED ISSUE #1050: [Bug] double free or corruption (out) on z_task_detach
### Describe the bug

z_task_detach failed, because of incorrect task cleanup in _z_task_free

### To reproduce

Run in debug mode
```
        z_owned_task_t thread;

        z_task_init(&thread, NULL, log_read_thread, NULL);
        z_task_detach(z_move(thread));
```

### System info

Linux

DenisBiryukov91:
Fix _z_condvar_wait_until for macOS
Fix z_task_join/detach to not free the task pointer

Fixes #1050


gmartin82 (approved):
Looks good. Test failures seem unreleated so just need to be rerun before merging.

src/system/common/platform.c
```diff
@@ -28,14 +28,12 @@ z_result_t z_task_init(z_owned_task_t *task, z_task_attr_t *attr, void *(*fun)(v
 z_result_t z_task_join(z_moved_task_t *task) {
     _z_task_t *ptr = &task->_this._val;
     z_result_t ret = _z_task_join(ptr);
-    _z_task_free(&ptr);
```
steils, line 31: Why? And how is that needed to add the unit test?
Won't it add memory leak at least on some platforms?

src/system/common/platform.c
```diff
@@ -28,14 +28,12 @@ z_result_t z_task_init(z_owned_task_t *task, z_task_attr_t *attr, void *(*fun)(v
...
     return ret;
 }
 
 z_result_t z_task_detach(z_moved_task_t *task) {
     _z_task_t *ptr = &task->_this._val;
     z_result_t ret = _z_task_detach(ptr);
-    _z_task_free(&ptr);
```
steils, line 38: Same question.

src/system/common/platform.c
DenisBiryukov91: It is a bug, we should never assume that owned objects were allocated on heap (and even if they do, it is not our responsibility to free them). The unit test crashes without it. Otherwise this function is not used anywhere inside zenoh-pico.

src/system/common/platform.c
DenisBiryukov91: Same response.

PR #1082, opened 2025-11-05
Fix disk usage issues with CI jobs.

gmartin82:


diogomatsubara:
@gmartin82 For more disk space, please use this: https://github.com/eclipse-zenoh/zenoh/blob/main/.github/workflows/ci.yml#L362C4-L378C70, you might still need to relocate docker but the free disk space action usually cleans up a lot of space.

gmartin82:
@diogomatsubara I've switched to using that action to free up disk space and it seems moving Docker root to /mnt/docker is no longer required. Thanks!

diogomatsubara (approved):

PR #949, opened 2025-04-25
Zephyr build fixes

teburd:
A work in progress PR that fixes some of the build issues I ran into when building using west with zephyr.

I want to find out just how small I can make Zenoh, but first I need to build it with Zephyr

PR #1083, opened 2025-11-05
enhancement: zephyr: build fixes/improvements (re-spin of 949)

### LINKED ISSUE #1070: Make BATCH_UNICAST_SIZE configurable by KConfig for Zephyr RTOS
### Describe the feature

Currently BATCH_UNICAST_SIZE cannot be configured for Zephyr RTOS as the config.in file is never called by the zephyr/CMakeLists.txt

It would be nice to be able to configure this by the KConfig provided by Zephyr

Emil-Juhl:
An updated version of https://github.com/eclipse-zenoh/zenoh-pico/pull/949 from @teburd (hence the authorship in the commit).

Basically, this hooks the Zephyr compilation steps into the same templated-config-file flow as (I believe) the rest of the integrations use. Without this, using the standard `west` tooling to compile a zephyr project with zenoh-pico as a module would lead to compiler warnings about redefined macros for the various feature flags such as the one below:
```
[snip...]/include/zenoh-pico/config.h:32: warning: "Z_FEATURE_PUBLICATION" redefined
   32 | #define Z_FEATURE_PUBLICATION CONFIG_ZENOH_PICO_PUBLICATION
```

~The approach here is to add individual CMake variables for all of the zenoh-pico feature flags. Some of them will then reflect their corresponding Kconfig option, while others will just take on a default value and can be overridden by passing flags to cmake when building.~
The approach here is to move the feature configuration logic from the base CMakeLists into a helper file and then include that from both the base CMakeLists and the one inside the zephyr folder. The zephyr one will set some of the feature flags prior to inclusion so that the values match their kconfig counterpart.

In addition, this adds a couple of new Kconfig options that map to some zenoh-pico feature flags.

Closes: https://github.com/eclipse-zenoh/zenoh-pico/issues/1070

---
## 🏷️ Label-Based Checklist

**Labels:** `enhancement`

## ✨ Enhancement Requirements

**Instructions:**


zephyr/CMakeLists.txt
```diff
@@ -1,32 +1,83 @@
...
+set(Z_FEATURE_RX_CACHE 0 CACHE STRING "Toggle RX_CACHE")
+set(Z_FEATURE_UNICAST_PEER CONFIG_ZENOH_PICO_UNICAST_PEER_MODE CACHE STRING "Toggle Unicast peer mode")
+set(Z_FEATURE_AUTO_RECONNECT CONFIG_ZENOH_PICO_AUTO_RECONNECT CACHE STRING "Toggle automatic reconnection")
+set(Z_FEATURE_LINK_TLS 0 CACHE STRING "Toggle TLS link")
+set(Z_FEATURE_BATCH_TX_MUTEX 0 CACHE STRING "Toggle batch tx mutex")
+set(Z_FEATURE_BATCH_PEER_MUTEX 0 CACHE STRING "Toggle batch peer mutex")
+set(Z_FEATURE_PERIODIC_TASKS 0 CACHE STRING "Toggle periodic task support")
+set(Z_FEATURE_ADVANCED_PUBLICATION 0 CACHE STRING "Toggle advanced publication")
+set(Z_FEATURE_ADVANCED_SUBSCRIPTION 0 CACHE STRING "Toggle advanced subscription")
+set(Z_FEATURE_MULTICAST_DECLARATIONS 0 CACHE STRING "Toggle multicast declarations")
```
anedergaard: To enhance modularity and avoid redundancy, I propose that these feature configurations be encapsulated within a separate file, such as `feature-config.cmake`, located within the root directory. This file can subsequently be included from both the zephyr module and the primary CMakeLists.txt file within the same directory.
By implementing this approach, we eliminate the need to duplicate the list of feature configurations, thereby streamlining the development process and ensuring consistency across different modules.

Emil-Juhl: Very valid point. I didn't actually snoop around that much within the CMake chain for building without using the west module infrastructure.
I can see that this is basically done in the base cmake and can just be factored out into an include file as you suggest.

I've given it a shot.

anedergaard (approved):
Thanks @Emil-Juhl for implementing my suggestion. I validated it as a Zephyr module and also used the CMakefile in the root directory.

Emil-Juhl:
I rebased, and added one additional fix for a compiler warning when compiling with MULTI_THREAD and UNICAST_TRANSPORT enabled while UNICAST_PEER is left disabled. (see commit message below or in the commit itself).

Sorry, I don't normally try to put additional things on top of approved PRs, but since no one from the team with write access has looked yet, I figured I could do it this time.

```
bug: transport: unicast: unused function warning

When compiling zenoh-pico with Z_FEATURE_MULTI_THREAD and
Z_FEATURE_UNICAST_TRANSPORT enabled, while leaving
Z_FEATURE_UNICAST_PEER disabled, the `_zp_unicast_process_peer_event`
function will be defined but unused.

Only compile the function, alongside its own additional helpers, if
Z_FEATURE_UNICAST_PEER is enabled.
```

sashacmc (approved):

Emil-Juhl:
> Hello @Emil-Juhl, thanks for the contribution, it is looks good for me, but could you please resolve conflicts and make sure to all CI passed?

Hey! Sorry for the late reply I have been on vacation for a while. I will rebase within this week. Thanks for the feedback!

cr1sschwager:
Hey guys, is there a plan to integrate this in the near future as well as the fix to https://github.com/eclipse-zenoh/zenoh-pico/issues/1232?

Emil-Juhl:
> Hey guys, is there a plan to integrate this in the near future as well as the fix to #1232?

Sorry I basically forgot about this since last rebase as I actually don't use zenoh-pico myself anymore.

I rebased and added the fix for https://github.com/eclipse-zenoh/zenoh-pico/issues/1232 on top.

I tested by compiling this branch for x86 (just `make` from project root) and as a zephyr module (revived my old branch in the project I used zenoh-pico for). Both compiled OK.

I have not actually deployed the artifacts and tested like that. And the zephyr module integration is still not great as some of the zenoh-pico cmake still more or less calls for build-time modifications within the source tree (here I am looking at zenoh-pico.h with the versioning).

Emil-Juhl:
Hi @sashacmc I rebased this PR again and as far as I can tell CI checks pass. As mentioned I have compiled locally as standalone (just using `make` from project root) and as a west module for Zephyr.

Let me know if anything needs to change. Thanks!

cr1sschwager:
@DenisBiryukov91, @anedergaard it seems that @sashacmc is not reacting to this PR. Could you maybe takeover?

zephyr/CMakeLists.txt
```diff
@@ -1,39 +1,66 @@
...
+configure_file(
+  ${CMAKE_CURRENT_SOURCE_DIR}/../include/zenoh-pico.h.in
+  ${CMAKE_CURRENT_SOURCE_DIR}/../include/zenoh-pico.h
+  @ONLY
+)
+
+function(assign_kconfig feature text)
+  string(REPLACE Z_FEATURE CONFIG_ZENOH_PICO kconfig ${feature})
+  if(${kconfig})
+    set(${feature} ${kconfig} CACHE STRING ${text})
```
steils: This stores the Kconfig symbol name into the feature variable, not a value. For example, this generates `Z_FEATURE_PUBLICATION=CONFIG_ZENOH_PICO_PUBLICATION`, not `Z_FEATURE_PUBLICATION=1`.
Is it your intention?

zephyr/Kconfig.zenoh
```diff
@@ -5,6 +5,14 @@ config ZENOH_PICO
 
 if ZENOH_PICO
 
+config ZENOH_PICO_INTEREST
```
steils, line 8: All there `ZENO_PICO_*` features don't have defaults. So `CONFIG_ZENOH_PICO=y` alone disables most features.
That may be intentional, but then the sample `prj.conf` files and docs should explicitly enable the features they need. Or we need to set default here matching the CMake defaults from `feature_config.cmake`.

zephyr/Kconfig.zenoh
```diff
@@ -65,5 +88,14 @@ config ZENOH_PICO_LINK_WS
 	help
 	  WS Link
 
-endif
+config ZENOH_PICO_UNICAST_PEER_MODE
```
steils: Looks like a dead option or a typo. `zephyr/CMakeLists.txt:50` expects ZENOH_PICO_UNICAST_PEER, not ZENOH_PICO_UNICAST_PEER_MODE.

zephyr/CMakeLists.txt
```diff
@@ -1,39 +1,66 @@
...
+# Read version file and split it on '.' (dot) to extract components
+file(READ ${CMAKE_CURRENT_SOURCE_DIR}/../version.txt version)
+string(STRIP "${version}" version)
+string(REPLACE "." ";" version_list "${version}")
+
+# Populate the zenoh-pico version variables
+set(ZENOH_PICO ${version})
+list(POP_FRONT version_list ZENOH_PICO_MAJOR ZENOH_PICO_MINOR ZENOH_PICO_PATCH ZENOH_PICO_TWEAK)
+
+configure_file(
```
steils: This generates `zenoh-pico.h` into the tracked source tree. We stopped doing this in 3b3ab65cadbb10a8d7f32ba04cb15c26b8435dd5

steils (changes requested):
Thanks, the direction of this PR is good.  I would request changes before merge.

zephyr/CMakeLists.txt
Emil-Juhl: This was intentional, but I think I can change it to grab the value instead. The original reason was that I didn't have the conditional block and as such needed the preprocessor to expand the kconfig symbol in order to get the correct values at the end of the day.

I later realized that some checks in the code breaks if we substitute the CONFIG_... of a kconfig which is not set in the zephyr application. I also see that this was what already happened in the cmake logic, so I can just restore that cleanly for a bit of a nicer diff.

zephyr/Kconfig.zenoh
Emil-Juhl: Right. Would you prefer one solution over the other? I'd lean towards keeping most things disabled by default and updating the samples.

zephyr/Kconfig.zenoh
Emil-Juhl: Most likely a leftover from my rebase since this PR was created a while ago. I'll double check it when I update the code again later this week.

zephyr/CMakeLists.txt
Emil-Juhl: I did notice some files no longer get generated into the source tree, but as far as I can tell this one still does: https://github.com/eclipse-zenoh/zenoh-pico/blob/0ffe44a0502fd2cebeb974147239d4bca85b2ca7/CMakeLists.txt#L34-L39

That's actually why I kept it here, though I figured this one might just be an oversight.

zephyr/Kconfig.zenoh
Emil-Juhl: Sorry, even if I try the upstream main of zenoh-pico using platformio as described in the project root README, I cannot build the examples at all. This is the output I get where it seems that fundamental project configuration is wrong (no CONFIG_ZENOH_PICO exists, i.e. the extra kconfig from zenoh-pico has not been picked up):
```
-- Configuring incomplete, errors occurred!
[ ..snip.. ]
warning: ETH_NATIVE_POSIX (defined at drivers/ethernet/Kconfig.native_posix:6) was assigned the
value 'y' but got the value 'n'.
[ ..snip.. ]
warning: ETH_NATIVE_POSIX_RANDOM_MAC (defined at drivers/ethernet/Kconfig.native_posix:102) was
assigned the value 'y' but got the value 'n'.
[ ..snip.. ]
warning: NET_SOCKETS_POSIX_NAMES (defined at subsys/net/lib/sockets/Kconfig:13) was assigned the
value 'y' but got the value 'n'. 
[ ..snip.. ]
warning: TEST_RANDOM_GENERATOR (defined at subsys/random/Kconfig:8) was assigned the value 'y' but
got the value 'n'. Check these unsatisfied dependencies: (!ENTROPY_HAS_DRIVER) (=n).
[ ..snip.. ]
/home/emil/projects/zenoh-test/zephyr/prj.conf:81: warning: attempt to assign the value 'y' to the undefined symbol ZENOH_PICO

error: Aborting due to Kconfig warnings

CMake Error at /home/emil/.platformio/packages/framework-zephyr/cmake/kconfig.cmake:269 (message):
  command failed with return code: 1
Call Stack (most recent call first):
  /home/emil/.platformio/packages/framework-zephyr/cmake/app/boilerplate.cmake:542 (include)
  CMakeLists.txt:7 (include)
```
(I cut out a lot of extra text to just leave in the warnings and ultimately the failure). Of course I could be holding it wrong - I have never used platformio before, but just followed each step in the README and everything succeeded up until after adding zenoh-pico and trying to use the reel_board example as described.

I have also tried to build the examples against various zephyr versions and if I go all the way back to v4.0.0 of zephyr I get multiple warnings about the kconfig choices and ultimately fail to build with "No SOURCES given to target: app".
```
warning: ETH_NATIVE_POSIX (defined at drivers/ethernet/Kconfig.native_posix:6) was assigned the
value 'y' but got the value 'n'.
[ ..snip.. ]
warning: ETH_NATIVE_POSIX_RANDOM_MAC (defined at drivers/ethernet/Kconfig.native_posix:50) was
assigned the value 'y' but got the value 'n'.
[ ..snip.. ]
warning: NET_SOCKETS_POSIX_NAMES (defined at subsys/net/lib/sockets/Kconfig:26) was assigned the
value 'y' but got the value 'n'.
[ ..snip.. ]
-- Configuring done (2.9s)
CMake Error at /home/emil/projects/zephyr-zenoh-app/.zephyrproject/zephyr/cmake/modules/extensions.cmake:457 (add_library):
  No SOURCES given to target: app
  ```
  This was my attempt using west directly: `$ west build -p always -b reel_board lib/3rdparty/zenoh-pico/docs/zephyr/reel_board/`.
  
  So I must admit that while I can do a best guess fixup of the kconfig and their default values, the samples already seem fully broken on `main`. Actually, only two of the four board examples even set `CONFIG_ZENOH_PICO=y` in the first place. With this in mind, maybe the best approach is to figure out the docs and examples for using zenoh-pico with zephyr in a separate PR?

**Comment by Emil-Juhl** [comment]
@steils Thank you for the review! I have rebased again and fixed up most of the things you commented on. With regards to the zephyr examples and docs, I would actually suggest to give them a more in-depth update/refresh in a separate PR. It seems to me that they are already out of date on the main branch as-is.

File: zephyr/CMakeLists.txt [hunk 6]
```diff
@@ -1,39 +1,66 @@
...
+  ${CMAKE_CURRENT_SOURCE_DIR}/../include/zenoh-pico.h.in
+  ${CMAKE_CURRENT_SOURCE_DIR}/../include/zenoh-pico.h
+  @ONLY
+)
+
+function(assign_kconfig feature text)
+  string(REPLACE Z_FEATURE CONFIG_ZENOH_PICO kconfig ${feature})
+  if(${kconfig})
+    set(${feature} 1 CACHE STRING ${text})
```
steils: Both assignments in this function write to the CMake cache without overwriting the existing value. I reproduced this by configuring with `CONFIG_ZENOH_PICO_PUBLICATION=y`, then disabling it and reconfiguring the same build directory: Zephyr's .config showed the feature as disabled, but the generated config.h still contained Z_FEATURE_PUBLICATION 1.

Please make Kconfig authoritative on every configure, for example by using FORCE or by avoiding persistent cache entries for these values.

cmake/feature_config.cmake
```diff
@@ -0,0 +1,105 @@
+# Zenoh pico feature configuration options
+set(FRAG_MAX_SIZE 4096 CACHE STRING "Use this to override the maximum size for fragmented messages")
+set(BATCH_UNICAST_SIZE 2048 CACHE STRING "Use this to override the maximum unicast batch size")
```
steils, line 3: This makes BATCH_UNICAST_SIZE configurable through a CMake argument, but not through Zephyr Kconfig: there is still no corresponding integer option that can be set in prj.conf.
If this PR is intended to close #1070, please add and map something like CONFIG_ZENOH_PICO_BATCH_UNICAST_SIZE.

zephyr/CMakeLists.txt
steils: include/zenoh-pico.h is tracked, and it again during the Zephyr build is incorrect. It modifies the source tree and, with the current three-component version.txt, also replaces ZENOH_PICO_TWEAK 0 with an empty definition.

Please remove the Zephyr-side version parsing and this configure_file() block, and use the existing tracked include/zenoh-pico.h instead. The feature-specific zenoh-pico/config.h can continue to be generated in the binary include directory.

steils (changes requested):
Hello again. Sorry for the delay, and thanks for the update.
I see you have fixed Kconfig boolean conversion and other issues.
I agree that the already outdated Zephyr examples and documentation can be refreshed in a separate PR. I am also not insisting on the request to add default y: explicit feature selection is probably a reasonable policy for Zephyr.

I still found issues that should be addressed before merge, please see the inline comments.

cmake/feature_config.cmake
Emil-Juhl: Sorry, I never intended for this PR to solve #1070 - it was added to my initial PR description by @sashacmc , not by me. I'm not sure why, but I never bothered to ask.

Since it's there I decided to go ahead and make the kconfig cover all of the feature toggles as well as the integer settings like batch sizes. So it should be covered now.

I added this as a follow up commit on its own, but I hope to not have to extend the PR with further features as this has already grown way past the original intent which was really just plumbing/boilerplate for a better zephyr build.

zephyr/CMakeLists.txt
Emil-Juhl: Okay, I have removed it in a follow up commit.

But I'll repeat that the top-level CMakeLists.txt still does this exact auto-generation in-tree. I did not remove it in order to avoid creeping scope where I'd have to try all of the different builds as well.

zephyr/CMakeLists.txt
Emil-Juhl: Applied using the FORCE option.

Ideally the cmake logic would just set these values as the `CONFIG_ZENOH_PICO_...` string which zephyr cmake handles definition of. But then multiple `<setting> == 1` checks would need to be altered (hence why I just ended up with the 1 or 0 approach in the setter)

Emil-Juhl (commented):
@steils Thanks for another round of review! I should have addressed your comments now.

With regards to the default kconfig values I ended up a bit on the fence myself. Ultimately I think that trying to replicate the cmake defaults in kconfig is more prone to drifting and will by default include more than is necessary within zephyr builds. So I stayed with the requirement to explicitly opt-in to all feature flags.

I ended up extending the PR with the remaining features added to kconfig since I wanted to cover the unicast_size setting from #1070 . I figured I might as well make the list comprehensive.

fj-blanco:
Hi, I tested this PR on an ESP32 with Zephyr 4.1.0 while working on micro-ROS support through `rmw_zenoh_pico`.

The related work is described in [rmw_zenoh_pico issue #8](https://github.com/esol-community/rmw_zenoh_pico/issues/8). My test application is available in the [`test/zenoh-pico-pr-1083` branch of my `zephyr_apps` fork](https://github.com/fj-blanco/zephyr_apps/tree/test/zenoh-pico-pr-1083/apps/esp32_zenoh_talker).

I tested zenoh-pico commit `2bcb432`. This PR allowed me to remove the previous patch I required to use for generating `config.h` and mapping the Zephyr Kconfig options. The application can now select the required features directly from `prj.conf`, including disabling `TCP_NODELAY` for this ESP32 configuration.

During the hardware test, I found one additional issue in the Zephyr system backend. The current implementation of `z_realloc()` always returns `NULL`:

```c
void *z_realloc(void *ptr, size_t size) {
    // k_realloc not implemented in Zephyr
    return NULL;
}
```

The bytes serializer used by `rmw_zenoh_pico` needs to grow its buffer while creating the Zenoh attachment. As a result, the session opened correctly, but every publish failed with `_Z_ERR_SYSTEM_OUT_OF_MEMORY` (`-78`).

Zephyr 4.1.0 provides `k_realloc()`. I tested the following change:

```c
void *z_realloc(void *ptr, size_t size) {
    return k_realloc(ptr, size);
}
```

With this change, the ESP32 joined Wi-Fi, connected to a zenoh 1.10.1 router, initialized `rmw_zenoh_pico`, and published messages on `/chatter` once per second using the original memory configuration.

The small downstream patch used for the test is available [here](https://github.com/fj-blanco/zephyr_apps/blob/test/zenoh-pico-pr-1083/apps/esp32_zenoh_talker/patches/zenoh_pico_zephyr_realloc.patch).

Would this change be appropriate for this PR? If you need to retain support for Zephyr versions without `k_realloc()`, I can look into a compatible fallback. I would also be happy to prepare a separate PR or add a small Zephyr test if that is preferred.

Is this PR still planned for merge, and is there anything else I can test or help with?

Emil-Juhl:
> Hi, I tested this PR on an ESP32 with Zephyr 4.1.0 while working on micro-ROS support through `rmw_zenoh_pico`.
> 
> [... snip ...]
> 

Thanks for testing and I'm happy to hear that this has helped you out!

> The small downstream patch used for the test is available [here](https://github.com/fj-blanco/zephyr_apps/blob/test/zenoh-pico-pr-1083/apps/esp32_zenoh_talker/patches/zenoh_pico_zephyr_realloc.patch).
> 
> Would this change be appropriate for this PR? If you need to retain support for Zephyr versions without `k_realloc()`, I can look into a compatible fallback. I would also be happy to prepare a separate PR or add a small Zephyr test if that is preferred.
>

Thanks for sharing! I'd rather not try to push more stuff into this PR. Partly because I actually no longer use zenoh-pico for any project and as such I'd prefer to just get the existing set of changes merged. And partly because this PR has already been "stuck" for a long time (since I don't use zenoh-pico anymore I don't really push on it myself, so that's also on me) and more changes makes more churn makes more bugs makes more comments and so on :) But I would think adding a separate PR with more zephyr related stuff would be super!

> Is this PR still planned for merge, and is there anything else I can test or help with?
>

With regards to merging, I suppose maybe @steils or @sashacmc can give their take. I would hope that this can merge at some point. I honestly think the changes are meaningful, but I suspect that Zephyr support isn't high on the priority list which is fair enough. From my side, I just come back to rebase and/or fixup review comments whenever there is activity here.

Thanks for chiming in, it's nice knowing that the changes here has helped out others as well :)

PR #1084, opened 2025-11-06
Check allocation succeeds before appending.

### LINKED ISSUE #1074: [Bug] Malloc fail can yield into a memcpy to address 0x0
### Describe the bug

When using the `z_bytes_from_static_buf` and the system is out of memory. 
Then the internal function `_z_svec_append` will call `_z_svec_make` there malloc could fail.
As a result it keeps the `_z_svec_t` of 0 but `_z_svec_append` will still attempt to copy into it.

Image

### To reproduce

Start a publisher with a `z_bytes_from_static_buf`, then make sure your system memory is full and so malloc can fail then in my case the MPU triggered because a write occured into address which isn't allowed.

### System info

NXP i.MXRT1064
NuttX PX4

gmartin82:
Closes #1074.

### DIFF
--- a/src/protocol/iobuf.c
+++ b/src/protocol/iobuf.c
@@ -189,9 +189,12 @@ void _z_zbuf_free(_z_zbuf_t **zbf) {
 }
 
 /*------------------ WBuf ------------------*/
-static void _z_wbuf_add_iosli(_z_wbuf_t *wbf, _z_iosli_t *ios) {
-    wbf->_w_idx++;
-    _z_iosli_svec_append(&wbf->_ioss, ios, false);
+static z_result_t _z_wbuf_add_iosli(_z_wbuf_t *wbf, _z_iosli_t *ios) {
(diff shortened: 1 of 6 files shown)


src/collections/vec.c
```diff
@@ -240,6 +240,9 @@ z_result_t _z_svec_expand(_z_svec_t *v, z_element_move_f move, size_t element_si
 z_result_t _z_svec_append(_z_svec_t *v, const void *e, z_element_move_f move, size_t element_size, bool use_elem_f) {
     if (v->_capacity == 0) {
         *v = _z_svec_make(1, element_size);
+        if (v->_capacity == 0) {
```
DenisBiryukov91: I would rather change the logic of _z_svec_expand altogether to allow it to work on empty vectors

gmartin82: I've implemented your suggestion.

steils (approved):

PR #1091, opened 2025-11-14
seqnumber: Fix C++ interopability

PetervdPerk-NXP:
`_Atomic` and `std::atomic` are not guaranteed to be compatible. Additionally, `std::atomic` introduces dependencies on the C++ standard library, which is often too large for resource-constrained embedded systems. 

This change moves all atomic operations into seqnumber.c and uses the C11 atomic_xx operations, which are safe for both C and C++ callers. It also fixes a bug where `_z_seqnumber_null` was not implemented as an atomic operation.

Supersedes #1087 and #1089

### DIFF
--- a/include/zenoh-pico/collections/seqnumber.h
+++ b/include/zenoh-pico/collections/seqnumber.h
@@ -21,30 +21,22 @@
 #include "zenoh-pico/system/platform.h"
 #include "zenoh-pico/utils/result.h"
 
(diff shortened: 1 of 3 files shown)


gmartin82:
Thanks @PetervdPerk-NXP. In reviewing this PR I found an existing issue where the mutex wasn't being cleaned up correctly in the C99 case.

I've created yet another PR based on this PR to address this: https://github.com/eclipse-zenoh/zenoh-pico/pull/1092

PR #1095, opened 2025-11-17
collections: _z_slice_copy should handle zero-length slices

steils:
Treat zero-length inputs as _z_slice_null() so callers can always call _z_slice_copy() without special cases for zero-length input. Same for _z_slice_n_copy when len == 0.

### DIFF
--- a/src/collections/slice.c
+++ b/src/collections/slice.c
@@ -37,7 +37,11 @@ _z_delete_context_t _z_delete_context_static(void) { return _z_delete_context_cr
 
 /*-------- Slice --------*/
 z_result_t _z_slice_init(_z_slice_t *bs, size_t capacity) {
-    assert(capacity != 0);
+    assert(bs != NULL);
+    if (capacity == 0) {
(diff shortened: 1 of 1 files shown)


src/collections/slice.c
```diff
@@ -75,6 +75,10 @@ void _z_slice_free(_z_slice_t **bs) {
 }
 
 z_result_t _z_slice_copy(_z_slice_t *dst, const _z_slice_t *src) {
+    if (src == NULL || src->len == 0 || src->start == NULL) {
```
DenisBiryukov91: Do we really need to check for `src == NULL` or `src->start == NULL` given that we normally expect all structs to be properly initialized when calling functions on them? It is probably more appropriate to replace these checks with asserts `assert(src != NULL)`  and `assert(src->len == 0 || src->start != NULL)`.

DenisBiryukov91:
I think we also need to fix _z_slice_eq and _z_slice_init to work with 0-length slices

PR #1097, opened 2025-11-25
Support automatic tasks start/stop

steils:
Add autostart flags per task (read/lease/periodic), default autostart for read and lease tasks.

### DIFF
--- a/src/api/api.c
+++ b/src/api/api.c
@@ -2359,16 +2359,12 @@ bool zp_read_task_is_running(const z_loaned_session_t *zs) {
         return false;
     }
     const _z_session_t *session = _Z_RC_IN_VAL(zs);
-    switch (session->_tp._type) {
-        case _Z_TRANSPORT_UNICAST_TYPE:
-            return session->_tp._transport._unicast._common._read_task_running;
-        case _Z_TRANSPORT_MULTICAST_TYPE:
-            return session->_tp._transport._multicast._common._read_task_running;
-        case _Z_TRANSPORT_RAWETH_TYPE:
-            return session->_tp._transport._raweth._common._read_task_running;
-        default:
-            return false;
+    _z_transport_common_t *common = _z_transport_get_common((_z_transport_t *)&session->_tp);
+
+    if (common == NULL) {
+        return false;
     }
+    return common->_read_task_running;
 #else
     _ZP_UNUSED(zs);
     return false;
@@ -2417,16 +2413,12 @@ bool zp_lease_task_is_running(const z_loaned_session_t *zs) {
         return false;
     }
     const _z_session_t *session = _Z_RC_IN_VAL(zs);
-    switch (session->_tp._type) {
-        case _Z_TRANSPORT_UNICAST_TYPE:
-            return session->_tp._transport._unicast._common._lease_task_running;
-        case _Z_TRANSPORT_MULTICAST_TYPE:
-            return session->_tp._transport._multicast._common._lease_task_running;
-        case _Z_TRANSPORT_RAWETH_TYPE:
-            return session->_tp._transport._raweth._common._lease_task_running;
-        default:
-            return false;
+    _z_transport_common_t *common = _z_transport_get_common((_z_transport_t *)&session->_tp);
+
+    if (common == NULL) {
+        return false;
     }
+    return common->_lease_task_running;
 #else
     _ZP_UNUSED(zs);
     return false;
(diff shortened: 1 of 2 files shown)


src/api/api.c
```diff
@@ -759,7 +771,26 @@ static z_result_t _z_session_rc_init(z_owned_session_t *zs, _z_id_t *zid) {
 }
 
 z_result_t z_open(z_owned_session_t *zs, z_moved_config_t *config, const z_open_options_t *options) {
+#if Z_FEATURE_MULTI_THREAD == 1
```
DenisBiryukov91, line 775: may be it is simpler to replace this logic with 
```
z_open_options_t opts;
if (options == NULL) {
     z_open_options_default(&opts);
} else {
   opts = *options;
}
```
to avoid copy-pasting default values ?

src/api/api.c
```diff
@@ -788,6 +809,50 @@ z_result_t z_open(z_owned_session_t *zs, z_moved_config_t *config, const z_open_
...
+        }
+    }
+
+#ifdef Z_FEATURE_UNSTABLE_API
+#if Z_FEATURE_PERIODIC_TASKS == 1
+    if (opts.auto_start_periodic_task) {
+        z_result_t task_ret = _zp_start_periodic_scheduler_task(session, NULL);
+        if (task_ret != _Z_RES_OK) {
+            if (opts.auto_start_read_task) {
+                _zp_stop_read_task(session);
```
DenisBiryukov91: If I'm not mistaken, session automatically stops tasks on drop, so it is not necessary to do it manually.

Using _Z_SET_IF_OK(...) to chain fallible actions and calling 
```            
z_session_drop(z_session_move(zs));
z_config_drop(config);
``` 
only once in the end should also simplify the code

src/transport/unicast/read.c
```diff
@@ -333,7 +333,15 @@ void *_zp_unicast_read_task(void *ztu_arg) {
 
     // Prepare the buffer
     _z_zbuf_reset(&ztu->_common._zbuf);
-    z_whatami_t mode = _z_transport_common_get_session(&ztu->_common)->_mode;
+
+    _z_session_t *session = _z_transport_common_get_session(&ztu->_common);
+    if (session == NULL) {
```
DenisBiryukov91: This check should not be necessary since transport should outlive the session (at least this is the assumption used in `_z_transport_common_get_session`)

include/zenoh-pico/api/primitives.h
```diff
@@ -1548,6 +1548,11 @@ void z_scout_options_default(z_scout_options_t *options);
  */
 z_result_t z_open(z_owned_session_t *zs, z_moved_config_t *config, const z_open_options_t *options);
 
+/**
+ * Builds a :c:type:`z_open_options_t` with default value.
+ */
+void z_open_options_default(z_open_options_t *options);
```
DenisBiryukov91, line 1554: we need to add all public api to docs/api.rst

include/zenoh-pico/api/types.h
```diff
@@ -174,9 +174,24 @@ typedef struct {
 
 /**
  * Represents the configuration used to configure a zenoh upon opening :c:func:`z_open`.
+ *
+ * Members (multi-thread builds):
+ *   bool auto_start_read_task: auto-start read task after `z_open()` (default true).
+ *   bool auto_start_lease_task: auto-start lease task after `z_open()` (default true).
+ *   bool auto_start_periodic_task: auto-start periodic scheduler (default false; only when periodic tasks enabled).
```
DenisBiryukov91:
```suggestion
 *   bool auto_start_periodic_task: auto-start periodic scheduler (default false; only when periodic tasks feature is enabled).
```

DenisBiryukov91 (approved):

src/api/api.c
```diff
@@ -2294,13 +2342,39 @@ z_result_t zp_start_read_task(z_loaned_session_t *zs, const zp_task_read_options
...
 #endif
 }
 
+bool zp_read_task_is_running(const z_loaned_session_t *zs) {
+#if Z_FEATURE_MULTI_THREAD == 1
+    if (_Z_RC_IS_NULL(zs)) {
+        return false;
+    }
+    const _z_session_t *session = _Z_RC_IN_VAL(zs);
+    switch (session->_tp._type) {
```
sashacmc: Please use `_z_transport_get_common` to avoid coping this switch in many places

PR #1113, opened 2025-12-08
Add missing Z_FEATURE_LINK_TLS arg to GNUmakefile and fix TLS tests

gmartin82:
The integration tests for TLS implemented in https://github.com/eclipse-zenoh/zenoh-pico/pull/1025 do not run due to the missing arg in GNUmakefile.

See:

https://github.com/eclipse-zenoh/zenoh-pico/blob/main/GNUmakefile#L93
https://github.com/eclipse-zenoh/zenoh-pico/blob/main/CMakeLists.txt#L741
https://github.com/eclipse-zenoh/zenoh-pico/actions/runs/19823589143/job/56791387490#step:5:19
https://github.com/eclipse-zenoh/zenoh-pico/actions/runs/19823589143/job/56791387490#step:6:175

Link 3 does not show the Z_FEATURE_LINK_TLS in the CMake args. Link 4 must contain the command with the arg --enable-tls, but it doesn't.

In addition, the test failures found while addressing this issue have been fixed.

---
## 🏷️ Label-Based Checklist

**Labels:** `bug`

## 🐛 Bug Fix Requirements

**Instructions:**

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -421,10 +421,12 @@ if(Z_FEATURE_LINK_TLS)
   message(STATUS "Using Mbed TLS ${MBEDTLS_VERSION}")
 
   pkg_search_module(MBEDX509 OPTIONAL
+    mbedx509-3>=3.0.0
     mbedx509>=3.0.0
     mbedx509>=2.0.0)
 
   pkg_search_module(MBEDCRYPTO OPTIONAL
+    mbedcrypto-3>=3.0.0
     mbedcrypto>=3.0.0
     mbedcrypto>=2.0.0)
 


CMakeLists.txt
```diff
@@ -410,17 +410,30 @@ if(Z_FEATURE_LINK_TLS)
...
-    target_include_directories(${Libname}_static PRIVATE ${MBEDTLS_INCLUDE_DIRS})
-  endif()
-  if(PICO_SHARED)
-    target_include_directories(${Libname}_shared PRIVATE ${MBEDTLS_INCLUDE_DIRS})
-  endif()
+    # Start from what pkg-config gave us (usually just "mbedtls")
+    set(MBEDTLS_LINK_LIBRARIES ${MBEDTLS_LIBRARIES})
+
+    # Ensure we also link the other components by *name*
+    list(APPEND MBEDTLS_LINK_LIBRARIES mbedx509 mbedcrypto)
```
steils: Here it may add unversioned mbedx509 and mbedcrypto. On a system with both versions 2 and 3 installed, the unversioned names resolve to the older libs, so both versions get linked and mix ABIs.

CMakeLists.txt
```diff
@@ -410,17 +410,33 @@ if(Z_FEATURE_LINK_TLS)
...
     message(STATUS "Using Mbed TLS ${MBEDTLS_VERSION}")
-  endif()
 
-  if(PICO_STATIC)
-    target_include_directories(${Libname}_static PRIVATE ${MBEDTLS_INCLUDE_DIRS})
-  endif()
-  if(PICO_SHARED)
-    target_include_directories(${Libname}_shared PRIVATE ${MBEDTLS_INCLUDE_DIRS})
-  endif()
+    pkg_check_modules(MBEDTLS REQUIRED mbedtls)
```
steils: Again  it unconditionally runs pkg_check_modules for unversioned mbedtls/mbedx509/mbedcrypto, overwriting the package found by the initial `pkg_search_module`. On a system with both v2 and v3 installed this can still find a different library version.

steils (changes requested):
Can you try like this? (it works for me, I hope it wouldn't break CI)

```
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 9b530c72..14fcbfa9 100755
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -417,14 +417,7 @@ if(Z_FEATURE_LINK_TLS)
     endif()
     message(STATUS "Using Mbed TLS ${MBEDTLS_VERSION}")

-    pkg_check_modules(MBEDTLS REQUIRED mbedtls)
-    pkg_check_modules(MBEDX509 REQUIRED mbedx509)
-    pkg_check_modules(MBEDCRYPTO REQUIRED mbedcrypto)
-
-    set(MBEDTLS_LINK_LIBRARIES
-      ${MBEDTLS_LIBRARIES}
-      ${MBEDX509_LIBRARIES}
-      ${MBEDCRYPTO_LIBRARIES})
+    set(MBEDTLS_LINK_LIBRARIES ${MBEDTLS_LIBRARIES})

     if(PICO_STATIC)
       target_include_directories(${Libname}_static PUBLIC ${MBEDTLS_INCLUDE_DIRS})
@@ -657,6 +650,14 @@ if(UNIX OR MSVC)
     target_link_libraries(z_condvar_wait_until_test zenohpico::lib)
     target_link_libraries(z_sync_group_test zenohpico::lib)
     target_link_libraries(z_cancellation_token_test zenohpico::lib)
+    if(Z_FEATURE_LINK_TLS AND MBEDTLS_FOUND)
+      target_include_directories(z_tls_config_test PRIVATE ${MBEDTLS_INCLUDE_DIRS})
+      target_link_libraries(z_tls_config_test ${MBEDTLS_LIBRARIES})
+      find_library(MBEDTLS_CRYPTO_LIBRARY NAMES mbedcrypto-3 mbedcrypto PATHS ${MBEDTLS_LIBRARY_DIRS})
+      if(MBEDTLS_CRYPTO_LIBRARY)
+        target_link_libraries(z_tls_config_test ${MBEDTLS_CRYPTO_LIBRARY})
+      endif()
+    endif()
     target_link_libraries(z_local_loopback_test zenohpico::lib)
     target_include_directories(z_local_loopback_test PRIVATE ${PROJECT_SOURCE_DIR}/include)
     target_compile_definitions(z_local_loopback_test PRIVATE Z_LOOPBACK_TESTING=1)
```

CMakeLists.txt
gmartin82: That patch doesn't work for me when running the same command as the CI.

Also the examples will need to be linked against the libraries as well so just adding the libraries for specific tests won't work.

gmartin82: Okay, hopefully the latest change should fix it on all systems. Part of the problem is that when building from source mbedtls pkg config doesn't include the other libraries.

PR #1125, opened 2025-12-19
Add an admin space

gmartin82:
## Description

This PR introduces an initial implementation of an admin space for zenoh-pico, exposing runtime session and transport information via a queryable hierarchy under:

`@/<zid>/pico/**`

The admin space is implemented as a session-attached queryable and can be:
* started and stopped explicitly via the public API, or
* started automatically when opening a session using the appropriate open option.

The implementation returns structured JSON describing:
* the local session mode (whatami)
* active transports (unicast, multicast, raweth)
* transport links and their capabilities
* connected peers and peer-specific information

In addition, this PR adds a test suite that:
* verifies explicit and automatic start/stop behaviour
* validates that queries succeed only when the admin space is running
* checks that the returned JSON contains the expected structure and fields

The admin space API is currently marked unstable and may evolve in future releases as the schema and exposed information are refined.

---
## 🏷️ Label-Based Checklist

**Labels:** `enhancement`

## ✨ Enhancement Requirements

**Instructions:**

### DIFF
--- a/src/api/admin_space.c
+++ b/src/api/admin_space.c
@@ -229,7 +229,7 @@ static z_result_t _ze_admin_space_encode_peer_multicast(_z_json_encoder_t *je,
     return _Z_RES_OK;
 }
 
-void _ze_admin_space_reply_null(_ze_admin_space_reply_t *reply) {
+static void _ze_admin_space_reply_null(_ze_admin_space_reply_t *reply) {
     z_internal_keyexpr_null(&reply->ke);
     z_internal_bytes_null(&reply->payload);
 }
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -28,7 +28,9 @@ with open("../version.txt", "rt") as f:
 master_doc = 'index'
 extensions = ['sphinx_c_autodoc', 'sphinx_c_autodoc.napoleon']
 language = 'c'
-c_autodoc_roots = ['../include/zenoh-pico/api/', '../include/zenoh-pico/system/']
+c_autodoc_roots = ['../include/zenoh-pico/api/',
+                   '../include/zenoh-pico/collections/',
+                   '../include/zenoh-pico/system/']
 c_autodoc_compilation_args = [
     "-DSPHINX_DOCS",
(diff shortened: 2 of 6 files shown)


src/api/admin_space.c
```diff
@@ -0,0 +1,432 @@
...
+    z_owned_string_t s;
+    _Z_RETURN_IF_ERR(z_id_to_string(id, &s));
+    z_result_t r = _z_keyexpr_append_substr(ke, z_string_data(z_string_loan(&s)), z_string_len(z_string_loan(&s)));
+    z_string_drop(z_string_move(&s));
+    return r;
+}
+
+// ke = _Z_KEYEXPR_AT / ZID / _Z_KEYEXPR_PICO / _Z_KEYEXPR_SESSION
+static z_result_t _ze_admin_space_ke(z_owned_keyexpr_t *ke, const z_id_t *zid) {
+    z_internal_keyexpr_null(ke);
```
DenisBiryukov91: maybe it is better to construct the resulting ke string in place instead of calling _z_keyexpr_append_str multiple times which performs an allocation on each call ?
Same for other similar functions below.

gmartin82: I've changed the implementation so that the key expressions are formed from a composed string rather than via multiple joins.

src/api/admin_space.c
```diff
@@ -0,0 +1,463 @@
...
+    if (_Z_RC_IS_NULL(&session_rc)) {
+        _Z_ERROR("Dropped admin space query - session closed");
+        return;
+    }
+
+    _z_session_t *session = _Z_RC_IN_VAL(&session_rc);
+    z_id_t zid = session->_local_zid;
+
+    z_result_t res = _Z_RES_OK;
+
```
DenisBiryukov91, line 450: Here we try to access transport internals, not protected by mutex, it looks like it could lead to a data race in case of transport failure/reconnect, given that lease task is running independently of rx.

gmartin82: I've changed the code to hold the session and common transport mutexes when accessing the internals.

PR #1126, opened 2025-12-23
 ThreadX: Fix auto-reconnect and session cleanup

g4sp3r:
In our system (stm32 - threadx) with serial link to rmw_zenohd we experience issues if router gets restarted - zenoh-pico client would just keep publishing with no indication anything is wrong.
After restarting the router the topics are of course not shown because no re-connection attempt is done by the client. This is even with enabling Z_FEATURE_AUTO_RECONNECT.

1. First issue is found in `src/transport/unicast/read.c` where `peer->common._received` is always set to `true` even if no data is received. Probably this problem was created by STM32 ThreadX network.c implementation which uses non blocking read in `_z_read_serial_internal`. Non-blocking read was needed because otherwise client never re-sends the initial handshake to the router, meaning router would need to be started before the client.
2. With 1 fixed, the client now detects loss of lease when the router is restarted. Next problem is closing and reopening the connection. Some ThreadX system support functions were missing to properly stop all running tasks and are added in this PR.

We tried using Z_FEATURE_AUTO_RECONNECT unsuccessfully, so in our system we currently check if the read_task is still running, if not we undeclare pub/sub and close the session. After small timeout we again open the session and declare pub/sub.
Like this we finally have a system that survives common communication disturbances without any reboots.

---
## 🏷️ Label-Based Checklist

Current labels: _No labels_


g4sp3r:
Needs more work to support Z_FEATURE_AUTO_RECONNECT=1

g4sp3r:
Current version working with both manual and automatic reconnect.

Added cleanup queue for tasks that need to self-terminate (auto-reconnect case). Tasks are freed from the queue when creating new tasks. Currently one task remains in queue between reconnects due to _z_reopen() creating new tasks before _z_task_exit() is called. 

Another issue is found in manual reconnect case where trying to drop session and undeclare resources after lease expires leads to sending messages with cleared link. This locks the thread in our system in assert->exit() and leads to segmentation fault in tests with tcp Linux client. Adding the NULL checks in send_msg allows us to release all the memory before re-opening the connection again.

After quick test of memory usage we observe ~22KB used with Z_FEATURE_AUTO_RECONNECT=0 and ~40KB used with Z_FEATURE_AUTO_RECONNECT=1. Considering memory usage is x2 and we currently loose another 4KB with the "stuck" task we will be using the manual re-connection method handled by application logic for now.

knedl1k:
Any update on this?

PR #1139, opened 2026-01-14
Fix deadlock in advanced subscriber when recovery queries are answered locally.

### LINKED ISSUE #1138: [Bug] Deadlock with local Advanced Subscriber when recovery queries are answered locally
### Describe the bug

A deadlock occurs in Zenoh Pico Advanced Subscriber when a recovery or retransmission query is issued from inside a subscriber callback while holding the subscriber state mutex, and that query is answered locally (same session, AdvancedPublisher cache).

This happens with local subscribers + local queryables and does not require network transport or packet loss to reproduce.

The issue manifests as a thread blocking indefinitely on a mutex already held by the same thread (non-recursive mutex re-entry).

### To reproduce

1. Have an advanced publisher and subscriber within the same session on the same key expression with the following settings:

ze_advanced_publisher_options_t pub_opts;
ze_advanced_publisher_options_default(&pub_opts);
pub_opts.cache.is_enabled = true;
pub_opts.cache.max_samples = 1;
pub_opts.sample_miss_detection.is_enabled = true;

ze_advanced_subscriber_options_t sub_opts;
ze_advanced_subscriber_options_default(&sub_opts);
ze_advanced_subscriber_history_options_default(&sub_opts.history);
ze_advanced_subscriber_recovery_options_default(&sub_opts.recovery);
ze_advanced_subscriber_last_sample_miss_detection_options_default(&sub_opts.recovery.last_sample_miss_detection);

2. Publish a sample.

### System info

- Platform: Ubuntu 24.04
- Zenoh Version: 1.7.2

gmartin82:
Where the `Z_FEATURE_LOCAL_QUERYABLE` feature is enabled, local queries are executed synchronously in the calling thread. In certain code paths, the advanced subscriber initiated queries while holding an internal mutex, causing query callbacks to attempt to re-lock the same mutex, which resulted in a deadlock.

This change fixes the issue by ensuring queries are never invoked while holding the subscriber mutex, and by enforcing correct lock/unlock and rollback semantics when query execution fails. The fix is minimal and scoped to the affected advanced subscriber callbacks.

Fixes #1138.

---
## 🏷️ Label-Based Checklist

**Labels:** `bug`

## 🐛 Bug Fix Requirements

**Instructions:**

### DIFF
--- a/src/api/advanced_subscriber.c
+++ b/src/api/advanced_subscriber.c
@@ -996,23 +996,13 @@ static void _ze_advanced_subscriber_periodic_query_handler(void *ctx) {
         return;
     }
 
-    z_owned_keyexpr_t query_keyexpr;
-    if (z_keyexpr_clone(&query_keyexpr, z_keyexpr_loan(&state->_query_keyexpr)) != _Z_RES_OK) {
-#if Z_FEATURE_MULTI_THREAD == 1
-        z_mutex_unlock(z_mutex_loan_mut(&states->_mutex));
-#endif
-        _Z_WARN("Failed to clone keyexpr for periodic query");
-        return;
-    }
-
     state->_pending_queries++;
 
 #if Z_FEATURE_MULTI_THREAD == 1
     z_mutex_unlock(z_mutex_loan_mut(&states->_mutex));
 #endif
-    res = _ze_advanced_subscriber_sequenced_query(&query_ctx->_statesref, z_keyexpr_loan(&query_keyexpr), params,
-                                                  &query_ctx->source_id);
-    z_keyexpr_drop(z_keyexpr_move(&query_keyexpr));
+    res = _ze_advanced_subscriber_sequenced_query(&query_ctx->_statesref, z_keyexpr_loan(&state->_query_keyexpr),
+                                                  params, &query_ctx->source_id);
     if (res != _Z_RES_OK) {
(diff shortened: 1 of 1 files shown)


src/api/advanced_subscriber.c
```diff
@@ -1070,68 +1100,104 @@ static z_result_t __unsafe_ze_advanced_subscriber_spawn_periodic_query(_ze_advan
...
-            _Z_ERROR("Failed to handle sample in subscriber callback: %i", ret);
+            _Z_ERROR("Failed to prepare query for missing samples");
             return;
         }
 
-        const z_source_info_t *source_info = z_sample_source_info(sample);
-        if (source_info != NULL) {
-            z_entity_global_id_t source_id = z_source_info_id(source_info);
+        z_owned_keyexpr_t query_keyexpr;
+        if (z_keyexpr_clone(&query_keyexpr, z_keyexpr_loan(&state->_query_keyexpr)) != _Z_RES_OK) {
```
DenisBiryukov91: do you really need to clone key expression here and in other places given that you are holding an rc copy of state and it will not be dropped until this function returns (which is largely enough to send this ke over the wire) ?

gmartin82: I've changed the code to avoid the clone.

PR #1140, opened 2026-01-15
subscription: optimize exact keyexpr lookup

steils:
### What does this PR do?
Add a fast path in subscription resolution for exact keyexprs with no wildcard, avoiding unnecessary keyexpr expansion/allocations.

Optimize message handling when receiving exact subscription keys.

---
## 🏷️ Label-Based Checklist

**Labels:** `enhancement`

## ✨ Enhancement Requirements

- [ ] **Performance improvement measured** - If applicable, before/after metrics provided
- [ ] **Documentation updated** - Existing docs updated to reflect improvements
- [ ] **User impact documented** - How users benefit from this enhancement

**Instructions:**

### DIFF
--- a/src/session/subscription.c
+++ b/src/session/subscription.c
@@ -30,6 +30,7 @@
 #include "zenoh-pico/session/utils.h"
 #include "zenoh-pico/utils/locality.h"
 #include "zenoh-pico/utils/logging.h"
+#include "zenoh-pico/utils/string.h"
 
 #if Z_FEATURE_SUBSCRIPTION == 1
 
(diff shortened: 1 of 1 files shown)


src/session/subscription.c
```diff
@@ -231,8 +232,14 @@ static z_result_t _z_subscription_get_infos(_z_session_t *zn, _z_subscriber_kind
     } else {  // Construct data and add to cache
         _Z_DEBUG("Resolving %d - %.*s on mapping 0x%x", infos->ke_in._id, (int)_z_string_len(&infos->ke_in._suffix),
                  _z_string_data(&infos->ke_in._suffix), (unsigned int)infos->ke_in._mapping);
-        infos->ke_out = __unsafe_z_get_expanded_key_from_key(zn, &infos->ke_in, false, peer);
-        ret = _z_keyexpr_has_suffix(&infos->ke_out) ? _Z_RES_OK : _Z_ERR_KEYEXPR_UNKNOWN;
+        if ((infos->ke_in._id == Z_RESOURCE_ID_NONE) && _z_keyexpr_has_suffix(&infos->ke_in) &&
```
DenisBiryukov91, line 235: You could just replace force_alias with true in __unsafe_z_get_expanded_key_from_key above and achieve the same. Also it is unclear why check for "$\*" (first you could really check only for "\*", since "$" never comes alone, and second if id == Z_RESOURCE_ID_NONE it means that ke is sent as is without any optimization so it does not require expansion)
The issue is that taking alias here seems unsafe, since you are aliasing ke from rx buffer. I'm not entirely sure but it seems that closing the session/transport during callback execution will invalidate the buffer and thus leave user with a dangling pointer.

steils: You are right that I could achieve the same with setting force_alias to true, thanks.
Regarding the dangling pointers. The callback is already receiving aliased data from RX buffer: payload and encoding at least. If I understand correctly these pointers point into the RX buffer as well, and may become invalid if the session/transport is closed during the callback. So if we are okay with that, why not alias the keyexpr as well?

PR #1142, opened 2026-01-22
Separate key expression into wire expression and actual key expression

DenisBiryukov91:
## Description
Separate key expression into wire expression and actual key expression as it is done in zenoh-rust.

### What does this PR do?
- Uses _z_wireexpr_t type when passing/receiving key exp on the wire - it corresponds to previous _z_keyexpr_t 
as it holds id + suffix + mapping.
- Makes _z_keyexpr_t always carry the complete key expression string, since storing id + suffix does not provide any tangible advantage, given that user-created kes, always carry a complete ke string, and kes provided to the user also do the same (to allow getting ke string pointer from _z_keyexpr_t)
- Stores ref-counted declaration information (including weak session pointer) inside _z_keyexpr_t - this allows to correctly convert ke into wireexpr and fixes the bug where ke declared on one session would be incorrectly passed on the wire from another.
- Makes key expression automatically trigger undeclare once ref count of its declaration drops to 0 - this ensures that no keyexprs are undeclared on the wire as long as user holds at least one ke referencing this declaration.
- Cleans up related code and fixes several related data races during zenoh entities decalration/undeclaration.

To make code similar to what we have in zenoh-rust and improve its readability and correctness.

### Related Issues
None yet.

---
## 🏷️ Label-Based Checklist

**Labels:** `internal`

## 🏠 Internal Change

**Instructions:**

### DIFF
--- a/src/api/api.c
+++ b/src/api/api.c
@@ -1960,10 +1960,11 @@ z_result_t z_declare_keyexpr(const z_loaned_session_t *zs, z_owned_keyexpr_t *ke
 z_result_t z_undeclare_keyexpr(const z_loaned_session_t *zs, z_moved_keyexpr_t *keyexpr) {
     _z_keyexpr_wire_declaration_rc_t *declaration = &keyexpr->_this._val._declaration;
     z_result_t ret = _Z_RES_OK;
-    if (!_Z_RC_IS_NULL(declaration) &&
-        !_z_keyexpr_wire_declaration_is_declared_on_session(_Z_RC_IN_VAL(declaration), _Z_RC_IN_VAL(zs))) {
+    if (_Z_RC_IS_NULL(declaration)) {
+        ret = _Z_ERR_INVALID;
+    } else if (!_z_keyexpr_wire_declaration_is_declared_on_session(_Z_RC_IN_VAL(declaration), _Z_RC_IN_VAL(zs))) {
         ret = _Z_ERR_KEYEXPR_DECLARED_ON_ANOTHER_SESSION;
-    } else if (!_Z_RC_IS_NULL(declaration)) {
+    } else if (_z_rc_strong_count(declaration->_cnt) == 1) {
         ret = _z_keyexpr_wire_declaration_undeclare(_Z_RC_IN_VAL(declaration));
     }
     z_keyexpr_drop(keyexpr);
--- a/src/net/primitives.c
+++ b/src/net/primitives.c
@@ -280,7 +280,6 @@ z_result_t _z_undeclare_subscriber(_z_subscriber_t *sub) {
         }
         _z_network_message_t n_msg;
         _z_n_msg_make_declare(&n_msg, declaration, _z_optional_id_make_none());
-        ret = _z_send_undeclare(zn, &n_msg);
         if (_z_send_undeclare(zn, &n_msg) != _Z_RES_OK) {
             ret = _Z_ERR_TRANSPORT_TX_FAILED;
         }
--- a/src/session/interest.c
+++ b/src/session/interest.c
@@ -545,7 +545,7 @@ z_result_t _z_interest_process_interest(_z_session_t *zn, const _z_wireexpr_t *w
         }
         if (ret == _Z_RES_OK && _Z_HAS_FLAG(flags, _Z_INTEREST_FLAG_SUBSCRIBERS)) {
             _Z_DEBUG("Sending declare subscribers");
-            _z_interest_send_decl_subscriber(zn, id, NULL, restr_key_opt);
+            ret = _z_interest_send_decl_subscriber(zn, id, NULL, restr_key_opt);
         }
         if (ret == _Z_RES_OK && _Z_HAS_FLAG(flags, _Z_INTEREST_FLAG_QUERYABLES)) {
             _Z_DEBUG("Sending declare queryables");


src/session/interest.c
```diff
@@ -512,37 +522,44 @@ z_result_t _z_interest_process_interest_final(_z_session_t *zn, uint32_t id) {
...
+    }
+    _z_keyexpr_t *restr_key_opt = _z_keyexpr_check(&restr_key) ? &restr_key : NULL;
+    z_result_t ret = _Z_RES_OK;
     // Current flags process
     if (_Z_HAS_FLAG(flags, _Z_INTEREST_FLAG_CURRENT)) {
         // Send all declare
         if (_Z_HAS_FLAG(flags, _Z_INTEREST_FLAG_KEYEXPRS)) {
             _Z_DEBUG("Sending declare resources");
-            _Z_RETURN_IF_ERR(_z_interest_send_decl_resource(zn, id, NULL, restr_key));
+            ret = _z_interest_send_decl_resource(zn, id, NULL, restr_key_opt);
```
gmartin82, line 544: I'm not sure if it's possible for more than one if block to trigger but a non Z_RES_OK ret is potentially overwritten here.

gmartin82 (changes requested):

src/session/interest.c
DenisBiryukov91: fixed

src/session/interest.c
```diff
@@ -512,37 +522,44 @@ z_result_t _z_interest_process_interest_final(_z_session_t *zn, uint32_t id) {
...
+        if (ret == _Z_RES_OK && _Z_HAS_FLAG(flags, _Z_INTEREST_FLAG_KEYEXPRS)) {
             _Z_DEBUG("Sending declare resources");
-            _Z_RETURN_IF_ERR(_z_interest_send_decl_resource(zn, id, NULL, restr_key));
+            ret = _z_interest_send_decl_resource(zn, id, NULL, restr_key_opt);
         }
-        if (_Z_HAS_FLAG(flags, _Z_INTEREST_FLAG_SUBSCRIBERS)) {
+        if (ret == _Z_RES_OK && _Z_HAS_FLAG(flags, _Z_INTEREST_FLAG_SUBSCRIBERS)) {
             _Z_DEBUG("Sending declare subscribers");
-            _Z_RETURN_IF_ERR(_z_interest_send_decl_subscriber(zn, id, NULL, restr_key));
+            _z_interest_send_decl_subscriber(zn, id, NULL, restr_key_opt);
```
steils: It doesn't check the return code here.

src/net/primitives.c
```diff
@@ -296,75 +269,73 @@ z_result_t _z_undeclare_subscriber(_z_subscriber_t *sub) {
...
+        if (zn->_mode == Z_WHATAMI_CLIENT) {
+            declaration = _z_make_undecl_subscriber(sub->_entity_id, NULL);
+        } else {
+            _z_wireexpr_t expr = _z_keyexpr_alias_to_wire(&_Z_RC_IN_VAL(&s)->_key, zn);
+            declaration = _z_make_undecl_subscriber(sub->_entity_id, &expr);
+        }
+        _z_network_message_t n_msg;
+        _z_n_msg_make_declare(&n_msg, declaration, _z_optional_id_make_none());
+        ret = _z_send_undeclare(zn, &n_msg);
+        if (_z_send_undeclare(zn, &n_msg) != _Z_RES_OK) {
```
steils, line 283: Here `_z_send_undeclare` is called the second time.
Shouldn't it be
```
if (ret != _Z_RES_OK)
```

src/api/api.c
```diff
@@ -2096,43 +1943,30 @@ z_result_t z_keyexpr_from_str(z_owned_keyexpr_t *key, const char *name) {
...
 z_result_t z_undeclare_keyexpr(const z_loaned_session_t *zs, z_moved_keyexpr_t *keyexpr) {
+    _z_keyexpr_wire_declaration_rc_t *declaration = &keyexpr->_this._val._declaration;
     z_result_t ret = _Z_RES_OK;
-
-    ret = _z_undeclare_resource(_Z_RC_IN_VAL(zs), keyexpr->_this._val._id);
+    if (!_Z_RC_IS_NULL(declaration) &&
+        !_z_keyexpr_wire_declaration_is_declared_on_session(_Z_RC_IN_VAL(declaration), _Z_RC_IN_VAL(zs))) {
+        ret = _Z_ERR_KEYEXPR_DECLARED_ON_ANOTHER_SESSION;
+    } else if (!_Z_RC_IS_NULL(declaration)) {
+        ret = _z_keyexpr_wire_declaration_undeclare(_Z_RC_IN_VAL(declaration));
```
steils, line 1968: Here you undeclare the declaration bypassing the refcount, before `z_keyexpr_drop(keyexpr)`. What if the declaration has other users?
Shouldn't we just skip this `else if` path? Dropping will undeclare the declaration when RC drops to 0, right?

src/session/interest.c
DenisBiryukov91: fixed

src/api/api.c
DenisBiryukov91: fixed

src/net/primitives.c
DenisBiryukov91: fixed

PR #1149, opened 2026-02-05
Fix segfault when more than 4 subscribers or queryables are triggered on the same key expression.

### LINKED ISSUE #1148: [Bug] Segfault when more than 4 subscribers are triggered on the same key expression.
### Describe the bug

When collecting information about matching subscriptions `__unsafe_z_get_subscriptions_by_key()` currently appends to the `_z_subscription_rc_svec_t` using the element function for move. As this operation is not defined for `_z_subscription_rc_t` the function fails silently when vector expansion is needed leading to a later segfault.

### To reproduce

1. Start a publisher on a key expression.
2. Start 5 subscribers all on the same publisher key expression.
3. Observe a seg fault.

### System info

- Platform: Ubuntu 22.04

gmartin82:
## Description
When collecting information about matching subscriptions `__unsafe_z_get_subscriptions_by_key()` currently appends to the `_z_subscription_rc_svec_t` using the element function for move. As this operation is not defined for `_z_subscription_rc_t` the function fails silently when vector expansion is needed leading to a later segfault.

The same issue also affects queryables.

This PR fixes the issue.

### Related Issues
Fixes #1148

---
## 🏷️ Label-Based Checklist

**Labels:** `bug`

## 🐛 Bug Fix Requirements

**Instructions:**

### DIFF
--- a/src/session/queryable.c
+++ b/src/session/queryable.c
@@ -123,7 +123,7 @@ static z_result_t __unsafe_z_get_session_queryables_by_key(_z_session_t *zn, con
                                         : _z_locality_allows_local(qle_val->_allowed_origin);
         if (origin_allowed && _z_keyexpr_intersects(&qle_val->_key, key)) {
             _z_session_queryable_rc_t qle_clone = _z_session_queryable_rc_clone(qle);
-            _Z_CLEAN_RETURN_IF_ERR(_z_session_queryable_rc_svec_append(qle_infos, &qle_clone, true),
+            _Z_CLEAN_RETURN_IF_ERR(_z_session_queryable_rc_svec_append(qle_infos, &qle_clone, false),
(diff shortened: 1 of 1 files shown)


DenisBiryukov91:
@gmartin82 We should also do the same for _z_session_queryable_rc_svec_append and check other usage of append.
Overall I think we should rework our collections to make them more transparent and less error-prone (i.e. without the need to specify all _elem_ functions, and if some of them are not provided just do not implement those operations)

gmartin82:
@DenisBiryukov91 I've also added a test and fix for `_z_session_queryable_rc_svec_append`. I checked other append operations and the rest seem correct.

I agree it would be good to refactor the collections to avoid issues like this.

DenisBiryukov91 (approved):

PR #1150, opened 2026-02-06
session: preparse keyexpr to optimize matching

steils:
### What does this PR do?
Adds a preparsed keyexpr cache for subscriptions. Parses subscription keyexprs once at declare time. Extends keyexpr tests to verify that preparsed intersection results match regular.

Profiling showed wildcard matching with multiple subscriptions in the receive path as a bottleneck. Moving parsing to declaration time reduces per-message matching overhead.

---
## 🏷️ Label-Based Checklist

**Labels:** `internal`

## 🏠 Internal Change

**Instructions:**

### DIFF
--- a/src/net/primitives.c
+++ b/src/net/primitives.c
@@ -218,7 +218,6 @@ z_result_t _z_declare_subscriber(_z_subscriber_t *subscriber, const _z_session_r
     s._allowed_origin = allowed_origin;
     s._key = _z_keyexpr_null();
     _Z_CLEAN_RETURN_IF_ERR(_z_keyexpr_declare_non_wild_prefix(zn, &s._key, keyexpr), _z_subscription_clear(&s));
-    s._key._preparsed = _z_keyexpr_parse(&s._key);
 
     _z_subscription_rc_t sp_s = _z_register_subscription(_Z_RC_IN_VAL(zn), _Z_SUBSCRIBER_KIND_SUBSCRIBER, &s);
     if (_Z_RC_IS_NULL(&sp_s)) {
@@ -310,7 +309,6 @@ z_result_t _z_declare_queryable(_z_queryable_t *queryable, const _z_session_rc_t
     q._allowed_origin = allowed_origin;
     q._key = _z_keyexpr_null();
     _Z_CLEAN_RETURN_IF_ERR(_z_keyexpr_declare_non_wild_prefix(zn, &q._key, keyexpr), _z_session_queryable_clear(&q));
-    q._key._preparsed = _z_keyexpr_parse(&q._key);
 
     // Create session_queryable entry, stored at session-level, do not drop it by the end of this function.
     _z_session_queryable_rc_t sp_q = _z_register_session_queryable(_Z_RC_IN_VAL(zn), &q);
--- a/src/session/keyexpr.c
+++ b/src/session/keyexpr.c
@@ -1081,9 +1081,16 @@ size_t _z_keyexpr_non_wild_prefix_len(const _z_keyexpr_t *key) {
 
 z_result_t _z_keyexpr_declare_non_wild_prefix(const _z_session_rc_t *zs, _z_keyexpr_t *out,
                                               const _z_keyexpr_t *keyexpr) {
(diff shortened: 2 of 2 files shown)


include/zenoh-pico/session/keyexpr.h
```diff
@@ -58,13 +61,30 @@ typedef struct {
     _z_string_t _keyexpr;
 } _z_keyexpr_t;
 
+typedef struct {
+    _z_str_se_t _keyexpr;
+    size_t _n_verbatim;
+    int8_t _wildness;
+    size_t _n_chunks;
+    _z_str_se_t *_chunks;
```
DenisBiryukov91, line 64: I'm not sure it is worth storing both start and end of each chunk, given that they are contiguous - only starts should be enough

steils: Because everywhere in common helpers like `_z_ke_chunk_matcher` ` _z_str_se_t` is used. I would keep it here.

include/zenoh-pico/session/session.h
```diff
@@ -73,6 +73,7 @@ typedef void (*_z_closure_sample_callback_t)(_z_sample_t *sample, void *arg);
 
 typedef struct {
     _z_keyexpr_t _key;
+    _z_keyexpr_parsed_t _preparsed;
```
DenisBiryukov91: I suggest to either optionally store _preparsed data inside ke, or create a dedicated type that will store both: original ke and preparsed info.

steils: Done.

src/session/subscription.c
```diff
@@ -120,7 +119,7 @@ static z_result_t __unsafe_z_get_subscriptions_by_key(_z_session_t *zn, _z_subsc
         const _z_subscription_t *sub_val = _Z_RC_IN_VAL(sub);
         bool origin_allowed = is_remote ? _z_locality_allows_remote(sub_val->_allowed_origin)
                                         : _z_locality_allows_local(sub_val->_allowed_origin);
-        if (origin_allowed && _z_keyexpr_intersects_preparsed(&sub_val->_preparsed, &sub_val->_key, key)) {
+        if (origin_allowed && _z_keyexpr_intersects_preparsed(&sub_val->_key._preparsed, &sub_val->_key, key)) {
```
DenisBiryukov91: Why pass both ._key._preparsed and ._key ?

It is not clear why subscriber/queryable struct needs to know about internals of key expression, so it seems that either preparsed should be constructed on the fly (and be ref counted ?) during first call to intersect, or it is probably better to have a dedicated struct _z_preparsed_keyexpr_t containing simple _z_keyexpr_t and preparsed data, with a dedicated function:_z_preparsed_keyexpr_intersects(const _z_preparsed_keyexpr_t*, const _z_keyexpr_t*)

I think 2nd option is more preferable if you do not want all kes to end up filled with preparsed data.

steils: You are right. It's a leftover from when preparsed ke was a separate struct. Fixed.

src/net/primitives.c
```diff
@@ -216,7 +216,9 @@ z_result_t _z_declare_subscriber(_z_subscriber_t *subscriber, const _z_session_r
     s._dropper = dropper;
     s._arg = arg;
     s._allowed_origin = allowed_origin;
+    s._key = _z_keyexpr_null();
     _Z_CLEAN_RETURN_IF_ERR(_z_keyexpr_declare_non_wild_prefix(zn, &s._key, keyexpr), _z_subscription_clear(&s));
+    s._key._preparsed = _z_keyexpr_parse(&s._key);
```
DenisBiryukov91: Why expose internal details of key expression inside subscriber/queryable declaration ?

steils: Fixed

DenisBiryukov91:
The problem comes from intersection/inclusion algorithm in pico which runs through the ke string up to 3(!) times, so I believe the issue should be addressed by fixing the algorithm (which should improve performance by a factor of 3?) rather than by preparsing.

PR #1151, opened 2026-02-06
[ZEN-836] Cancellation refactoring

DenisBiryukov91:
## Description

Add method to remove on cancel handlers from cancellation tokens, to prevent unbounded memory growth until cancellation token is dropped/cancelled.

To prevent unbound memory growth in cancellation token.

### Related Issues
ZEN-836

---
## 🏷️ Label-Based Checklist

**Labels:** `internal`

## 🏠 Internal Change

**Instructions:**

There is indeed a slight behavior change - now the get request fails if passed cancellation token is already cancelled (similar to the case if session is already closed) - this is why corresponding cpp tests fail. This should be acceptable, given that cancellation functionality is unstable.

### DIFF
--- a/src/net/liveliness.c
+++ b/src/net/liveliness.c
@@ -247,7 +247,7 @@ z_result_t _z_liveliness_pending_query_register_cancellation(_z_liveliness_pendi
         pq->_cancellation_data._cancellation_token = _z_cancellation_token_rc_clone(opt_cancellation_token);
         pq->_cancellation_data._handler_id = handler_id;
         _Z_CLEAN_RETURN_IF_ERR(
-            _z_cancellation_token_get_notifier(_Z_RC_IN_VAL(opt_cancellation_token), &pq->_cancellation_data._notifer),
+            _z_cancellation_token_get_notifier(_Z_RC_IN_VAL(opt_cancellation_token), &pq->_cancellation_data._notifier),
             _z_pending_query_cancellation_data_clear(&pq->_cancellation_data));
     }
     return _Z_RES_OK;
--- a/include/zenoh-pico/session/session.h
+++ b/include/zenoh-pico/session/session.h
@@ -140,7 +140,7 @@ typedef struct _z_pending_query_t _z_pending_query_t;
 typedef struct {
     _z_cancellation_token_rc_t _cancellation_token;
     size_t _handler_id;
-    _z_sync_group_notifier_t _notifer;
+    _z_sync_group_notifier_t _notifier;
 } _z_pending_query_cancellation_data_t;
 
 static inline _z_pending_query_cancellation_data_t _z_pending_query_cancellation_data_null(void) {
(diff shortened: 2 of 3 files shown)


DenisBiryukov91:
> Looks good to me. Only comment is on the failing CPP tests (`test_cancellation_zenohpico`). Is this a problem or just something to be adjusted downstream?

Yes, as I mentioned it is a minor change in the behavior of unstable feature: now if user passes cancelled token into GET, the get will return a error (similar to what happens when session is closed), instead of silently doing nothing - so this requires a small adjustment to cpp tests for cancellation token.

src/net/primitives.c
```diff
@@ -532,30 +539,30 @@ z_result_t _z_query(_z_session_t *zn, const _z_keyexpr_t *keyexpr, const char *p
...
     _z_session_mutex_unlock(zn);
-    *out_qid = qid;
     // Send query message
     _z_slice_t params =
         (parameters == NULL) ? _z_slice_null() : _z_slice_alias_buf((uint8_t *)parameters, parameters_len);
-
-    if (remote_possible) {
+#if Z_FEATURE_LOCAL_QUERYABLE == 1
+    if (ret == _Z_RES_OK && allow_local) {
+        ret = _z_session_deliver_query_locally(zn, keyexpr, &params, consolidation, payload, encoding, attachment,
```
steils: Local delivery now runs before remote send. But `_z_session_deliver_query_locally()` steals payload, encoding, and attachment. When both local and remote delivery are `Z_LOCALITY_ANY` with a local queryable, won't  `_z_n_msg_make_query()` and remote queryables will receive these fields empty?
Is it necessary to change order from "remote then local" to "local then remote"?

src/net/liveliness.c
```diff
@@ -200,35 +200,108 @@ z_result_t _z_undeclare_liveliness_subscriber(_z_subscriber_t *sub) {
...
         _z_wireexpr_t wireexpr = _z_keyexpr_alias_to_wire(keyexpr, zn);
-        _z_interest_t interest = _z_make_interest(&wireexpr, id,
+        _z_interest_t interest = _z_make_interest(&wireexpr, query_id,
                                                   _Z_INTEREST_FLAG_KEYEXPRS | _Z_INTEREST_FLAG_TOKENS |
                                                       _Z_INTEREST_FLAG_RESTRICTED | _Z_INTEREST_FLAG_CURRENT);
         _z_network_message_t n_msg;
         _z_n_msg_make_interest(&n_msg, interest);
+        ret = _z_send_declare(zn, &n_msg);
+
         if (_z_send_declare(zn, &n_msg) != _Z_RES_OK) {
```
steils: `_z_send_declare` is called twice here.

tests/z_session_test.c
```diff
@@ -12,6 +12,7 @@
 //   ZettaScale Zenoh Team, <zenoh@zettascale.tech>
 //
 
+#if Z_FEATURE_MULTI_THEAD == 1
```
steils: `Z_FEATURE_MULTI_THREAD`

src/session/cancellation.c
```diff
@@ -17,214 +17,174 @@
...
+        (_z_cancellation_token_on_cancel_handler_t *)z_malloc(sizeof(_z_cancellation_token_on_cancel_handler_t));
+    if (handler_on_heap == NULL) {
+        return _Z_ERR_SYSTEM_OUT_OF_MEMORY;
     }
-    _z_cancellation_token_on_cancel_handler_drop(handler);
-    return ret;
+    _z_cancellation_token_on_cancel_handler_t *ret = _z_cancellation_token_on_cancel_handler_intmap_insert(
+        &storage->_handlers, storage->_next_handler_id, handler_on_heap);
+    if (ret == NULL) {
+        return _Z_ERR_SYSTEM_OUT_OF_MEMORY;
```
steils, line 61: `z_free(handler_on_heap)` here.

Issue #1007, opened 2025-08-01
Improve auto-reconnect feature


spynetcamera:
### Describe the feature

When using  the `Z_FEATURE_AUTO_RECONNECT` feature, on failure, the unicast/lease closes the transport, the enters a loop trying to reconnect.
This has some disadvantages:
- the `main` is not notified that something is going wrong, how can it know?
- if someone (the user) decided to close the connection, an error occurred since then transport is already shutdown, is there a safe way to cleanup?
- if the device can't reconnect, it seams like there's no way to abort the `_z_reopen()` loop.

It is okay to try reconnecting, but there should be a way to check this state and eventually abort it, let's say to try some other settings if needed.

Thank!

jean-roland:
Hello @spynetcamera, basically we need to implement an API to notify or let applications check connection status and events. This "connectivity event API" feature is planned but not currently in development.

DenisBiryukov91:
Resolved by connectivity API added in #1159

Issue #369, opened 2024-03-09
[Bug] [ESP-IDF] Unicast lease (possibly others?) does not kill session on timeout


meganukebmp:
### Describe the bug

Whenever the keepalive lease times out (for example, zenohd shut down), the session never actually gets closed. Doing `z_check(session)` keeps returning true as well. The session remains active so for example a publisher will keep publishing to nothing.

### To reproduce

1. Open a session to zenohd
2. Check if the session is up
3. Close session
4. z_check(session) returns true after timeout.

### System info

Zenohd 0.10.1
zenoh-pico 0.10.1
esp32-eth01, esp32-lolin

meganukebmp:
Dissecting this issue further. Log states session will close:
https://github.com/eclipse-zenoh/zenoh-pico/blob/main/src/transport/unicast/lease.c#L56

Then disables the lease loop. It then calls this:
https://github.com/eclipse-zenoh/zenoh-pico/blob/main/src/transport/unicast/transport.c#L281

That then simply creates a close message and transmits it to the router (if it exists, generic session close command)
https://github.com/eclipse-zenoh/zenoh-pico/blob/main/src/transport/unicast/transport.c#L272
But at no point does it actually close the session on its own.

meganukebmp:
Confirmed, this issue is present on all platforms.

Easy test on linux machine:
1. Run `zenohd -l tcp/192.168.1.100:7447 -l udp/192.168.1.100:7447`
2. Build the examples of zenoh-pico
3. Run `z_pub -n 999 -m client -e tcp/192.168.1.100:7447
4. Wait for session to connect and begin publishing messages
5. Kill the `zenohd` router
6. `z_pub` keeps publishing data to nothing forever

jean-roland:
Related to #333

Issue #333, opened 2024-01-29
Connectivity Events


p-avital:
### Describe the feature

Connectivity events allow users to detect when the topology of the network changes (within what's visible to their local node).

As of now, these events are only supported in Zenoh, and should be added to Pico.

This is however blocked by eclipse-zenoh/zenoh#692

PR #1158, opened 2026-02-16
[ZEN-836] Callback drop on Zenoh entities undeclare

DenisBiryukov91:
## Description

Ensure that user callbacks passed during creation of zenoh entities are dropped upon their undeclaration

### What does this PR do?
Ensures that user callbacks passed during creation of zenoh entities are dropped upon their undeclaration

### Why is this change needed?
To provide clear guarantees that user passed callbacks are destroyed before zenoh entities undeclare/drop methods return

### Related Issues
ZEN-836

---
## 🏷️ Label-Based Checklist

Based on the labels applied to this PR, please complete these additional requirements:

**Labels:** `enhancement`

## ✨ Enhancement Requirements

Since this PR enhances existing functionality:

- [x] **Enhancement scope documented** - Clear description of what is being improved
- [x] **Minimum necessary code** - Implementation is as simple as possible, doesn't overcomplicate the system
- [x] **Backwards compatible** - Existing code/APIs still work unchanged
- [x] **No new APIs added** - Only improving existing functionality
- [x] **Tests updated** - Existing tests pass, new test cases added if needed
- [x] **Performance improvement measured** - If applicable, before/after metrics provided
- [x] **Documentation updated** - Existing docs updated to reflect improvements
- [x] **User impact documented** - How users benefit from this enhancement

**Remember:** Enhancements should not introduce new APIs or breaking changes.

**Instructions:**
1. Check off items as you complete them (change `- [ ]` to `- [x]`)
2. The PR checklist CI will verify these are completed

*This checklist updates automatically when labels change, but preserves your checked boxes.*

PR #1159, opened 2026-02-17
Connectivity events and API

steils:
Connectivity status and events support for zenoh-pico, introduced in https://github.com/eclipse-zenoh/zenoh/pull/2301

- Add connectivity info and listener APIs for transports/links.
- Implement transport/link connectivity events on peer connect/disconnect.
- Enable this feature via Z_FEATURE_CONNECTIVITY (requires
  Z_FEATURE_UNSTABLE_API).
- Add test z_api_connectivity_test.
- Support the feature in z_info example.

### Related Issues

https://github.com/eclipse-zenoh/zenoh-pico/issues/333
https://github.com/eclipse-zenoh/zenoh-pico/issues/369
https://github.com/eclipse-zenoh/zenoh-pico/issues/1007

---
## 🏷️ Label-Based Checklist

**Labels:** `new feature`

## 🆕 New Feature Requirements

Since this PR adds a new feature:

- [x] **Feature scope documented** - Clear description of what the feature does and why it's needed
- [x] **New APIs well-designed** - Public APIs are intuitive, consistent with existing APIs
- [x] **Comprehensive tests** - All functionality is tested (happy path + edge cases + error cases)
- [x] **Examples provided** - Usage examples in code comments or separate example files
- [x] **Documentation added** - New docs explaining the feature, its use cases, and API
- [x] **Feature flag considered** - Can the feature be enabled/disabled for gradual rollout?
- [x] **Performance impact assessed** - Memory, CPU, storage implications measured
- [x] **Integration tested** - Feature works with existing features

**Consider:** Can this feature be split into smaller, incremental PRs?

**Instructions:**

### DIFF
--- a/src/api/admin_space.c
+++ b/src/api/admin_space.c
@@ -22,9 +22,7 @@
 
 #if Z_FEATURE_ADMIN_SPACE == 1
 
-#define _ZE_ADMIN_SPACE_KE_BUF_LEN 128
-#define _ZE_ADMIN_SPACE_SESSION_TRANSPORT_KE_BUF_LEN 192
-#define _ZE_ADMIN_SPACE_SESSION_LINK_KE_BUF_LEN 256
+#define _ZE_ADMIN_SPACE_KE_BUF_LEN 256
 
 typedef struct {
     const char *data;
@@ -168,60 +166,44 @@ static z_result_t _ze_admin_space_session_queryable_ke(z_owned_keyexpr_t *ke, co
 // ke = _Z_KEYEXPR_AT / ZID / _Z_KEYEXPR_SESSION / _Z_KEYEXPR_TRANSPORT_UNICAST / PEER_ZID
 static z_result_t _ze_admin_space_session_transport_ke(z_owned_keyexpr_t *ke, const z_id_t *zid,
                                                        const z_id_t *peer_zid) {
-    z_internal_keyexpr_null(ke);
-
-    z_owned_string_t zid_str;
-    _Z_RETURN_IF_ERR(z_id_to_string(zid, &zid_str));
     z_owned_string_t peer_zid_str;
-    _Z_CLEAN_RETURN_IF_ERR(z_id_to_string(peer_zid, &peer_zid_str), z_string_drop(z_string_move(&zid_str)));
+    _Z_RETURN_IF_ERR(z_id_to_string(peer_zid, &peer_zid_str));
 
-    char buf[_ZE_ADMIN_SPACE_SESSION_TRANSPORT_KE_BUF_LEN];
-    int n = snprintf(buf, sizeof(buf), "%s%s%.*s%s%s%s%s%s%.*s", _Z_KEYEXPR_AT, _Z_KEYEXPR_SEPARATOR,
-                     (int)z_string_len(z_string_loan(&zid_str)), z_string_data(z_string_loan(&zid_str)),
-                     _Z_KEYEXPR_SEPARATOR, _Z_KEYEXPR_SESSION, _Z_KEYEXPR_SEPARATOR, _Z_KEYEXPR_TRANSPORT_UNICAST,
-                     _Z_KEYEXPR_SEPARATOR, (int)z_string_len(z_string_loan(&peer_zid_str)),
-                     z_string_data(z_string_loan(&peer_zid_str)));
+    const z_loaned_string_t *peer_zid_loan = z_string_loan(&peer_zid_str);
+    const _ze_admin_space_ke_segment_t segments[] = {
+        {_Z_KEYEXPR_SESSION, _Z_KEYEXPR_SESSION_LEN},
+        {_Z_KEYEXPR_TRANSPORT_UNICAST, _Z_KEYEXPR_TRANSPORT_UNICAST_LEN},
+        {z_string_data(peer_zid_loan), z_string_len(peer_zid_loan)},
+    };
 
-    z_string_drop(z_string_move(&zid_str));
+    z_result_t ret = _ze_admin_space_build_ke(ke, zid, segments, 3);
     z_string_drop(z_string_move(&peer_zid_str));
-
-    if (n < 0 || (size_t)n >= sizeof(buf)) {
-        return _Z_ERR_INVALID;
-    }
-    return z_keyexpr_from_str(ke, buf);
+    return ret;
 }
 
 // ke = _Z_KEYEXPR_AT / ZID / _Z_KEYEXPR_SESSION / _Z_KEYEXPR_TRANSPORT_UNICAST / PEER_ZID / _Z_KEYEXPR_LINK /
 // LINK_ID
 static z_result_t _ze_admin_space_session_link_ke(z_owned_keyexpr_t *ke, const z_id_t *zid, const z_id_t *peer_zid,
                                                   const z_id_t *link_id) {
-    z_internal_keyexpr_null(ke);
-
-    z_owned_string_t zid_str;
-    _Z_RETURN_IF_ERR(z_id_to_string(zid, &zid_str));
     z_owned_string_t peer_zid_str;
-    _Z_CLEAN_RETURN_IF_ERR(z_id_to_string(peer_zid, &peer_zid_str), z_string_drop(z_string_move(&zid_str)));
+    _Z_RETURN_IF_ERR(z_id_to_string(peer_zid, &peer_zid_str));
     z_owned_string_t link_id_str;
-    _Z_CLEAN_RETURN_IF_ERR(z_id_to_string(link_id, &link_id_str), z_string_drop(z_string_move(&zid_str));
-                           z_string_drop(z_string_move(&peer_zid_str)));
+    _Z_CLEAN_RETURN_IF_ERR(z_id_to_string(link_id, &link_id_str), z_string_drop(z_string_move(&peer_zid_str)));
 
-    char buf[_ZE_ADMIN_SPACE_SESSION_LINK_KE_BUF_LEN];
-    int n = snprintf(buf, sizeof(buf), "%s%s%.*s%s%s%s%s%s%.*s%s%s%s%.*s", _Z_KEYEXPR_AT, _Z_KEYEXPR_SEPARATOR,
-                     (int)z_string_len(z_string_loan(&zid_str)), z_string_data(z_string_loan(&zid_str)),
-                     _Z_KEYEXPR_SEPARATOR, _Z_KEYEXPR_SESSION, _Z_KEYEXPR_SEPARATOR, _Z_KEYEXPR_TRANSPORT_UNICAST,
-                     _Z_KEYEXPR_SEPARATOR, (int)z_string_len(z_string_loan(&peer_zid_str)),
-                     z_string_data(z_string_loan(&peer_zid_str)), _Z_KEYEXPR_SEPARATOR, _Z_KEYEXPR_LINK,
-                     _Z_KEYEXPR_SEPARATOR, (int)z_string_len(z_string_loan(&link_id_str)),
-                     z_string_data(z_string_loan(&link_id_str)));
+    const z_loaned_string_t *peer_zid_loan = z_string_loan(&peer_zid_str);
+    const z_loaned_string_t *link_id_loan = z_string_loan(&link_id_str);
+    const _ze_admin_space_ke_segment_t segments[] = {
+        {_Z_KEYEXPR_SESSION, _Z_KEYEXPR_SESSION_LEN},
+        {_Z_KEYEXPR_TRANSPORT_UNICAST, _Z_KEYEXPR_TRANSPORT_UNICAST_LEN},
+        {z_string_data(peer_zid_loan), z_string_len(peer_zid_loan)},
+        {_Z_KEYEXPR_LINK, _Z_KEYEXPR_LINK_LEN},
+        {z_string_data(link_id_loan), z_string_len(link_id_loan)},
+    };
 
-    z_string_drop(z_string_move(&zid_str));
+    z_result_t ret = _ze_admin_space_build_ke(ke, zid, segments, 5);
     z_string_drop(z_string_move(&peer_zid_str));
     z_string_drop(z_string_move(&link_id_str));
-
-    if (n < 0 || (size_t)n >= sizeof(buf)) {
-        return _Z_ERR_INVALID;
-    }
-    return z_keyexpr_from_str(ke, buf);
+    return ret;
 }
 #endif  // Z_FEATURE_CONNECTIVITY == 1
 
@@ -413,33 +395,8 @@ void _ze_admin_space_reply_clear(_ze_admin_space_reply_t *reply) {
     _ze_admin_space_reply_null(reply);
 }
 
-static z_result_t _ze_admin_space_add_reply(_z_json_encoder_t *je, const z_loaned_keyexpr_t *ke,
-                                            _ze_admin_space_reply_list_t **replies) {
-    _ze_admin_space_reply_t *reply = z_malloc(sizeof(_ze_admin_space_reply_t));
-    if (reply == NULL) {
-        return _Z_ERR_SYSTEM_OUT_OF_MEMORY;
-    }
-    _ze_admin_space_reply_null(reply);
-
-    _Z_CLEAN_RETURN_IF_ERR(z_keyexpr_clone(&reply->ke, ke), z_free(reply));
-    _Z_CLEAN_RETURN_IF_ERR(_z_json_encoder_finish(je, &reply->payload), _ze_admin_space_reply_clear(reply);
-                           z_free(reply));
-
-    _ze_admin_space_reply_list_t *old = *replies;
-    _ze_admin_space_reply_list_t *tmp = _ze_admin_space_reply_list_push(*replies, reply);
-    if (tmp == old) {
-        _ze_admin_space_reply_clear(reply);
-        z_free(reply);
-        return _Z_ERR_SYSTEM_OUT_OF_MEMORY;
-    }
-    *replies = tmp;
-
-    return _Z_RES_OK;
-}
-
-#if Z_FEATURE_CONNECTIVITY == 1
-static z_result_t _ze_admin_space_add_reply_payload(const z_loaned_keyexpr_t *ke, z_moved_bytes_t *payload,
-                                                    _ze_admin_space_reply_list_t **replies) {
+static z_result_t _ze_admin_space_add_reply_bytes(const z_loaned_keyexpr_t *ke, z_moved_bytes_t *payload,
+                                                  _ze_admin_space_reply_list_t **replies) {
     _ze_admin_space_reply_t *reply = z_malloc(sizeof(_ze_admin_space_reply_t));
     if (reply == NULL) {
         z_bytes_drop(payload);
@@ -458,9 +415,17 @@ static z_result_t _ze_admin_space_add_reply_payload(const z_loaned_keyexpr_t *ke
         return _Z_ERR_SYSTEM_OUT_OF_MEMORY;
     }
     *replies = tmp;
+
     return _Z_RES_OK;
 }
-#endif  // Z_FEATURE_CONNECTIVITY == 1
+
+static z_result_t _ze_admin_space_add_reply(_z_json_encoder_t *je, const z_loaned_keyexpr_t *ke,
+                                            _ze_admin_space_reply_list_t **replies) {
+    z_owned_bytes_t payload;
+    z_internal_bytes_null(&payload);
+    _Z_RETURN_IF_ERR(_z_json_encoder_finish(je, &payload));
+    return _ze_admin_space_add_reply_bytes(ke, z_bytes_move(&payload), replies);
+}
 
 static z_result_t _ze_admin_space_encode_transport_common(_z_json_encoder_t *je, const _z_transport_common_t *common) {
     return _ze_admin_space_encode_link(je, common->_link);
@@ -857,8 +822,8 @@ static void _ze_admin_space_query_handle_connectivity_session(const z_loaned_que
                 ret = _ze_admin_space_encode_connectivity_transport_payload(&payload, &peer->common._remote_zid,
                                                                             peer->common._remote_whatami, false, false);
                 if (ret == _Z_RES_OK) {
-                    ret = _ze_admin_space_add_reply_payload(z_keyexpr_loan(&transport_ke), z_bytes_move(&payload),
-                                                            replies);
+                    ret =
+                        _ze_admin_space_add_reply_bytes(z_keyexpr_loan(&transport_ke), z_bytes_move(&payload), replies);
                 }
                 if (ret != _Z_RES_OK) {
                     _Z_WARN("Failed to add connectivity transport status reply: %d", ret);
@@ -884,7 +849,7 @@ static void _ze_admin_space_query_handle_connectivity_session(const z_loaned_que
                 ret = _ze_admin_space_encode_connectivity_link_payload(
                     &payload, &peer->common._link_src, &peer->common._link_dst, mtu, is_streamed, is_reliable);
                 if (ret == _Z_RES_OK) {
-                    ret = _ze_admin_space_add_reply_payload(z_keyexpr_loan(&link_ke), z_bytes_move(&payload), replies);
+                    ret = _ze_admin_space_add_reply_bytes(z_keyexpr_loan(&link_ke), z_bytes_move(&payload), replies);
                 }
                 if (ret != _Z_RES_OK) {
                     _Z_WARN("Failed to add connectivity link status reply: %d", ret);


docs/api.rst
```diff
@@ -1764,6 +1900,7 @@ Enabling the Admin Space
 
 The Admin Space is an **optional feature** and must be explicitly enabled at
 build time by defining the ``Z_FEATURE_ADMIN_SPACE`` configuration flag.
+It requires both ``Z_FEATURE_UNSTABLE_API`` and ``Z_FEATURE_CONNECTIVITY``.
```
gmartin82: The admin space feature should have no dependency on the connectivity feature.

If extra information can be contributed to the admin space when the connectivity feature is enabled, then only the provision of that additional information should be dependent on the connectivity feature.

docs/config.rst
```diff
@@ -165,4 +166,4 @@ The following options are here to reduce binary sizes for users that don't need
 * `Z_FEATURE_LINK_SERIAL`: (DEFAULT: OFF) Toggle compilation of Serial link support.
 * `Z_FEATURE_LINK_SERIAL_USB`: (DEFAULT: OFF) Toggle compilation of Serial USB link support.
 * `Z_FEATURE_LINK_TLS`: (DEFAULT: OFF) Toggle compilation of TLS support.
-* `Z_FEATURE_ADMIN_SPACE`: (DEFAULT: OFF) Toggle compilation of admin space API functions.
```
gmartin82, line 168: See previous comment. Admin space should not depend on the connectivity feature.

src/api/admin_space.c
```diff
@@ -241,17 +386,18 @@ void _ze_admin_space_reply_clear(_ze_admin_space_reply_t *reply) {
     _ze_admin_space_reply_null(reply);
 }
 
-static z_result_t _ze_admin_space_add_reply(_z_json_encoder_t *je, const z_loaned_keyexpr_t *ke,
-                                            _ze_admin_space_reply_list_t **replies) {
+static z_result_t _ze_admin_space_add_reply_payload(const z_loaned_keyexpr_t *ke, z_owned_bytes_t *payload,
+                                                    _ze_admin_space_reply_list_t **replies) {
     _ze_admin_space_reply_t *reply = z_malloc(sizeof(_ze_admin_space_reply_t));
     if (reply == NULL) {
+        z_bytes_drop(z_bytes_move(payload));
```
gmartin82: If this function can consume bytes, then it should accept a z_moved_bytes_t to make this explicit.

src/api/admin_space.c
```diff
@@ -93,141 +111,268 @@ static z_result_t _ze_admin_space_transport_link_ke(z_owned_keyexpr_t *ke, const
...
-            return _z_json_encoder_write_string(je, "peer");
-        case Z_WHATAMI_CLIENT:
-            return _z_json_encoder_write_string(je, "client");
-        default:
-            return _Z_ERR_INVALID;
+static inline const char *_ze_admin_space_transport_kind_from_session(const _z_session_t *session, bool is_multicast) {
+    if (session != NULL && session->_tp._type == _Z_TRANSPORT_RAWETH_TYPE) {
+        return _Z_KEYEXPR_TRANSPORT_RAWETH;
     }
+    return is_multicast ? _Z_KEYEXPR_TRANSPORT_MULTICAST : _Z_KEYEXPR_TRANSPORT_UNICAST;
```
gmartin82: Why are we determining the type in this way? _type is an enum value.

src/api/admin_space.c
```diff
@@ -93,141 +111,268 @@ static z_result_t _ze_admin_space_transport_link_ke(z_owned_keyexpr_t *ke, const
...
+    if (z_link_event_kind(event) == Z_SAMPLE_KIND_PUT) {
+        z_owned_bytes_t payload;
+        ret = _ze_admin_space_encode_link_event_payload(&payload, link);
+        if (ret == _Z_RES_OK) {
+            z_put_options_t opts;
+            z_put_options_default(&opts);
+            z_owned_encoding_t encoding;
+            if (z_encoding_clone(&encoding, z_encoding_application_json()) == _Z_RES_OK) {
+                opts.encoding = z_encoding_move(&encoding);
+                ret = z_put(&session_rc, z_keyexpr_loan(&ke), z_bytes_move(&payload), &opts);
```
gmartin82: This should be gated on `Z_FEATURE_PUBLICATION`. Admin space should still function without this feature if `Z_FEATURE_QUERYABLE` is enabled.

CMakeLists.txt
```diff
@@ -242,6 +242,7 @@ set(Z_TRANSPORT_LEASE_EXPIRE_FACTOR 3 CACHE STRING "Default session lease expire
 set(ZP_PERIODIC_SCHEDULER_MAX_TASKS 64 CACHE STRING "Maximum number of tasks in the periodic scheduler")
 
 set(Z_FEATURE_UNSTABLE_API 0 CACHE STRING "Toggle unstable Zenoh-C API")
+set(Z_FEATURE_CONNECTIVITY 1 CACHE STRING "Toggle connectivity status/events API (unstable)")
```
DenisBiryukov91, line 245: We should probably also expose Z_FEATURE_CONNECTIVITY in PackageConfig.cmake.in - since it affects public API and will be handy for conditional compilation of C++ tests/examples.

src/api/connectivity.c
```diff
@@ -0,0 +1,669 @@
...
+                                                         listener_state) == NULL) {
+        _z_session_mutex_unlock(session);
+        _z_connectivity_transport_listener_clear(listener_state);
+        z_free(listener_state);
+        ret = _Z_ERR_SYSTEM_OUT_OF_MEMORY;
+        goto exit;
+    }
+    _z_session_mutex_unlock(session);
+
+    _z_session_weak_t weak = _z_session_rc_clone_as_weak(zs);
```
DenisBiryukov91, line 553: this call normally can not fail, since it is just an atomic variable increase

include/zenoh-pico/net/session.h
```diff
@@ -39,6 +39,84 @@ extern "C" {
  */
 struct _z_write_filter_registration_t;
 
+#if Z_FEATURE_CONNECTIVITY == 1
+typedef struct {
+    _z_void_rc_t _callback;
```
DenisBiryukov91, line 45: If you store callbacks under rc and thus want to allow executing them outside of session mutex, you need to provide guarantees that callbacks will be dropped upon listener undeclaration/session close, by probably using sync_groups as introduced by https://github.com/eclipse-zenoh/zenoh-pico/pull/1158.

src/api/admin_space.c
steils: Fixed

docs/api.rst
steils: Fixed

src/api/api.c
```diff
@@ -896,6 +1041,206 @@ z_result_t z_info_routers_zid(const z_loaned_session_t *zs, z_moved_closure_zid_
...
+}
+
+void z_info_links_options_default(z_info_links_options_t *options) { options->transport = NULL; }
+
+z_result_t z_info_transports(const z_loaned_session_t *zs, z_moved_closure_transport_t *callback) {
+    z_result_t ret = _Z_RES_OK;
+    _z_session_t *session = _Z_RC_IN_VAL(zs);
+    _z_transport_t *transport = &session->_tp;
+    _z_transport_common_t *transport_common = _z_transport_get_common(transport);
+    if (transport_common != NULL) {
```
DenisBiryukov91, line 1074: This is not really thread-safe (see zp_unicast_failed)

src/api/connectivity.c
steils: Normally it's almost impossible. But it I see that `name##_rc_clone_as_weak` can return null, so why not check it?

DenisBiryukov91: just to remove boilerplate code. Indeed it can return null, but for this (outside of bugs in rc implementation) you would require to have ~2^32 copies of rc pointer which would require roughly 32 GB of ram - you will run out of memory much earlier.

steils: > outside of bugs in rc implementation

What if the rc implementation changes sometime? I would still keep the check here:  the cost of this check is minimal, while keeping the API usage explicit and correct.
Or if we assume that it cannot return null except for bugs, then I think it would be more correct to use an assert inside `name##_rc_clone_as_weak()`, not to remove checks at call places.

include/zenoh-pico/net/session.h
steils: Fixed, I hope

src/api/api.c
steils: I hope it would it be okay to add a new _mutex_transport. If so, fixed.

src/api/api.c
```diff
@@ -896,6 +1041,211 @@ z_result_t z_info_routers_zid(const z_loaned_session_t *zs, z_moved_closure_zid_
...
+
+    if (transport_common != NULL) {
+        _z_transport_peer_mutex_unlock(transport_common);
+    }
+    _z_session_transport_mutex_unlock(session);
+
+out:
+    z_transport_drop(opt.transport);
+
+    void *ctx = callback->_this._val.context;
```
DenisBiryukov91: why not just z_closure_link_drop() ?

src/api/api.c
```diff
@@ -896,6 +1041,211 @@ z_result_t z_info_routers_zid(const z_loaned_session_t *zs, z_moved_closure_zid_
...
+        default:
+            break;
+    }
+
+    if (transport_common != NULL) {
+        _z_transport_peer_mutex_unlock(transport_common);
+    }
+    _z_session_transport_mutex_unlock(session);
+
+    void *ctx = callback->_this._val.context;
```
DenisBiryukov91: why not just z_closure_transport_drop() ?

src/api/connectivity.c
```diff
@@ -0,0 +1,958 @@
...
+            (_z_connectivity_transport_cb_state_t *)callback_snapshot._val;
+        _z_connectivity_replay_transport_history(session, callback_state);
+        _z_void_rc_drop(&callback_snapshot);
+    }
+
+exit:
+    if (locked_transport_common != NULL) {
+        _z_transport_peer_mutex_unlock(locked_transport_common);
+    }
+    if (_z_sync_group_check(&callback_drop_sync_group)) {
```
DenisBiryukov91: no need for this check, _z_sync_group_t is double drop safe

src/api/connectivity.c
```diff
@@ -0,0 +1,958 @@
...
+        _z_connectivity_link_cb_state_t *callback_state = (_z_connectivity_link_cb_state_t *)callback_snapshot._val;
+        _z_connectivity_replay_link_history(session, callback_state, has_transport_filter, &transport_filter);
+        _z_void_rc_drop(&callback_snapshot);
+    }
+
+exit:
+    if (locked_transport_common != NULL) {
+        _z_transport_peer_mutex_unlock(locked_transport_common);
+    }
+    if (_z_sync_group_check(&callback_drop_sync_group)) {
```
DenisBiryukov91: No need for this check, since _z_sync_group_t is double-drop safe

src/api/api.c
steils: Thanks, fixed

src/api/api.c
steils: Thanks, fixed

src/api/connectivity.c
steils: Thanks, fixed

src/api/connectivity.c
steils: Thanks, fixed

src/api/admin_space.c
```diff
@@ -66,25 +74,32 @@ static z_result_t _ze_admin_space_transport_ke(z_owned_keyexpr_t *ke, const z_id
...
-
-    char buf[128];
-    int n = snprintf(buf, sizeof(buf), "%s%s%.*s%s%s%s%s%s%s%s%s%s%.*s", _Z_KEYEXPR_AT, _Z_KEYEXPR_SEPARATOR,
+    z_owned_string_t peer_zid_str;
+    _Z_CLEAN_RETURN_IF_ERR(z_id_to_string(peer_zid, &peer_zid_str), z_string_drop(z_string_move(&zid_str)));
+    z_owned_string_t link_id_str;
+    _Z_CLEAN_RETURN_IF_ERR(z_id_to_string(link_id, &link_id_str), z_string_drop(z_string_move(&zid_str));
+                           z_string_drop(z_string_move(&peer_zid_str)));
+
+    char buf[256];
```
DenisBiryukov91: Shouldn't these and other fixed sizes be exposed somewhere as macros ?

steils: Done

src/transport/multicast/lease.c
```diff
@@ -148,10 +148,28 @@ void *_zp_multicast_lease_task(void *ztm_arg) {
...
+                    _z_connectivity_peer_disconnected(
+                        s, &disconnected_peer, true, mtu, is_streamed, is_reliable,
+                        _z_string_check(&disconnected_peer._link_src) ? &disconnected_peer._link_src : NULL,
+                        _z_string_check(&disconnected_peer._link_dst) ? &disconnected_peer._link_dst : NULL);
+                    _z_transport_peer_common_clear(&disconnected_peer);
+                    _z_transport_peer_mutex_lock(&ztm->_common);
+                    curr_list = ztm->_peers;
+                    prev = NULL;
+                    prev_drop = NULL;
+                    continue;
```
DenisBiryukov91: it does not look like this continue is needed here

src/transport/multicast/lease.c
```diff
@@ -148,10 +148,28 @@ void *_zp_multicast_lease_task(void *ztm_arg) {
...
                     ztm->_peers = _z_transport_peer_multicast_slist_drop_element(ztm->_peers, prev_drop);
+#if Z_FEATURE_CONNECTIVITY == 1
+                    _z_transport_peer_mutex_unlock(&ztm->_common);
+                    _z_connectivity_peer_disconnected(
+                        s, &disconnected_peer, true, mtu, is_streamed, is_reliable,
+                        _z_string_check(&disconnected_peer._link_src) ? &disconnected_peer._link_src : NULL,
+                        _z_string_check(&disconnected_peer._link_dst) ? &disconnected_peer._link_dst : NULL);
+                    _z_transport_peer_common_clear(&disconnected_peer);
+                    _z_transport_peer_mutex_lock(&ztm->_common);
+                    curr_list = ztm->_peers;
```
DenisBiryukov91: you are resetting a list cursor here, which means that if you had previously peers that received a message, they were just flipped to not received. After cursor reset you are going to check them again and since they were just marked as not received you are going falsely drop them.

I think you would need to gather all your notifications into buffer while holding a mutex, and report them after the loop ends.

Also given that dispatching notifications might block due to the user, all the time management logic here becomes very unreliable (it sort of already was, but not to the same degree), probably it is worth starting to use z_clock_t for more reliable time measurements, not sure if this should be the subject this PR though - it is probably ok to just create an issue ...

steils: Thanks. Fixed

src/transport/multicast/lease.c
```diff
@@ -148,31 +176,22 @@ void *_zp_multicast_lease_task(void *ztm_arg) {
...
                     _z_session_t *s = _z_transport_common_get_session(&ztm->_common);
 #if Z_FEATURE_CONNECTIVITY == 1
-                    _z_transport_peer_common_t disconnected_peer = {0};
-                    uint16_t mtu = 0;
-                    bool is_streamed = false;
-                    bool is_reliable = false;
-                    _z_transport_get_link_properties(&ztm->_common, &mtu, &is_streamed, &is_reliable);
-                    _z_transport_peer_common_copy(&disconnected_peer, &curr_peer->common);
+                    _z_transport_peer_multicast_slist_t *new_dropped_peers =
+                        _z_transport_peer_multicast_slist_push(dropped_peers, curr_peer);
```
DenisBiryukov91: copying peers might be an expensive operation (and it is also broken as of now, since it does not copy resource tree - but this is irrelevant here). Could you rather gather all the peers you are about to drop in a separate list (can probably use _z_slist_extract_filter to split  peers based on their common._received value - this way you do not even need to do any allocations), and then process drop event reporting separately.

src/transport/unicast/lease.c
DenisBiryukov91: Why use a list here, if you only have a single element ?

src/transport/unicast/lease.c
```diff
@@ -171,31 +199,22 @@ void *_zp_unicast_lease_task(void *ztu_arg) {
                     if (drop_peer) {
                         _z_session_t *zs = _z_transport_common_get_session(&ztu->_common);
 #if Z_FEATURE_CONNECTIVITY == 1
-                        _z_transport_peer_common_t disconnected_peer = {0};
-                        uint16_t mtu = 0;
-                        bool is_streamed = false;
-                        bool is_reliable = false;
-                        _z_transport_get_link_properties(&ztu->_common, &mtu, &is_streamed, &is_reliable);
-                        _z_transport_peer_common_copy(&disconnected_peer, &curr_peer->common);
+                        _z_transport_peer_unicast_slist_t *new_dropped_peers =
```
DenisBiryukov91: same comment as for multicast peers

src/transport/unicast/lease.c
steils: Because `ztu->_peers` is always a slist in both client and peer modes.
In client mode it is effectively a single active peer. But in peer mode it can have multiple accepted connections.

src/transport/unicast/lease.c
steils: Done.

src/transport/multicast/rx.c
```diff
@@ -352,8 +421,25 @@ static z_result_t _z_multicast_handle_join_inner(_z_transport_multicast_t *ztm,
         // Check representing capabilities
         if ((msg->_seq_num_res != Z_SN_RESOLUTION) || (msg->_req_id_res != Z_REQ_RESOLUTION) ||
             (msg->_batch_size != Z_BATCH_MULTICAST_SIZE)) {
+#if Z_FEATURE_CONNECTIVITY == 1
+            _z_transport_peer_common_t disconnected_peer = {0};
+            uint16_t mtu = 0;
+            bool is_streamed = false;
+            bool is_reliable = false;
+            _z_transport_get_link_properties(&ztm->_common, &mtu, &is_streamed, &is_reliable);
+            _z_transport_peer_common_copy(&disconnected_peer, &entry->common);
```
DenisBiryukov91: do we really need to make this copy (which might also include copy of the wbuf) to report an event ?

steils: I think you are right. We need only certain fields, not the entire structure. Fixed.

src/api/admin_space.c
```diff
@@ -681,21 +833,82 @@ static void _ze_admin_space_query_handle_pico(const z_loaned_query_t *query, _z_
...
-        _Z_ERROR("Dropped admin space query - session closed");
+#if Z_FEATURE_CONNECTIVITY == 1
+static void _ze_admin_space_query_handle_connectivity_session(const z_loaned_query_t *query, _z_session_t *session,
+                                                              _ze_admin_space_reply_list_t **replies) {
+    _z_session_transport_mutex_lock(session);
+    if (session->_tp._type != _Z_TRANSPORT_UNICAST_TYPE) {
+        _z_session_transport_mutex_unlock(session);
         return;
     }
+#if Z_FEATURE_MULTI_THREAD == 1
```
DenisBiryukov91: Z_FEATURE_MULTI_THREAD guard is not needed here and in other places, since we have a stub version of _z_transport_peer_mutex_lock/unlock for single threaded case

src/api/admin_space.c
```diff
@@ -720,67 +933,531 @@ static void _ze_admin_space_query_handler(z_loaned_query_t *query, void *ctx) {
...
+        session->_admin_space_queryable_id = 0;
+    } else if (ret == _Z_RES_OK) {
+        ret = undeclare_queryable_ret;
+    }
+}
+    return ret;
+#endif
 }
 
 z_result_t zp_stop_admin_space(z_loaned_session_t *zs) {
```
DenisBiryukov91, line 1356: This function and zp_start_admin_space is not thread safe (not that they were before, though), because unprotected reading a modifying non-atomic _admin_space_queryable_id, _admin_space_session_queryable_id, etc at the same time is undefined behavior.

src/api/connectivity.c
```diff
@@ -0,0 +1,874 @@
...
+    }
+
+    _z_connectivity_transport_listener_intmap_clear(&snapshot);
+}
+
+static void _z_connectivity_dispatch_link_event(_z_session_t *session, _z_info_link_event_t *event, bool is_multicast) {
+    _z_connectivity_link_listener_intmap_t snapshot = _z_connectivity_link_listener_intmap_make();
+
+    _z_session_mutex_lock(session);
+    snapshot = _z_connectivity_link_listener_intmap_clone(&session->_connectivity_link_event_listeners);
```
DenisBiryukov91, line 283: I'm wondering, maybe it is better to just put this map under a separate mutex and hold it while dispatching events, given that link events are pretty rare, and thus should not introduce a lot of latency? This would also simplify synchronization of callbacks drop on undeclare.

src/session/utils.c
```diff
@@ -174,17 +215,35 @@ void _z_session_clear(_z_session_t *zn) {
 #if Z_FEATURE_AUTO_RECONNECT == 1
     _z_config_clear(&zn->_config);
 #endif
-    _z_transport_clear(&zn->_tp);
+    _z_session_transport_mutex_lock(zn);
```
DenisBiryukov91, line 234: why not just call _z_transport_clear while holding the mutex ?

src/api/connectivity.c
steils: As discussed, we'll make it a separate task.

src/session/utils.c
steils: Just some optimization. It would unnecessarily extend the critical section over task joins and transport cleanup. Also after the detach, generic `_z_transport_clear()` can no longer be used anyway, since it dispatches by zt->_type.

src/api/admin_space.c
steils: Fixed

src/api/admin_space.c
steils: Fixed

src/session/utils.c
DenisBiryukov91: I do not think we care much about extending critical section here, because we are closing session, so very likely terminating the application - so it is not in the hot path. I believe having a copy-pasted piece of code is way worse here.  Detach issue will need to be addressed separately though.

steils: Ok, let's address the detach issue separately.
Fixed, cleaned up the code.

.github/workflows/integration.yaml
```diff
@@ -76,7 +76,7 @@ jobs:
 
       - name: Test debug
         run: make test
-        timeout-minutes: 30
+        timeout-minutes: 40
```
gmartin82, line 79: While not related to this particular ticket, I think we really need to stop increasing this and fix the underlying issue (at some point, not on this ticket) by building zenohd separately and potentially splitting up the tests.

src/api/admin_space.c
```diff
@@ -152,6 +155,76 @@ static z_result_t _ze_admin_space_pico_transports_0_peers_peer_ke(z_owned_keyexp
...
+                                                       const z_id_t *peer_zid) {
+    z_internal_keyexpr_null(ke);
+
+    z_owned_string_t zid_str;
+    _Z_RETURN_IF_ERR(z_id_to_string(zid, &zid_str));
+    z_owned_string_t peer_zid_str;
+    _Z_CLEAN_RETURN_IF_ERR(z_id_to_string(peer_zid, &peer_zid_str), z_string_drop(z_string_move(&zid_str)));
+
+    char buf[_ZE_ADMIN_SPACE_SESSION_TRANSPORT_KE_BUF_LEN];
+    int n = snprintf(buf, sizeof(buf), "%s%s%.*s%s%s%s%s%s%.*s", _Z_KEYEXPR_AT, _Z_KEYEXPR_SEPARATOR,
```
gmartin82: Why not use `_ze_admin_space_build_ke()`?

src/api/admin_space.c
```diff
@@ -152,6 +155,76 @@ static z_result_t _ze_admin_space_pico_transports_0_peers_peer_ke(z_owned_keyexp
...
+    z_owned_string_t zid_str;
+    _Z_RETURN_IF_ERR(z_id_to_string(zid, &zid_str));
+    z_owned_string_t peer_zid_str;
+    _Z_CLEAN_RETURN_IF_ERR(z_id_to_string(peer_zid, &peer_zid_str), z_string_drop(z_string_move(&zid_str)));
+    z_owned_string_t link_id_str;
+    _Z_CLEAN_RETURN_IF_ERR(z_id_to_string(link_id, &link_id_str), z_string_drop(z_string_move(&zid_str));
+                           z_string_drop(z_string_move(&peer_zid_str)));
+
+    char buf[_ZE_ADMIN_SPACE_SESSION_LINK_KE_BUF_LEN];
+    int n = snprintf(buf, sizeof(buf), "%s%s%.*s%s%s%s%s%s%.*s%s%s%s%.*s", _Z_KEYEXPR_AT, _Z_KEYEXPR_SEPARATOR,
```
gmartin82: Why not use `_ze_admin_space_build_ke()`?

src/api/admin_space.c
```diff
@@ -720,67 +929,534 @@ static void _ze_admin_space_query_handler(z_loaned_query_t *query, void *ctx) {
...
+    if (admin_space_queryable_id != 0) {
+        z_result_t queryable_ret = _ze_admin_space_undeclare_queryable(zs, admin_space_queryable_id);
+        if (queryable_ret == _Z_RES_OK) {
+            session->_admin_space_queryable_id = 0;
+        } else if (ret == _Z_RES_OK) {
+            ret = queryable_ret;
+        }
     }
+
+out:
```
gmartin82, line 1423: I have some concerns over the introduction of using goto's in the code base. This isn't something we normally use and is discouraged by MISRA C.

src/api/admin_space.c
```diff
@@ -720,67 +929,534 @@ static void _ze_admin_space_query_handler(z_loaned_query_t *query, void *ctx) {
...
+err_queryable: {
+    z_result_t undeclare_queryable_ret = _ze_admin_space_undeclare_queryable(zs, session->_admin_space_queryable_id);
+    if (undeclare_queryable_ret == _Z_RES_OK) {
+        session->_admin_space_queryable_id = 0;
+    } else if (ret == _Z_RES_OK) {
+        ret = undeclare_queryable_ret;
+    }
+}
+#endif
+out:
```
gmartin82, line 1351: Same comment re: gotos in code base.

src/api/admin_space.c
```diff
@@ -720,67 +929,534 @@ static void _ze_admin_space_query_handler(z_loaned_query_t *query, void *ctx) {
...
+    }
+
+    ret = _ze_admin_space_session_queryable_ke(&ke, &zid);
+    if (ret != _Z_RES_OK) {
+        _z_session_weak_drop(session_session_weak);
+        z_free(session_session_weak);
+        goto err_queryable;
+    }
+
+    ret = _ze_admin_space_declare_session_queryable(zs, &session->_admin_space_session_queryable_id,
```
gmartin82, line 1252: Why are we declaring two queryables in different ways here?

src/api/admin_space.c
```diff
@@ -310,6 +437,31 @@ static z_result_t _ze_admin_space_add_reply(_z_json_encoder_t *je, const z_loane
     return _Z_RES_OK;
 }
 
+#if Z_FEATURE_CONNECTIVITY == 1
+static z_result_t _ze_admin_space_add_reply_payload(const z_loaned_keyexpr_t *ke, z_moved_bytes_t *payload,
```
gmartin82: Why are we introducing a second mechanism to add a reply here instead of keeping things consistent?

gmartin82 (changes requested):
I have some concerns about style inconsistencies here that should be cleaned up one way or another.

I’m also concerned about the introduction of `goto`, which the existing codebase has generally avoided to date. While `goto` can sometimes make control flow cleaner in C (and I think it does in this case), MISRA C:2023 (e.g. rule 15.1) discourages its use, and I do not think we should start introducing it without an explicit decision on whether we want to allow that pattern going forward. At the moment this feels inconsistent with the current style of the codebase.

.github/workflows/integration.yaml
steils: I agree, but let's work on it in a separate issue.

src/api/admin_space.c
steils: I can use it, but we'll have to increase _ZE_ADMIN_SPACE_KE_BUF_LEN from 128 to 256, so the connectivity keyexprs would fit.

src/api/admin_space.c
steils: It's a usual C-style rollback path, to avoid duplicating lots of code on error paths, and improves readability in general.
MISRA C generally discourages goto, but doesn't ban it. Structured, forward-only jumps to a common cleanup/rollback label in the same function is a usual pattern in C. See MISRA C:2023 15.2 and later, in particular 15.5 (having the single point of exit is why we use goto here).

src/api/admin_space.c
steils: Replied there

src/api/admin_space.c
steils: Because the two queryables have different locality requirements.
The `@/<zid>/pico/**` queryable is a normal admin-space queryable, so z_declare_queryable() is sufficient.
The `@/<zid>/session/**` queryable must be **SessionLocal** according to the RFC [https://github.com/eclipse-zenoh/roadmap/blob/main/rfcs/ALL/Connectivity%20Status%20and%20Events.md]. For that I call `_z_register_queryable(..., Z_LOCALITY_SESSION_LOCAL, ...)` in `_ze_admin_space_declare_session_queryable()`.

src/api/admin_space.c
steils: Done.

src/api/admin_space.c
steils: Done.

src/api/admin_space.c
steils: Thanks. Fixed.

src/api/admin_space.c
gmartin82: My point is you’re the only person using gotos in this code base. It needs wider discussion.

PR #1163, opened 2026-02-20
Improves the robustness of integration test scripts

sashacmc:
## Description
This PR improves the robustness of integration test scripts for local development environments.

### What does this PR do?
 **tests/routed.sh:**
 - Fixes TESTDIR to use absolute path instead of relative path (prevents cp failures)
 - Adds cleanup for corrupted zenohd state from interrupted runs
 - Validates zenohd binary exists and is executable before use
 - Verifies zenohd process started successfully before running tests
 - Checks if process is running before attempting to terminate it

 **CMakeLists.txt:**
 - Replaces configure_file() with file() for test scripts to explicit execute permissions on copied scripts

 The integration test scripts had several issues when running locally

---
## 🏷️ Label-Based Checklist

**Labels:** `internal`

## 🏠 Internal Change

**Instructions:**

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -781,11 +781,9 @@ endif()
       target_link_libraries(z_multi_pubsub_test zenohpico::lib)
       target_link_libraries(z_multi_queryable_test zenohpico::lib)
 
-      file(COPY ${PROJECT_SOURCE_DIR}/tests/routed.sh 
-                ${PROJECT_SOURCE_DIR}/tests/tls_verify.sh 
-                ${PROJECT_SOURCE_DIR}/tests/api.sh
-           DESTINATION ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}
-           FILE_PERMISSIONS OWNER_READ OWNER_WRITE OWNER_EXECUTE GROUP_READ GROUP_EXECUTE WORLD_READ WORLD_EXECUTE)
+      configure_file(${PROJECT_SOURCE_DIR}/tests/routed.sh ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/routed.sh COPYONLY)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -783,11 +780,12 @@ endif()
       target_link_libraries(z_api_admin_space_test zenohpico::lib)
       target_link_libraries(z_multi_pubsub_test zenohpico::lib)
       target_link_libraries(z_multi_queryable_test zenohpico::lib)
-      target_link_libraries(z_api_callback_drop_on_undeclare_test zenohpico::lib)
 
-      configure_file(${PROJECT_SOURCE_DIR}/tests/routed.sh ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/routed.sh COPYONLY)
-      configure_file(${PROJECT_SOURCE_DIR}/tests/tls_verify.sh ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/tls_verify.sh COPYONLY)
-      configure_file(${PROJECT_SOURCE_DIR}/tests/api.sh ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/api.sh COPYONLY)
+      file(COPY ${PROJECT_SOURCE_DIR}/tests/routed.sh
```
steils: I think replacing `configure_file` with `file(COPYONLY ...)` here means that changes to `tests/routed.sh`, `tests/tls_verify.sh`, and `tests/api.sh` are no longer tracked as reconfigure dependencies. So after the initial configure, incremental builds will keep the old script copies in `build/tests`. Maybe keep `configure_file` here? Setting +x is not necessarily required here, because we run these scripts are run by `bash <script>` anyway.
But if we really want it, `configure_file` also supports setting `FILE_PERMISSIONS` (since 3.20, so we'll have to increase `cmake_minimum_required` then).

sashacmc: Unfortunately we can't update Cmake to 3.20 now, so let just revert the CMakeLists.txt changes

PR #1166, opened 2026-02-23
Fix segmentation fault while calling batching operations on disconnected session.

### LINKED ISSUE #1165: [Bug] Segmentation fault when calling batching operations during transport reconnect attempt
### Describe the bug

Calling the `zp_batch_start`, `zp_batch_flush` or `zp_batch_stop` operations while the session is attempting to reconnect the transport after a lease time out results in a segmentation fault.

### To reproduce

1. Open two sessions connected directly to each other - one in peer mode, one in client mode.
2. Disconnect the session in peer mode.
3. Call batching operations in client session.

### System info

- Platform: Ubuntu 24.04
- Zenoh Pico commit: 6a30948

gmartin82:
## Description
Calling the zp_batch_start, zp_batch_flush or zp_batch_stop operations while the session is attempting to reconnect the transport after a lease time out results in a segmentation fault. This PR fixes the issue.

### Related Issues
Fixes #1165.

---
## 🏷️ Label-Based Checklist

**Labels:** `bug`

## 🐛 Bug Fix Requirements

**Instructions:**


src/transport/transport.c
```diff
@@ -86,10 +86,13 @@ void _z_transport_free(_z_transport_t **zt) {
 }
 
 #if Z_FEATURE_BATCHING == 1
-bool _z_transport_start_batching(_z_transport_t *zt) {
+z_result_t _z_transport_start_batching(_z_transport_t *zt) {
     _z_transport_common_t *ztc = _z_transport_get_common(zt);
```
DenisBiryukov91, line 90: the update of ztc->_batch_state is not atomic, nor is protected by any kind of lock, so this check will not prevent problems in all the cases. I believe this issue should be addressed by first making transport api thread safe

steils (approved):

src/transport/transport.c
sashacmc: @DenisBiryukov91, thanks for the feedback, I agree that we should make transport API thread safe, please create an issue for it, but lets merge this PR as is, since it is fix the original problem.

Issue #1120, opened 2025-12-15
[Bug] z_session_drop might not immediately close the session


DenisBiryukov91:
### Describe the bug

z_session_drop might not immediately close the session. 
The issue comes from the fact that session is a strong rc, and internally we occasionally upgrade weak session pointers to strong temporary to perform various operations, if this happens concurrently with z_session_drop, the session will not be closed upon return due to lingering strong session pointers.

### To reproduce

call concurrently z_session_drop with any other operation temporary upgrading session (like z_publisher_put for example).

### System info

Any multi-threaded

DenisBiryukov91:
Fixed in 1.8.0

PR #1168, opened 2026-02-25
Close tls link only after cleaning all peers

DenisBiryukov91:
## Description

Split _z_common_tranport_clear into _z_transport_common_stop_tasks and _z_transport_common_clear to allow separation of stopping accept/lease/listen tasks from clearing common resources - this is needed in particular to be able to clean transport peers list before cleaning the link, since the data in latter (particularly in case of tis connection) can be referenced by the peers.

### What does this PR do?
Split _z_common_tranport_clear into _z_transport_common_stop_tasks and _z_transport_common_clear to allow separation of stopping accept/lease/listen tasks from clearing common resources.

When connecting a new peer via tls link, we call mbedtls_ssl_setup which takes a pointer to data inside tis _z_link_t. According to docs https://sourcevu.sysprogs.com/espressif/lib/mbedtls/symbols/mbedtls_ssl_setup this pointer is being kept until the peer is cleared - which leads to access after free in case if tls link is released before  its peers.

### Related Issues

---
## 🏷️ Label-Based Checklist

**Labels:** `bug`

## 🐛 Bug Fix Requirements

**Instructions:**

### DIFF
--- a/src/transport/common/transport.c
+++ b/src/transport/common/transport.c
@@ -20,7 +20,7 @@
 #include "zenoh-pico/transport/unicast/accept.h"
 #include "zenoh-pico/utils/result.h"
 
-void _z_common_transport_clear(_z_transport_common_t *ztc, bool detach_tasks) {
+void _z_transport_common_stop_tasks(_z_transport_common_t *ztc, bool detach_tasks) {
 #if Z_FEATURE_MULTI_THREAD == 1
     // Clean up tasks
     if (ztc->_read_task != NULL) {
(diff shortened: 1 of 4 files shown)


DenisBiryukov91:
To reliably reproduce the issue we first need to address #1120, so the tests will be added in another PR.

src/transport/common/transport.c
```diff
@@ -56,7 +56,5 @@ void _z_common_transport_clear(_z_transport_common_t *ztc, bool detach_tasks) {
     // Clean up the buffers
     _z_wbuf_clear(&ztc->_wbuf);
     _z_zbuf_clear(&ztc->_zbuf);
-
-    _z_link_free(&ztc->_link);
```
gmartin82, line 60: I understand why the change is necessary but I do have one concern.

Previously, `_z_common_transport_clear()` cleared all common-owned resources, including `_link`. After this change, it no longer does — which means `clear()` no longer fully clears the common state. This creates a hidden ownership split and relies on every caller remembering to free `_link` separately. That’s fragile and, while not a problem at the moment, could lead to issues later as the code base evolves.

PR #1173, opened 2026-03-06
Modify the admin space to return rolled up session information.

gmartin82:
## Description
This PR updates the admin space session replies so that higher-level endpoints return rolled-up session information, rather than requiring clients to reconstruct it from multiple lower-level replies.

In particular, the `@/<zid>/pico` and related session endpoints now expose session, transport, and peer information in a more useful aggregated form, while still keeping the more specific endpoints available.

### What does this PR do?
- Rolls up session information in the admin space so higher-level queries return more complete data
- Keeps the endpoint hierarchy under `@/<zid>/pico/**` consistent:
  - `@/<zid>/pico`
  - `@/<zid>/pico/session`
  - `@/<zid>/pico/session/transports`
  - `@/<zid>/pico/session/transports/0`
  - `@/<zid>/pico/session/transports/0/peers`
  - `@/<zid>/pico/session/transports/0/peers/<peer_zid>`
- Makes admin-space query handling more robust by continuing to process other endpoints even if building or handling one endpoint fails
- Updates the admin-space tests to cover:
  - subtree queries for `@/<zid>/pico/**`
  - exact queries for each fixed endpoint
  - combined coverage for individual peer endpoints

Previously, clients querying the admin space needed to rely more heavily on multiple lower-level replies to reconstruct a useful picture of the session state.

This change makes the admin space easier to consume by returning richer information at the higher-level endpoints, while preserving the more specific endpoints for callers that want finer-grained queries.

It also improves resilience of admin-space handling by making endpoint processing best-effort rather than aborting the whole query when one endpoint fails.

### Related Issues
https://github.com/eclipse-zenoh/zenoh-pico/pull/1125

---
## 🏷️ Label-Based Checklist

**Labels:** `enhancement`

## ✨ Enhancement Requirements

**Instructions:**

### DIFF
--- a/src/api/admin_space.c
+++ b/src/api/admin_space.c
@@ -675,9 +675,8 @@ static const _ze_admin_space_endpoint_t _ze_admin_space_pico_endpoints[] = {
 
 static void _ze_admin_space_query_handle_pico(const z_loaned_query_t *query, _z_session_t *session,
                                               _ze_admin_space_reply_list_t **replies) {
-    _ze_admin_space_query_handle_endpoints(
-        query, session, _ze_admin_space_pico_endpoints,
-        sizeof(_ze_admin_space_pico_endpoints) / sizeof(_ze_admin_space_pico_endpoints[0]), replies);
+    _ze_admin_space_query_handle_endpoints(query, session, _ze_admin_space_pico_endpoints,
+                                           _ZP_ARRAY_SIZE(_ze_admin_space_pico_endpoints), replies);
 
     _ze_admin_space_query_handle_pico_transport_0_peers(query, session, replies);
 }


src/api/admin_space.c
```diff
@@ -22,75 +22,125 @@
...
+    for (size_t i = 0; i < segment_count; i++) {
+        n = snprintf(buf + off, sizeof(buf) - off, "%s%.*s", _Z_KEYEXPR_SEPARATOR, (int)segments[i].len,
+                     segments[i].data);
+        if (n < 0 || (size_t)n >= sizeof(buf) - off) {
+            return _Z_ERR_INVALID;
+        }
+        off += (size_t)n;
+    }
 
     return z_keyexpr_from_str(ke, buf);
```
DenisBiryukov91: this probably should be from z_keyexpr_from_substr with specified length (which is faster), or buf must be explicitly zero initialized, unless I missed the place where null-terminator is written.

src/api/admin_space.c
```diff
@@ -265,209 +301,418 @@ static z_result_t _ze_admin_space_add_reply(_z_json_encoder_t *je, const z_loane
...
+
         default:
             break;
     }
+}
 
-    _z_session_mutex_unlock(session);
+static void _ze_admin_space_query_handle_pico(const z_loaned_query_t *query, _z_session_t *session,
+                                              _ze_admin_space_reply_list_t **replies) {
+    z_owned_keyexpr_t ke;
```
DenisBiryukov91, line 644: Could we rather define an array of {admin_space_getter, encode_function, err_message?) and just iterate over it ?

src/api/admin_space.c
gmartin82: snprintf writes a null terminator, but I've switched the implementation to `z_keyexpr_from_substr` to avoid having to calculate the string length again when it's already known.

src/api/admin_space.c
gmartin82: Done. Note I've kept the transport peers separately as this is handled slightly differently.

src/api/admin_space.c
```diff
@@ -632,70 +639,45 @@ static void _ze_admin_space_query_handle_pico_transport_0_peers(const z_loaned_q
...
+    {"pico/session/transports/0", _ze_admin_space_pico_transports_0_ke, _ze_admin_space_encode_pico_transport_0},
+    {"pico/session/transports/0/peers", _ze_admin_space_pico_transports_0_peers_ke,
+     _ze_admin_space_encode_pico_transport_0_peers},
+};
+
+static void _ze_admin_space_query_handle_pico(const z_loaned_query_t *query, _z_session_t *session,
+                                              _ze_admin_space_reply_list_t **replies) {
+    _ze_admin_space_query_handle_endpoints(
+        query, session, _ze_admin_space_pico_endpoints,
+        sizeof(_ze_admin_space_pico_endpoints) / sizeof(_ze_admin_space_pico_endpoints[0]), replies);
```
DenisBiryukov91: we have a _ZP_ARRAY_SIZE macro that does exactly this

DenisBiryukov91:
LGTM, besides a minor nitpick regarding array size calculation.

src/api/admin_space.c
gmartin82: Changed thanks.

PR #1175, opened 2026-03-07
Faster key expression matching

DenisBiryukov91:
## Description

Implement faster key expression intersection and inclusion.

### What does this PR do?

Improves performance of key expression matching. Now intersection/inclusion complexity is linear in (ke1_len + ke2_len) if each of key expressions contain no more than one superwild `**` chunk, and each chunk contains no more than one stardsl `$*`, which should cover majority of use-cases. If these conditions are violated the complexity might increase up to `k1_len * ke2_len`. This is still a significant improvement to previous `ke1_len * ke2_len` complexity in case of matching kes and potentially `ke1_len^max(1, n1) * ke2_len^max(1, n2)` in case of non-matching kes, where n1, n2 are numbers of superwild `**` chunks  or (stardsl`$*` characters  if estimating chunk match complexity). 
In addition whenever possible the ke expressions are parsed only once (sometimes even only partially), which also improves performance even for kes, without `**` or `$*`.

The new matching algorithm also changes behavior of matching verbatim chunks containing `$*`, to actually allow it to match any characters, i.e. `@abc` now intersects with `@a$*c`. This is not relevant since it was decided to forbid using `$*` in verbatim chunks (see #1177).

To improve performance of key expression matching.

### Related Issues

---
## 🏷️ Label-Based Checklist

**Labels:** `enhancement`

## ✨ Enhancement Requirements

**Instructions:**


tests/z_keyexpr_test.c
```diff
@@ -129,7 +129,7 @@ void test_intersects(void) {
     TEST_FALSE_INTERSECT("@a", "@a/b")
     TEST_FALSE_INTERSECT("@a", "@a/*")
     TEST_FALSE_INTERSECT("@a", "@a/*/**")
-    TEST_FALSE_INTERSECT("@a", "@a$*/**")
+    // TEST_FALSE_INTERSECT("@a", "@a$*/**")
```
gmartin82: What is the reason for these tests being commented out?

include/zenoh-pico/session/keyexpr_match_template.h
```diff
@@ -0,0 +1,532 @@
...
+        result.lend = _z_chunk_end(lbegin, lkend);
+        result.rend = rbegin + 1;
+        return result;
+    }
+    // at this stage we should only care about stardsl, as the presence of verbatim or wild is already checked.
+    while (lbegin < lkend && rbegin < rkend && *lbegin != _Z_DELIMITER && *rbegin != _Z_DELIMITER) {
+        if (*lbegin == _Z_DSL0) {
+            return _ZP_CAT(_z_chunk_forward_backward, _ZP_KE_MATCH_OP)(lbegin + _Z_DSL_LEN, lkend, rbegin, rkend);
+        } else if (_ZP_KE_MATCH_TYPE_INTERSECTS && *rbegin == _Z_DSL0) {
+            _z_chunk_forward_match_data_t res =
```
steils, line 240: Here the operands are swapped to reuse `_z_chunk_forward_backward_intersects()`, but the returned lend/rend are copied back without swapping them. As a result it breaks `intersects()` symmetry.
E.g.:
```
intersects("a", "$*a") == false, while intersects("$*a", "a") == true;
```
It should be symmetric, right?

include/zenoh-pico/session/keyexpr_match_template.h
```diff
@@ -0,0 +1,532 @@
...
+                                : _Z_CHUNK_MATCH_RESULT_NO;
+            return result;
+        } else if (*lbegin != *rbegin) {
+            return result;
+        }
+        lbegin++;
+        rbegin++;
+    }
+
+    if (lbegin == lend) {  // remainder of left is entirely matched, suffix stardsl can match the rest of right
```
steils, line 300: This path lets a non-verbatim `$*...` suffix after `**` match a verbatim chunk when reached through backward matching. That breaks the existing rule that wildcard matching must not consume `@...`.
`includes("a/**/$*b", "a/@b") == true` and `intersects("a/**/$*b", "a/@b") == true`, while upstream returns false for both.

include/zenoh-pico/session/keyexpr_match_template.h
```diff
@@ -0,0 +1,532 @@
...
+    while (lbegin < lend && rbegin < rend) {
+        _z_chunk_forward_match_data_t res = _ZP_CAT(_z_chunk_forward, _ZP_KE_MATCH_OP)(lbegin, lend, rbegin, rend);
+        if (res.result == _Z_CHUNK_MATCH_RESULT_YES) {
+            lbegin = res.lend + _Z_DELIMITER_LEN;
+            rbegin = res.rend + _Z_DELIMITER_LEN;
+        } else if (res.result == _Z_CHUNK_MATCH_RESULT_LEFT_SUPERWILD) {
+            if (lbegin + _Z_DOUBLE_STAR_LEN == lend) {  // left is exhausted
+                return _z_keyexpr_get_next_verbatim_chunk(rbegin, rend) == NULL;
+            }
+            return _ZP_CAT(_z_keyexpr_backward, _ZP_KE_MATCH_OP)(lbegin + _Z_DOUBLE_STAR_LEN + _Z_DELIMITER_LEN, lend,
```
steils, line 483: This sends `a/**/$*b` into the backward matcher above, which introduces the verbatim regression.

include/zenoh-pico/session/keyexpr_match_template.h
```diff
@@ -0,0 +1,532 @@
...
+    while (lbegin < lend && rbegin < rend) {
+        const char *lverbatim = _z_keyexpr_get_next_verbatim_chunk(lbegin, lend);
+        const char *rverbatim = _z_keyexpr_get_next_verbatim_chunk(rbegin, rend);
+        if (lverbatim == NULL && rverbatim == NULL) {
+            return _ZP_CAT(_z_keyexpr_forward, _ZP_KE_MATCH_OP)(lbegin, lend, rbegin, rend, false);
+        } else if (lverbatim == NULL || rverbatim == NULL) {  // different number of verbatim chunks, kes can not match
+            return false;
+        }
+
+        if (!_ZP_CAT(_z_keyexpr_forward, _ZP_KE_MATCH_OP)(lbegin, lverbatim - _Z_DELIMITER_LEN, rbegin,
```
steils: If the key expression starts with a verbatim chunk, `lverbatim == lbegin` here, so `lverbatim - 1` makes a pointer one byte before the object. Same for `rverbatim - 1`.

tests/z_keyexpr_test.c
```diff
@@ -189,7 +189,7 @@ void test_includes(void) {
     TEST_FALSE_INCLUDE("@a", "@a/b")
     TEST_FALSE_INCLUDE("@a", "@a/*")
     TEST_FALSE_INCLUDE("@a", "@a/*/**")
-    TEST_FALSE_INCLUDE("@a$*/**", "@a")
+    // TEST_FALSE_INCLUDE("@a$*/**", "@a")
```
steils: Same comment as for :132

steils (changes requested):

tests/z_keyexpr_test.c
DenisBiryukov91: Because, new implementation considers them as matching, while old one does not. The consensus was to simply forbid using `$*` in verbatim chunks, to avoid any user confusion in a separate PR, so this test is no longer relevant.

gmartin82: Should we just remove them then?

include/zenoh-pico/session/keyexpr_match_template.h
DenisBiryukov91: This is totally fine, since end pointer is never dereferenced and only compared with begin. In this case _z_keyexpr_forward will only return true if another sub ke prior to `@chunk` is also empty or double a double star.

include/zenoh-pico/session/keyexpr_match_template.h
DenisBiryukov91: fixed

include/zenoh-pico/session/keyexpr_match_template.h
DenisBiryukov91: fixed

include/zenoh-pico/session/keyexpr_match_template.h
DenisBiryukov91: explicitly made sub ke pointer not go below its begin, just in case

tests/z_keyexpr_test.c
DenisBiryukov91: Test removed as no longer relevant

tests/z_keyexpr_test.c
DenisBiryukov91: Removed

include/zenoh-pico/session/keyexpr_match_template.h
DenisBiryukov91: fixed backward matcher

gmartin82 (approved):

steils (approved):
I would add a test scenario for "verbatim + $*", like `intersects("@abc", "@a$*c") == true` and `includes("@a$*c", "@abc") == true`.
Otherwise LGTM.

sashacmc (approved):

PR #1187, opened 2026-03-16
Improve connection establishment.

gmartin82:
## Description

This PR updates `z_open()` connection-establishment behaviour so configured listen/connect locators can be retried instead of being attempted only once.

It adds independent retry timeout and failure-policy configuration for listen and connect locators, and clarifies peer-mode behaviour when a session can be opened with only partial peer connectivity.

### What does this PR do?

- Adds unstable connect configuration options:
  - `Z_CONFIG_CONNECT_TIMEOUT_KEY`
  - `Z_CONFIG_CONNECT_EXIT_ON_FAILURE_KEY`

- Adds unstable listen configuration options:
  - `Z_CONFIG_LISTEN_TIMEOUT_KEY`
  - `Z_CONFIG_LISTEN_EXIT_ON_FAILURE_KEY`

- The new connect/listen timeout and exit-on-failure configuration keys are unstable and are only exposed when `Z_FEATURE_UNSTABLE_API` is enabled. The corresponding default string constants remain available so the implementation can keep default behaviour consistent when the unstable keys are not enabled.

- Timeout values support:
  - `0`: no retry; attempt once.
  - `>0`: retry retryable failures until the timeout expires.
  - `-1`: retry indefinitely.

- Default behaviour:
  - Connect timeout defaults to `0`.
  - Listen timeout defaults to `0`.
  - Client connect exit-on-failure defaults to `true`.
  - Peer connect exit-on-failure defaults to `false`.
  - Listen exit-on-failure defaults to `true`.

- Client mode:
  - Only connect locators are used.
  - Configured connect locators are treated as alternatives.
  - `z_open()` succeeds when one connect locator succeeds.
  - `z_open()` fails if no connect locator can establish a transport.
  - `connect_exit_on_failure` has no effect in client mode, as a session cannot be opened without a transport.

- Peer mode:
  - `z_open()` requires a primary transport before returning successfully.
  - The primary transport can be established either by:
    - opening the configured listen locator, or
    - connecting to one configured connect locator.
  - If a listen locator is configured, it is attempted first.
  - If listening fails and `listen_exit_on_failure=false`, `z_open()` may still succeed by connecting to a configured connect locator.
  - If no listen or connect locator establishes a primary transport, `z_open()` fails.
  - Once the primary transport is established, remaining connect locators are added as additional peers.
  - If `connect_exit_on_failure=true`, `z_open()` fails if all required peer connections cannot be established within the configured timeout.
  - If `connect_exit_on_failure=false`, `z_open()` may return successfully with partial connectivity and leave remaining retryable peer connections to be attempted:
    - synchronously (single-thread mode, when session tasks are progressed), or
    - asynchronously (multi-thread mode, via background transport tasks).

- Retry semantics:
  - Only retryable transport errors are retried.
  - Non-retryable errors permanently exclude a locator from further attempts.
  - Retryable failures use exponential backoff until timeout expires.

- Adds `_Z_ERR_TRANSPORT_OPEN_PARTIAL_CONNECTIVITY`, returned when:
  - a primary transport has been established, but
  - required peer connectivity could not be completed under a strict (`exit_on_failure=true`) policy.

- Adds tests covering:
  - default open behaviour without unstable config keys
  - connect timeout when no endpoint is available (`Z_FEATURE_UNSTABLE_API`)
  - partial peer connectivity (`Z_FEATURE_UNSTABLE_API`)
  - late-joining endpoints becoming available within the timeout (`Z_FEATURE_UNSTABLE_API`)
  - behaviour differences between strict and best-effort failure policies (`Z_FEATURE_UNSTABLE_API`)

Previously, configured listen/connect locators were attempted once during `z_open()`. This made session establishment sensitive to startup ordering: if a peer was not yet listening, or a local listen endpoint was temporarily unavailable, `z_open()` could fail immediately.

This change makes connection establishment more robust by allowing applications to choose between:
- existing no-retry behaviour,
- bounded retry,
- infinite retry,
- strict failure behaviour, and
- best-effort peer connectivity.

### Related Issues

ZEN-838

---
## 🏷️ Label-Based Checklist

**Labels:** `enhancement`

## ✨ Enhancement Requirements

**Instructions:**


DenisBiryukov91:
I believe peer connection is better to be done as asynchronous background task based on https://github.com/eclipse-zenoh/zenoh-pico/pull/1190.

include/zenoh-pico/api/types.h
```diff
@@ -184,6 +213,8 @@ typedef struct {
  *     feature is enabled).
  */
 typedef struct {
+    uint32_t connect_timeout_ms;
```
DenisBiryukov91: Should this rather be a part of config as it is for Zenoh-rust ?
Should not we keep these options as unstable for the time being ?

src/net/session.c
```diff
@@ -176,58 +177,201 @@ static z_result_t _z_open_inner(_z_session_rc_t *zs, _z_string_t *locator, const
...
+    z_result_t ret = _Z_ERR_TRANSPORT_OPEN_FAILED;
+    z_clock_t now = z_clock_now();
+    uint32_t sleep_ms = _Z_OPEN_BACKOFF_MIN_MS;
+    size_t opened_connect_idx = SIZE_MAX;
+
+    /* Open a single primary locator.
+     * If listen locators are configured, one of them is used as the primary open.
+     * Otherwise one connect locator is opened with retry/backoff.
+     */
+    while (ret != _Z_RES_OK) {
```
steils: This retry loop currently treats every ret != _Z_RES_OK as non-fatal and retriable. But _z_open_inner() can also return non-retriable errors such as invalid locator/schema or protocol/handshake failures. I think we should not retry fatal errors.

src/net/session.c
```diff
@@ -176,58 +177,201 @@ static z_result_t _z_open_inner(_z_session_rc_t *zs, _z_string_t *locator, const
...
+        }
+    }
+
+    if (pending_mask == 0u) {
+        return _Z_RES_OK;
+    }
+
+    uint32_t sleep_ms = _Z_OPEN_BACKOFF_MIN_MS;
+
+    while (pending_mask != 0u) {
```
steils: Same as for :293. _z_new_peer() failures are always retried until timeout. Should we distinguish fatal and non-fatal errors in this loop?

steils (changes requested):

sashacmc:
@DenisBiryukov91 can we close this PR, or should it be updated after the #1190

gmartin82:
@sashacmc It needs to be rebased and updated to add background connection establishment.

src/net/session.c
DenisBiryukov91: I believe locators should be validated in advance upon initialization in the config. 
Everything else should probably be considered as retriable

src/net/session.c
```diff
@@ -154,58 +150,382 @@ static z_result_t _z_open_inner(_z_session_rc_t *zs, _z_string_t *locator, const
...
+z_result_t _z_open_locators_client(_z_session_rc_t *zn, const _z_string_svec_t *connect_locators, const _z_id_t *zid,
+                                   _z_config_t *config, int32_t timeout_ms) {
+    size_t connect_len = _z_string_svec_len(connect_locators);
+
+    if (connect_len == 0) {
+        _Z_ERROR("No connect locators configured in client mode");
+        return _Z_ERR_CONFIG_LOCATOR_INVALID;
+    }
+
+    // This implementation uses a bitmask to track which connect locators remain retryable.
```
DenisBiryukov91: I believe if we ever receive a non-retriable error (i.e. tls certificate issue - I can not think of anything else, if locators are validated in advance when they are inserted in the config) we should just terminate with error and thus using mask seems to be unnecessary

gmartin82: The ticket is clearly scoped to improving the connection establishment not the config.

Please raise a separate issue if you want locators validated on insertion as this is not in scope.

DenisBiryukov91: Still this does not change the fact that we do not need to trace retriability of locators - if a error is not retriable we just abort everything - since the provided config is broken.

gmartin82: Validating locators at config insertion time is a separate behavioral change and is outside the scope of this PR. This PR is about improving connection establishment resilience for configured locators.

The retryability handling was added specifically to address the earlier review comment that the loop should not retry fatal open errors indefinitely. The mask is how we implement that per locator: retryable failures remain pending, non-retryable failures are removed from the pending set, and successful locators are removed as completed.

The set of retryable transport errors used here is aligned with the existing `_z_client_reopen_task_fn()` behavior. Whether a non-retryable locator failure should cause `z_open()` itself to fail is governed by the existing `exit_on_failure` options; when that option is false, one failed locator should not make every configured locator mandatory.

DenisBiryukov91: exit_on_failure should only be applied to retrievable failures (as it is the case zenoh-rust), it has nothing to do with non-retriable ones, which makes retry redundant.

docs/config.rst
```diff
@@ -41,13 +41,61 @@ Connect
...
+
+`Z_CONFIG_LISTEN_TIMEOUT_KEY` accepts the following values:
+
+* `0`: no retry, try opening the listen locator once.
+* `>0`: retry retryable listen failures until the timeout expires.
+* `-1`: retry indefinitely.
+
+The default listen timeout is `0`.
+
+`Z_CONFIG_LISTEN_EXIT_ON_FAILURE_KEY` accepts `true` or `false`.
```
sashacmc, line 108: Please mark new config parameters as unstable (add a comment), since I'm not sure that they will be maintained in the near future in the context of the upcoming changes in P2P

src/net/session.c
```diff
@@ -154,58 +150,382 @@ static z_result_t _z_open_inner(_z_session_rc_t *zs, _z_string_t *locator, const
...
+    if (connect_len > (sizeof(uint64_t) * 8u)) {
```
sashacmc: What does `8u` means?
Please avoid any "magic numbers" in the code.

src/net/session.c
```diff
@@ -154,58 +150,382 @@ static z_result_t _z_open_inner(_z_session_rc_t *zs, _z_string_t *locator, const
...
+        _Z_ERROR("Too many connect locators configured");
```
sashacmc: Do we have a limit on the number of locators? What is it? Where is it configured?

src/net/session.c
```diff
@@ -154,58 +150,382 @@ static z_result_t _z_open_inner(_z_session_rc_t *zs, _z_string_t *locator, const
...
-        // Add other locators as peers if applicable
-        if ((ret == _Z_RES_OK) && (mode == Z_WHATAMI_PEER)) {
-            for (size_t i = 1; i < len; i++) {
-                // Add peer
-                locator = _z_string_svec_get(&locators, i);
-                ret = _z_new_peer(&_Z_RC_IN_VAL(zn)->_tp, &_Z_RC_IN_VAL(zn)->_local_zid, locator, config);
-                if (ret != _Z_RES_OK) {
-                    break;
-                }
+    if (retry_mask != 0u) {
```
sashacmc: Frankly, the logic with the retry mask looks strange. In what case should we stop attempting to connect? If such a case arises, it shouldn't be hidden, but disclosed to the user. If this can be determined during the locator check, then it should be done there.

sashacmc (changes requested):

include/zenoh-pico/api/types.h
gmartin82: Moved to be part of the config and marked as unstable.

src/net/session.c
gmartin82: As requested, the implementation only retries locators that are deemed to be retryable. The set of error codes considered retryable in `_z_client_reopen_task_fn` was used as a basis for this.

Note: Zenoh retries all locators without considering returned errors.

src/net/session.c
gmartin82: Done.

src/net/session.c
gmartin82: When exit_on_failure is set, Zenoh exits on any error. The implementation matches this behaviour.

docs/config.rst
gmartin82: Done.

src/net/session.c
gmartin82: Use of magic numbers removed.

src/net/session.c
gmartin82: Changed to use a dynamicly sized svec of pending peers to avoid a limit.

src/net/session.c
gmartin82: The implementation has been changed to use a new data structure to track pending peers.

The validation comment is not relevant to this discussion and is not part of the scope of the ticket. Locators are currently validated by lower-level code which I haven't changed. If you wish validation to be addressed differently, a separate ticket should be raised for the issue.

sashacmc (approved):

PR #1188, opened 2026-03-17
Redesign platform integration

steils:
The previous platform model was too coarse. A platform was described by
platform-specific system/network files. Adding a new platform often meant
copying or changing those files even when the needed TCP, UDP, BT or
serial code was already available elsewhere.

This mixed platform-specific code with transport selection and made reuse
difficult.

Describe each platform with one small CMake file.

That file lists the platform-specific source files and may also define
compile definitions, include directories, compile options, link
libraries, a system-layer name override, and a platform header override
for custom platform-specific types.

A platform profile may use Zenoh-Pico source files directly or link imported
CMake targets from an external package.

Reorganize the source tree around four transport types:

  - tcp
  - udp
  - bt
  - serial

Move shared transport code out of platform-specific system/network files
into common and upper layers.

As part of this redesign:

  - add built-in platform profiles under `cmake/platforms`
  - make `ZP_PLATFORM_SOURCE_FILES` the only public platform source list
  - let external platform packages be added through
    `find_package(... CONFIG REQUIRED)` via `ZP_EXTERNAL_PACKAGES`
  - let external platform profiles link installed package targets through
    `ZP_PLATFORM_LINK_LIBRARIES`
  - select platform-specific transport source files explicitly instead of
    pulling them in through broad `src/link` globs
  - move UDP multicast out of `src/system/*/network.*` into the UDP
    transport path
  - move WebSocket support to the upper layer, with a generic
    byte-stream path and an Emscripten-specific override
  - move endpoint and address helpers into `src/link/transport/common`
  - update the platform documentation
  - add one-file in-tree and out-of-tree examples

The result is a simpler platform model: a new platform is usually one
CMake file that lists its sources and optional linked package targets,
and external platform packages can be added through standard CMake
packages without editing a central CMake file.

---
## 🏷️ Label-Based Checklist

**Labels:** `platfoms`, `internal`

## 🏠 Internal Change

**Instructions:**


include/zenoh-pico/link/transport/lwip_socket.h
```diff
@@ -0,0 +1,26 @@
+#ifndef ZENOH_PICO_LINK_BACKEND_LWIP_SOCKET_H
+#define ZENOH_PICO_LINK_BACKEND_LWIP_SOCKET_H
+
+#include "zenoh-pico/system/platform.h"
+
```
DenisBiryukov91, line 19: No #ifdef __cplusplus guards ?

steils: Why would it be necessary here?

DenisBiryukov91: It might not be required as of now, given that the file only contains definitions of 2 inline functions, but missing this guard might create problems later when/if some extra struct definitions get added later.

steils: Fixed

include/zenoh-pico/link/backend/datagram.h
```diff
@@ -0,0 +1,82 @@
+#ifndef ZENOH_PICO_LINK_BACKEND_DATAGRAM_H
```
DenisBiryukov91: All headers should contain a ZettaScale copyright header

src/system/arduino/esp32/network.cpp
DenisBiryukov91: Should these functions also go to backends ?

docs/platforms.rst
```diff
@@ -0,0 +1,215 @@
...
+   cmake -S . -B build \
+     -DZP_PLATFORM=linux \
+     -DZP_STREAM_BACKEND=tcp_posix \
+     -DZP_DATAGRAM_BACKEND=udp_posix \
+     -DZP_RAWIO_BACKEND=tty_posix
+
+Adding a new platform
+=====================
+
+There are two ways to add a platform descriptor:
```
DenisBiryukov91: Why not define a platform simply as a cmake subproject, which will be consumed by zenoh-pico via find_package for example  ?

src/link/backend/rawio/uart_mbed.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+static size_t _z_uart_mbed_write(_z_sys_net_socket_t sock, const uint8_t *ptr, size_t len) {
+    ssize_t wb = sock._serial->write(ptr, len);
+    if (wb <= 0) {
+        return SIZE_MAX;
+    }
+    return (size_t)wb;
+}
+
+extern const _z_rawio_ops_t _z_uart_mbed_rawio_ops = {
```
DenisBiryukov91: This is not a valid C++ code  as of C++17, see https://en.cppreference.com/w/cpp/language/aggregate_initialization.html.
I think this also should be addressed in other .c files, to allow compiling zenoh-pico with a c++ compiler.

src/system/arduino/esp32/network.cpp
steils: Rather not, at least not to the stream/datagram/rawio backends.
These functions work on the platform socket type (`_z_sys_net_socket_t`) and are used directly by the generic transport code. They are still closer to the system layer than to a transport backend. If we ever move them, I think it would probably move to a separate socket/system layer rather than into the current backend layer.

docs/platforms.rst
steils: Rather not.
A platform in this redesign is not meant to be an independently buildable and installable library. But rather a small subsystem of the single zenoh-pico library, selected by the build system. Using `find_package()` would make platforms and the building process much heavier than needed.

include/zenoh-pico/link/backend/datagram.h
steils: Thanks, fixed.

src/link/backend/rawio/uart_mbed.cpp
steils: Thanks.
Fixed in .cpp files.
But .c is compiled as C, and it's a valid code in C. Supporting the compilation of C code with a C++ compiler is completely another task.

DenisBiryukov91:
Regarding terminology. Why use term "backend" which is generally reserved for type of storage in zenoh-rust, instead of "link". Also why "streaming"/"datagram" instead of directly using "tcp"/"udp" ?
Also why call backend/platform "default" - if in reality we only have a single platform and a single backend ?

docs/platforms.rst
```diff
@@ -0,0 +1,215 @@
...
+The platform name is the descriptor file name without the ``.cmake`` suffix. For example,
+``cmake/platforms/myrtos.cmake`` defines the ``myrtos`` platform profile.
+
+In the common built-in case, adding a new platform only requires adding a new file under ``cmake/platforms/``.
+
+That file should:
+
+* declare the platform-specific compile definitions and sources;
+* choose the default ``stream``, ``datagram`` and ``rawio`` backends for the platform.
+* optionally add include directories or adjust ``CHECK_THREADS`` if the platform needs it.
```
DenisBiryukov91: as far as I understand extra include directories will not propagate on install (since we currently only take ${CMAKE_CURRENT_SOURCE_DIR}/include/zenoh-pico), which will lead to a broken installation

include/zenoh-pico/link/backend/default_ops.h
```diff
@@ -0,0 +1,67 @@
...
+#include "zenoh-pico/link/backend/datagram.h"
+#include "zenoh-pico/link/backend/rawio.h"
+#include "zenoh-pico/link/backend/stream.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#if (Z_FEATURE_LINK_TCP == 1 || Z_FEATURE_LINK_TLS == 1 || Z_FEATURE_LINK_WS == 1) && defined(ZP_DEFAULT_STREAM_OPS)
+extern const _z_stream_ops_t ZP_DEFAULT_STREAM_OPS;
```
DenisBiryukov91: I think it would be better to configure this via configure_file, instead of adding as compile definition - will be way easier to understand which backend is being used after build/install and also simplify debug in case of problems.

steils:
@DenisBiryukov91 

> Regarding terminology. Why use term "backend" which is generally reserved for type of storage in zenoh-rust, instead of "link".

Because "link" is already used for an abstraction that a user sees, a protocol like tcp, udp, serial, tls, ws. While tcp_posix, udp_lwip, tty_posix etc are not "links", but a platform-specific implementations of links, i.e. lower level.
We can replace "backend" with something else, but I couldn't think of anything better.

> Also why "streaming"/"datagram" instead of directly using "tcp"/"udp" ?

Because these are protocol families, not the protocols themselves. WS also goes to "streams", while "datagram" family has UDP unicast and UDP multicast which have different implementations. And in theory there could be more custom transport protocols.

> Also why call backend/platform "default" - if in reality we only have a single platform and a single backend ?

In a certain build — yes, only one. But a platform cmake file selects a default backend for this platform, while a user can override it. I made an example in the docs for it like:
```
   cmake -S . -B build \
     -DZP_EXTERNAL_CMAKE=/path/to/myrtos/zenoh_pico.cmake \
     -DZP_PLATFORM=linux \
     -DZP_RAWIO_BACKEND=uart_myrtos
```

DenisBiryukov91:
> Because these are protocol families, not the protocols themselves. WS also goes to "streams", while "datagram" family has UDP unicast and UDP multicast which have different implementations. And in theory there could be more custom transport protocols.

Does this imply that we cannot have both multicast/unicast UDP, or ws/tcp in the same build ?

DenisBiryukov91:
> In a certain build — yes, only one. But a platform cmake file selects a default backend for this platform, while a user can override it. I made an example in the docs for it like:

Why user needs to overwrite if he can simply provide a different system descriptor ? Why introduce multiple ways to do the same thing ?

steils:
> > Because these are protocol families, not the protocols themselves. WS also goes to "streams", while "datagram" family has UDP unicast and UDP multicast which have different implementations. And in theory there could be more custom transport protocols.
> 
> Does this imply that we cannot have both multicast/unicast UDP, or ws/tcp in the same build ?

A platform cmake file currently selects only one default backend per type (stream, datagram, rawio). But that does not mean only one protocol of that type can be enabled in the build. Link features are still controlled separately by `Z_FEATURE_LINK_*`. UDP unicast and UDP multicast can still be enabled together, as well as TCP and TLS, using cmake flags.
Maybe this looks not quite symmetrical and needs to be refined.

steils:
> > In a certain build — yes, only one. But a platform cmake file selects a default backend for this platform, while a user can override it. I made an example in the docs for it like:
> 
> Why user needs to overwrite if he can simply provide a different system descriptor ? Why introduce multiple ways to do the same thing ?

Because they are not exactly the same thing.
A platform descriptor defines a whole named build profile. A backend override only changes one default inside that profile. If a combination is meant to be a stable reusable profile, then yes, a separate platform descriptor makes sense. But if only one backend needs to be changed for experiments, CI, out-of-tree integrations, etc, overriding a single backend is simpler than copy-pasting the whole platform descriptor just to change one default.

docs/platforms.rst
steils: Fixed. Made the extra include directories private.

DenisBiryukov91: This does not fix the problem, since for example task/clock/mutex etc platform primitive definitions from user-provided headers are required to be included in install, since they define the layout of corresponding public types such as z_clock_t, z_mutex_t, z_task_t, etc

docs/platforms.rst
```diff
@@ -0,0 +1,218 @@
...
+``cmake/platforms/myrtos.cmake`` defines the ``myrtos`` platform profile.
+
+In the common built-in case, adding a new platform only requires adding a new file under ``cmake/platforms/``.
+
+That file should:
+
+* declare the platform-specific compile definitions and sources;
+* choose the default ``stream``, ``datagram`` and ``rawio`` backends for the platform.
+* optionally add extra include directories needed to build the platform sources
+  or adjust ``CHECK_THREADS`` if the platform needs it.
```
DenisBiryukov91: What if external platform definition relies on other 3rd-party libraries or some code generation mechanism ?

include/zenoh-pico/link/backend/default_ops.h
steils: Moved to config.h.in.

src/transport/unicast/accept.c
```diff
@@ -61,78 +62,124 @@ static void _zp_unicast_dispatch_connected_event(_z_transport_unicast_t *ztu, co
...
 _z_fut_fn_result_t _zp_unicast_accept_task_fn(void *ctx, _z_executor_t *executor) {
     _ZP_UNUSED(executor);
     _z_transport_unicast_t *ztu = (_z_transport_unicast_t *)ctx;
     const _z_sys_net_socket_t *socket_ptr = _z_link_get_socket(ztu->_common._link);
+    const _z_tcp_ops_t *tcp_ops = _zp_unicast_accept_tcp_ops(ztu->_common._link);
     if (socket_ptr == NULL) {
         _Z_ERROR_LOG(_Z_ERR_INVALID);
         return _z_fut_fn_result_ready();
     }
+    if (tcp_ops == NULL) {
```
DenisBiryukov91: why fail inside the task, where nobody will be notified about it, instead of doing this check before ?

src/transport/unicast/accept.c
```diff
@@ -61,78 +62,124 @@ static void _zp_unicast_dispatch_connected_event(_z_transport_unicast_t *ztu, co
...
+        default:
+            return NULL;
+    }
+}
+
 _z_fut_fn_result_t _zp_unicast_accept_task_fn(void *ctx, _z_executor_t *executor) {
     _ZP_UNUSED(executor);
     _z_transport_unicast_t *ztu = (_z_transport_unicast_t *)ctx;
     const _z_sys_net_socket_t *socket_ptr = _z_link_get_socket(ztu->_common._link);
+    const _z_tcp_ops_t *tcp_ops = _zp_unicast_accept_tcp_ops(ztu->_common._link);
```
DenisBiryukov91: why it should be hardcoded to be tcp-only ?

src/transport/unicast/accept.c
```diff
@@ -61,78 +62,124 @@ static void _zp_unicast_dispatch_connected_event(_z_transport_unicast_t *ztu, co
...
         }
+        return _z_fut_fn_result_wake_up_after(1000);
     }
+
     if (_z_transport_peer_unicast_slist_len(ztu->_peers) >= Z_LISTEN_MAX_CONNECTION_NB) {
-        // TODO: Send a connection refusal message before closing the socket
-        // TODO: Suspend task and only wake it up when a peer is removed and a new connection can be accepted.
         _Z_INFO("Refusing connection as max connections currently reached");
+#if Z_FEATURE_LINK_TLS == 1
+        _z_close_tls_socket(&con_socket);
```
DenisBiryukov91, line 88: why this can't be the part of _z_socket_close ?

steils: So that platform-specific socket layer (`src/system/*/network.c`) doesn't know anything about TLS.
We could make a 3rd function, a small helper on the transport layer that calls both functions, but I think that would not improve readability.

src/transport/unicast/accept.c
steils: Is it still relevant after dropping ops tables?

src/transport/unicast/accept.c
DenisBiryukov91: yes, but what if in the future we implement accept for udp or any other links, which has nothing to do with tls ?

src/transport/unicast/accept.c
steils: No longer relevant.

src/transport/unicast/accept.c
steils: Because UDP doesn't have `accept` at all?

DenisBiryukov91: it does not have as a protocol, but we still can accept udp connections in zenoh (although the logic is slightly different from tcp)

steils: Then we will add a higher-level helper.
But today this code path is only for TCP/TLS.

include/zenoh-pico/link/transport/socket.h
```diff
@@ -0,0 +1,41 @@
...
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+uintptr_t _z_socket_id(const _z_sys_net_socket_t *sock);
+z_result_t _z_socket_wait_readable(const _z_sys_net_socket_t *sockets, size_t count, uint8_t *ready,
+                                   uint32_t timeout_ms);
+
+z_result_t _z_socket_set_blocking(const _z_sys_net_socket_t *sock, bool blocking);
+z_result_t _z_socket_set_non_blocking(const _z_sys_net_socket_t *sock);
```
DenisBiryukov91: we should no longer need set_non_blocking, given that we have parametrized set_blocking, right ?

steils: Right, we have only one `_z_socket_set_non_blocking` call in the entire code. We `can _z_socket_set_blocking(, false)` instead.

src/system/socket/lwip.c
```diff
@@ -0,0 +1,133 @@
...
+    int fd = _z_lwip_socket_get(*sock);
+    int flags = lwip_fcntl(fd, F_GETFL, 0);
+    if (flags == -1) {
+        _Z_ERROR_RETURN(_Z_ERR_GENERIC);
+    }
+    if (lwip_fcntl(fd, F_SETFL, blocking ? (flags & ~O_NONBLOCK) : (flags | O_NONBLOCK)) == -1) {
+        _Z_ERROR_RETURN(_Z_ERR_GENERIC);
+    }
+    return _Z_RES_OK;
+}
```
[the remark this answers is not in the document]
steils: Invalid. Defined in `lwip/sockets.h`

src/link/transport/udp/udp_windows.c
```diff
@@ -0,0 +1,172 @@
...
+
+#include "zenoh-pico/utils/logging.h"
+#include "zenoh-pico/utils/pointers.h"
+
+static WSADATA _z_udp_windows_wsa_data;
+
+static z_result_t _z_udp_windows_endpoint_init(_z_sys_net_endpoint_t *ep, const char *s_address, const char *s_port) {
+    z_result_t ret = _Z_RES_OK;
+
+    if (WSAStartup(MAKEWORD(2, 2), &_z_udp_windows_wsa_data) == 0) {
```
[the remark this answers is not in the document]
steils: Not a bug, or far beyond the scope of this PR. May be a separate issue.

src/link/transport/udp/udp_windows.c
```diff
@@ -0,0 +1,172 @@
...
+
+static void _z_udp_windows_endpoint_clear(_z_sys_net_endpoint_t *ep) {
+    freeaddrinfo(ep->_ep._iptcp);
+    WSACleanup();
+}
+
+static z_result_t _z_udp_windows_open(_z_sys_net_socket_t *sock, const _z_sys_net_endpoint_t endpoint, uint32_t tout) {
+    z_result_t ret = _Z_RES_OK;
+
+    if (WSAStartup(MAKEWORD(2, 2), &_z_udp_windows_wsa_data) != 0) {
```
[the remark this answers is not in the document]
steils: Not a bug, or far beyond the scope of this PR. May be a separate issue.

src/link/transport/udp/udp_windows.c
```diff
@@ -0,0 +1,172 @@
...
+    return _Z_ERR_GENERIC;
+}
+
+static void _z_udp_windows_close(_z_sys_net_socket_t *sock) {
+    if (sock->_sock._fd != INVALID_SOCKET) {
+        closesocket(sock->_sock._fd);
+        sock->_sock._fd = INVALID_SOCKET;
+        WSACleanup();
+    }
+}
```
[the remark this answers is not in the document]
steils: Not a bug, or far beyond the scope of this PR. May be a separate issue.

src/link/transport/udp/udp_opencr.cpp
```diff
@@ -0,0 +1,167 @@
...
+    ep->_iptcp._addr = NULL;
+}
+
+static z_result_t _z_udp_opencr_open(_z_sys_net_socket_t *sock, const _z_sys_net_endpoint_t endpoint, uint32_t tout) {
+    _ZP_UNUSED(endpoint);
+    _ZP_UNUSED(tout);
+    z_result_t ret = _Z_RES_OK;
+
+    sock->_udp = new WiFiUDP();
+    if (!sock->_udp->begin(7447)) {  // FIXME: make it random
```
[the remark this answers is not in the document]
steils: Needs another PR

src/link/transport/udp/udp_opencr.cpp
```diff
@@ -0,0 +1,167 @@
...
+
+static size_t _z_udp_opencr_read(_z_sys_net_socket_t sock, uint8_t *ptr, size_t len) {
+    ssize_t rb = 0;
+    do {
+        rb = sock._udp->parsePacket();
+    } while (rb == 0);
+
+    if ((rb <= 0) || ((size_t)rb > len)) {
+        return 0;
+    }
```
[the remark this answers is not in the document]
steils: Needs another PR

src/system/socket/lwip.c
```diff
@@ -0,0 +1,133 @@
...
+        }
+    }
+
+    struct timeval timeout = {
+        .tv_sec = (time_t)(timeout_ms / 1000U),
+        .tv_usec = (suseconds_t)((timeout_ms % 1000U) * 1000U),
+    };
+    if (lwip_select(max_fd + 1, &read_fds, NULL, NULL, &timeout) <= 0) {
+        _Z_ERROR_RETURN(_Z_ERR_GENERIC);
+    }
```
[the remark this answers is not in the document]
steils: Needs another PR

src/link/unicast/tcp.c
```diff
@@ -119,29 +90,19 @@ uint16_t _z_get_link_mtu_tcp(void) {
...
-        ret = _Z_ERR_CONFIG_LOCATOR_INVALID;
-        goto cleanup;
-    }
+    z_result_t ret = _z_tcp_endpoint_init_from_address(&sys_endpoint, &endpoint->_locator._address);
 
-    ret = _z_create_endpoint_tcp(&sys_endpoint, s_address, s_port);
-    if (ret == _Z_RES_OK) {
-        ret = _z_open_tcp(socket, sys_endpoint, Z_CONFIG_SOCKET_TIMEOUT);
+    if (ret != _Z_RES_OK) {
+        _z_tcp_endpoint_clear(&sys_endpoint);
```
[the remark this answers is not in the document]
steils: Fixed

src/system/socket/esp32.c
```diff
@@ -0,0 +1,131 @@
...
+        }
+    }
+
+    struct timeval timeout = {
+        .tv_sec = (time_t)(timeout_ms / 1000U),
+        .tv_usec = (suseconds_t)((timeout_ms % 1000U) * 1000U),
+    };
+    if (select(max_fd + 1, &read_fds, NULL, NULL, &timeout) <= 0) {
+        _Z_ERROR_RETURN(_Z_ERR_GENERIC);
+    }
```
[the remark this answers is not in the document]
steils: Needs another PR

src/transport/unicast/read.c
```diff
@@ -141,6 +143,109 @@ z_result_t _zp_unicast_read(_z_transport_unicast_t *ztu, bool single_read) {
...
+}
+
+static z_result_t _z_unicast_socket_wait_snapshot_make(_z_transport_unicast_t *ztu,
+                                                       _z_unicast_socket_wait_snapshot_t *snapshot) {
+    _z_transport_peer_unicast_slist_t *curr = NULL;
+    size_t index = 0;
+
+    *snapshot = (_z_unicast_socket_wait_snapshot_t){0};
+
+    _z_transport_peer_mutex_lock(&ztu->_common);
```
DenisBiryukov91: Do we really need to acquire mutex here and in other functions, given that we only call them from read task, which, unless I miss something, cannot run concurrently with any other tasks that modify peer list (i.e. accept or connect)?

src/transport/unicast/read.c
```diff
@@ -141,6 +143,109 @@ z_result_t _zp_unicast_read(_z_transport_unicast_t *ztu, bool single_read) {
...
+        snapshot->count += 1;
+        curr = _z_transport_peer_unicast_slist_next(curr);
+    }
+
+    if (snapshot->count == 0) {
+        _z_transport_peer_mutex_unlock(&ztu->_common);
+        return _Z_RES_OK;
+    }
+
+    snapshot->sockets = (_z_sys_net_socket_t *)z_malloc(snapshot->count * sizeof(_z_sys_net_socket_t));
```
DenisBiryukov91: maybe snapshot should be just some sort of iterator over peers (or more precisely sockets), to avoid  repeatedly allocating memory (if my hypothesis above, that mutex here is not required is correct ?)

src/transport/unicast/read.c
steils: You're correct. Fixed.

src/transport/unicast/read.c
steils: Thanks. Reworked this part.

DenisBiryukov91 (approved):

sashacmc (approved):

PR #1189, opened 2026-03-19
fix(udp): read from _msock in multicast path

gmartin82:
## Description
Fixes an issue in the UDP multicast receive path where `_z_f_link_read_udp_multicast()` was using the unicast socket (`_sock`) instead of the multicast socket (`_msock`).

The read path now correctly uses `_msock` for multicast traffic.

### What does this PR do?
The multicast socket is correctly initialised and stored in `_msock` during setup, but the read path mistakenly used `_sock`. As a result, multicast packets were not received in most environments.

This issue could appear intermittently depending on file descriptor allocation (e.g. if both sockets happened to share the same descriptor), making it difficult to diagnose.

---
## 🏷️ Label-Based Checklist

**Labels:** `bug`

## 🐛 Bug Fix Requirements

**Instructions:**

### DIFF
--- a/src/link/multicast/udp.c
+++ b/src/link/multicast/udp.c
@@ -173,7 +173,7 @@ size_t _z_f_link_write_all_udp_multicast(const _z_link_t *self, const uint8_t *p
 }
 
 size_t _z_f_link_read_udp_multicast(const _z_link_t *self, uint8_t *ptr, size_t len, _z_slice_t *addr) {
-    return _z_read_udp_multicast(self->_socket._udp._sock, ptr, len, self->_socket._udp._lep, addr);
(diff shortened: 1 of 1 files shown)


gmartin82:
After further investigation, this change is not correct.

In the UDP multicast implementation, `_sock` and `_msock` serve different purposes:

* `_sock` is the multicast receive/listen socket and is used by the read path
* `_msock` is used for multicast send

UDP multicast links are created via `_z_listen_link()`, which calls `_z_f_link_listen_udp_multicast()` and initialises both sockets. As such, using `_sock` in `_z_f_link_read_udp_multicast()` is intentional and correct.

My original assumption that this was a typo was incorrect, so I’m closing this PR.

One observation is that `_z_f_link_open_udp_multicast()` is not used in practice and would be problematic for the write path if it were, since `_msock` is not initialised there. That may be worth addressing separately.

PR #1190, opened 2026-03-20
Asynchronous executor

DenisBiryukov91:
## Description

This PR introduces asynchronous executor. It allows to run all zenoh-pico tasks (accept, reconnect, read, send_join, send_keep_alive, lease, periodic tasks etc) in a single thread, which significantly simplifies task management. 

In addition, for single threaded mode zp_spin_once function is provided, which allows to spin executor manually. In particular this trivially extends all previously only multi-threaded features, like p2p, connectivity events, auto-reconnect and advanced pub-sub to single-threaded case.

### What does this PR do?

Replaces multiple threads with a single background executor.
Introduces zp_spin_once to allow spinning executor manually in single-threaded mode.
Updates all the examples: removes no-longer needed task start/stop calls from multi-threaded cases, replaces all zp_read/send_keep_alive/send_join with zp_spin_once in single-threaded cases.
Adds platform thread_id getters, and functions to measure elapsed time between 2 z_clock_t instances.

**cpp tests are failing because zp_read, zp_send_join, etc were put under Z_FEATURE_MULTI_THREAD=0 guard, this will be addressed separately on zenoh-cpp side**

**zp_read, zp_send_keep_alive, zp_send_join and related options were put under Z_FEATURE_MULTI_THREAD=0 guard and marked as deprecated**

**zp_start/stop_read_task and auto_start_read_task option were put under Z_FEATURE_MULTI_THREAD=1 guard and marked as deprecated, they currently start/stop entire executor**

**zp_start/stop_lease_task and auto_start_lease_task option were put under Z_FEATURE_MULTI_THREAD=1 guard and marked as deprecated, they currently do nothing**

**If no deprecated options are being used, the executor starts automatically if Z_FEATURE_MULTI_THREAD=1, the attributes for the background thread can be passed in executor_task_attributes field of z_open_options_t** 

Simplify task management.

### Related Issues

---
## 🏷️ Label-Based Checklist

**Labels:** `enhancement`

## ✨ Enhancement Requirements

- [ ] **No new APIs added** - Only improving existing functionality

**Instructions:**


src/transport/common/rx.c
```diff
@@ -32,55 +32,80 @@ size_t _z_read_stream_size(_z_zbuf_t *zbuf) {
...
+    return _Z_RES_OK;
+}
+
+z_result_t _z_link_recv_t_msg_cap_flow_datagram(const _z_link_t *zl, _z_zbuf_t *zbf, _z_sys_net_socket_t *socket) {
+    _ZP_UNUSED(socket);
+    if (_z_link_recv_zbuf(zl, zbf, NULL) == SIZE_MAX) {
+        return Z_ETIMEDOUT;
+    } else {
+        return _Z_ERR_TRANSPORT_RX_FAILED;
+    }
```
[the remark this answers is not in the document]
DenisBiryukov91: fixed

docs/config.rst
```diff
@@ -126,6 +126,8 @@ All the generated options must be changed in zenoh-pico's CMake (beware of CMake
 * `Z_BATCH_MULTICAST_SIZE`: Size of the multicast packet buffers, in bytes. Any packet bigger than this will be fragmented if possible.
 * `Z_CONFIG_SOCKET_TIMEOUT`: Timeout for socket options, if applicable, in milliseconds.
 * `Z_TRANSPORT_LEASE`: Maximum time without receiving messages from a connection before closing it, in milliseconds.
+* `Z_TRANSPORT_ACCEPT`: Link accept timeout in P2P mode in milliseconds (maximum amount of time the listening peer would wait to receive a response).
+* `Z_TRANSPORT_CONNECT`: Link connect timeout in milliseconds (maximum amount of time the connecting peer would wait to receive a response).
```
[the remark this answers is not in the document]
DenisBiryukov91: fixed

include/zenoh-pico/utils/mutex.h
```diff
@@ -0,0 +1,34 @@
...
+static inline z_result_t _z_mutex_rec_mt_lock(_z_mutex_rec_t *m) {
+    _ZP_UNUSED(m);
+    return _Z_RES_OK;
+}
+static inline z_result_t _z_mutex_rec_mt_unlock(_z_mutex_rec_t *m) {
+    _ZP_UNUSED(m);
+    return _Z_RES_OK;
+}
+#endif
+#endif
```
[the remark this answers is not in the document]
DenisBiryukov91: fixed

include/zenoh-pico/api/primitives.h
```diff
@@ -3359,22 +3302,20 @@ void zp_send_join_options_default(zp_send_join_options_t *options);
...
  *
  * Return:
  *   ``0`` if execution was successful, ``negative value`` otherwise.
- *
- * .. warning:: This API has been marked as unstable: it works as advertised, but it may be changed in a future release.
  */
-z_result_t zp_process_periodic_tasks(const z_loaned_session_t *zs);
+#if Z_FEATURE_MULTI_THREAD == 0
+void zp_spin_once(const z_loaned_session_t *zs);
 #endif
```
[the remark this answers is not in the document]
DenisBiryukov91: fixed

src/system/flipper/system.c
```diff
@@ -102,6 +102,9 @@ void _z_task_free(_z_task_t** task) {
     *task = NULL;
 }
 
+_z_task_id_t _z_task_get_id(const _z_task_t* task) { return NULL; }
+_z_task_id_t _z_task_current_id(void) { return NULL; }
+bool _z_task_id_equal(const _z_task_id_t* l, const _z_task_id_t* r) { return true; }
```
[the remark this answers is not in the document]
DenisBiryukov91: fixed

src/transport/multicast/lease.c
```diff
@@ -44,36 +45,32 @@ z_result_t _zp_multicast_send_keep_alive(_z_transport_multicast_t *ztm) {
...
 #if Z_FEATURE_AUTO_RECONNECT == 1
-    _z_reopen(&zs);
-    _z_session_rc_drop(&zs);
+    // Store weak session, to reuse for reconnection
+    ztm->_common._state = _Z_TRANSPORT_STATE_RECONNECTING;
+    ztm->_common._session = zs;
+    _z_fut_t f = _z_fut_null();
+    f._fut_arg = &ztm->_common;
+    f._fut_fn = _z_client_reopen_task_fn;
+    _z_executor_spawn(executor, &f);
```
steils: The result isn't checked here, so the transport task will be set to SUSPENDED anyway. But `_z_executor_spawn()` won't be able to run the reopen task, if we hit `Z_RUNTIME_MAX_TASKS`. That means it will suspend forever, right?

src/transport/unicast/lease.c
```diff
@@ -103,170 +96,113 @@ static void _zp_unicast_failed(_z_transport_unicast_t *ztu) {
...
-    _z_session_rc_drop(&zs_rc);
-    if (ret != _Z_RES_OK) {
-        _Z_ERROR("Reopen failed: %i", ret);
-    }
+    ztu->_common._state = _Z_TRANSPORT_STATE_RECONNECTING;
+    ztu->_common._session = weak_session_clone;
+    _z_fut_t f = _z_fut_null();
+    f._fut_arg = &ztu->_common;
+    f._fut_fn = _z_client_reopen_task_fn;
+    _z_executor_spawn(executor, &f);
```
steils: Same as for `src/transport/multicast/lease.c:69`

src/transport/multicast/lease.c
```diff
@@ -146,134 +126,69 @@ static void _zp_multicast_report_disconnected_events(_z_transport_multicast_t *z
...
+    }
+    ztm->_common._transmitted = false;
+    _z_transport_peer_mutex_lock(&ztm->_common);
+    unsigned long min_lease = (unsigned long)_z_get_minimum_lease(ztm->_peers, ztm->_common._lease);
+    _z_transport_peer_mutex_unlock(&ztm->_common);
 
-void *_zp_multicast_lease_task(void *ztm_arg) {
-    _ZP_UNUSED(ztm_arg);
-    return NULL;
+    return _z_fut_fn_result_wake_up_after(min_lease);
```
steils:
```
return _z_fut_fn_result_wake_up_after(min_lease / Z_TRANSPORT_LEASE_EXPIRE_FACTOR);
````
Reversed with `_zp_multicast_lease_task_fn()`?

File: src/transport/multicast/lease.c [hunk 9]
```diff
@@ -146,134 +126,69 @@ static void _zp_multicast_report_disconnected_events(_z_transport_multicast_t *z
...
+    while (curr_list != NULL) {
+        _z_transport_peer_multicast_t *curr_peer = _z_transport_peer_multicast_slist_value(curr_list);
+        curr_peer->common._received = false;
+        curr_list = _z_transport_peer_multicast_slist_next(curr_list);
     }
-    return 0;
+    unsigned long min_lease = (unsigned long)_z_get_minimum_lease(ztm->_peers, ztm->_common._lease);
+    _z_transport_peer_mutex_unlock(&ztm->_common);
+    _zp_multicast_report_disconnected_events(ztm, &dropped_peers);
+    return _z_fut_fn_result_wake_up_after(min_lease / Z_TRANSPORT_LEASE_EXPIRE_FACTOR);
```
steils:
```
return _z_fut_fn_result_wake_up_after(min_lease);
```
Reversed with `_zp_multicast_keep_alive_task_fn()`?

src/transport/raweth/read.c
```diff
@@ -44,94 +45,39 @@ z_result_t _zp_raweth_read(_z_transport_multicast_t *ztm, bool single_read) {
...
+        } else {
+            _z_t_msg_clear(&t_msg);
+            if (_z_raweth_update_rx_buff(ztm) != _Z_RES_OK) {
+                _Z_ERROR("Connection closed due to lack of memory to allocate rx buffer");
+                return _z_fut_fn_result_ready();
+            }
         }
+    } else if (ret != _Z_ERR_TRANSPORT_RX_FAILED) {
+        _Z_ERROR("Connection closed due to malformed message: %d", ret);
+        return _z_fut_fn_result_ready();
```
steils, line 83: Before the refactoring `_zp_raweth_read_task()` worked as `while (ztm->_common._read_task_running)`, and on `_Z_ERR_TRANSPORT_RX_FAILED` made `continue` and waited for the next message.
Now `_zp_raweth_read_task_fn()` always returns READY, even with `ret == _Z_ERR_TRANSPORT_RX_FAILED`. I presume it should be something like `continue` or maybe `wake_up_after()` here.

steils (changes requested):

src/transport/multicast/lease.c
DenisBiryukov91: Yes, so essentially the same situation as no attempt to reconnect. I believe it is worthing logging a error here and just do nothing.

src/api/api.c
steils, line 2522: `z_open()` now starts the transport tasks unconditionally (api.c:910). But the public API still exposes `zp_read()` and `zp_start/stop_*_task()`.
While `zp_start/stop_*_task()` are stubs, `zp_read` really calls `_zp_*_read`, where `ztu->_common._zbuf` is reset and then reused for reading and parsing. That looks unsafe.
We'll have either drop `zp_read` or make it a stub, or rework somehow.

steils (changes requested):

src/transport/manager.c
```diff
@@ -106,18 +107,20 @@ static z_result_t _z_new_transport_peer(_z_transport_t *zt, const _z_string_t *l
...
-                                                      false, NULL);
-                    }
+                    _z_transport_peer_unicast_add(&zt->_transport._unicast, &tp_param, *_z_link_get_socket(zl), false,
+                                                  NULL);
                 } else {
-                    _zp_unicast_start_accept_task(&zt->_transport._unicast);
+                    _z_fut_t f = _z_fut_null();
+                    f._fut_arg = &zt->_transport._unicast;
+                    f._fut_fn = _zp_unicast_accept_task_fn;
+                    _z_runtime_spawn(runtime, &f);
```
steils: No result check. If it fails `_z_new_transport_peer` will return success but it shouldn't.

include/zenoh-pico/runtime/runtime.h
```diff
@@ -0,0 +1,67 @@
...
+// Contributors:
+//   ZettaScale Zenoh Team, <zenoh@zettascale.tech>
+//
+
+#ifndef ZENOH_PICO_RUNTIME_RUNTIME_H
+#define ZENOH_PICO_RUNTIME_RUNTIME_H
+
+#include "zenoh-pico/config.h"
+
+#define _ZP_EXECUTOR_MAX_NUM_FUTURES Z_RUNTIME_MAX_TASKS
```
steils, line 20: Now this is used both for transport tasks and optional higher-level background tasks (for example query timeout processing or periodic advanced subscriber tasks), while they all have same priority. What worries me is that if we hit this limit it can block or delay reconnect / lease / keep-alive tasks.

steils (changes requested):

include/zenoh-pico/runtime/runtime.h
DenisBiryukov91: This is all relative: in original version query cleanup is a part of a lease task ... 
In this PR read/lease/keep_alive tasks are just resumed after reconnection instead of spawning a bunch of threads - which should be way faster. 
Periodic queries is a separate question, which is also problematic in original version, since it does not really scale well in a network with many publishers
To improve a bit we can spawn reconnection task upon transport creation in a suspended state, this way it will always have a guaranteed slot inside the executor, but I believe it is better to be done together with addressing p2p connection in a separate PR

src/transport/manager.c
DenisBiryukov91: fixed

src/transport/raweth/read.c
DenisBiryukov91: fixed

src/transport/multicast/lease.c
DenisBiryukov91: fixed

src/transport/multicast/lease.c
DenisBiryukov91: fixed

src/transport/unicast/lease.c
DenisBiryukov91: fixed

src/transport/multicast/lease.c
DenisBiryukov91: fixed

src/api/api.c
DenisBiryukov91: Yes, this is something worth discussing internally

src/transport/common/rx.c
```diff
@@ -32,55 +32,80 @@ size_t _z_read_stream_size(_z_zbuf_t *zbuf) {
...
+                _Z_ERROR_LOG(_Z_ERR_TRANSPORT_RX_FAILED);
+                return _Z_ERR_TRANSPORT_RX_FAILED;
+            }
+        } else {
+            _Z_ERROR_LOG(_Z_ERR_TRANSPORT_NO_SPACE);
+            return _Z_ERR_TRANSPORT_NO_SPACE;
+        }
+    } else if (read == 0 || read == SIZE_MAX) {
+        return Z_ETIMEDOUT;
+    } else {
```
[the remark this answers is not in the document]
DenisBiryukov91: fixed

src/api/advanced_publisher.c
```diff
@@ -166,10 +166,12 @@ static z_result_t _ze_advanced_publisher_ke_suffix(z_owned_keyexpr_t *suffix, co
...
 
-static void _ze_advanced_publisher_heartbeat_handler(void *ctx) {
+static _z_fut_fn_result_t _ze_advanced_publisher_heartbeat_handler(void *ctx, _z_executor_t *executor) {
+    _ZP_UNUSED(executor);
     _ze_advanced_publisher_state_weak_t *state_weak_rc = (_ze_advanced_publisher_state_weak_t *)ctx;
     _ze_advanced_publisher_state_rc_t state_rc = _ze_advanced_publisher_state_weak_upgrade(state_weak_rc);
 
+    unsigned long heartbeat_period_ms = state_rc._val->_heartbeat_period_ms;
     if (!_Z_RC_IS_NULL(&state_rc)) {
         _ze_advanced_publisher_state_t *state = _Z_RC_IN_VAL(&state_rc);
```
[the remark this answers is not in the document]
DenisBiryukov91: fixed

src/transport/unicast/lease.c
```diff
@@ -103,170 +96,120 @@ static void _zp_unicast_failed(_z_transport_unicast_t *ztu) {
...
-                    }
-                }
-                // Reset the keep alive parameters
-                ztu->_common._transmitted = false;
-                next_keep_alive = (int)(ztu->_common._lease / Z_TRANSPORT_LEASE_EXPIRE_FACTOR);
-            }
+        if (curr_peer->common._received) {
+            // Reset the lease parameters
+            curr_peer->common._received = false;
+            return _z_fut_fn_result_wake_up_after(Z_TRANSPORT_LEASE);
```
steils: Why we don't take into account `ztu->_common._lease` here? If the peer negotiated a smaller than `Z_TRANSPORT_LEASE` lease in handshake, won't that mean this will send keep-alives later than needed?

src/transport/unicast/lease.c
```diff
@@ -103,170 +96,120 @@ static void _zp_unicast_failed(_z_transport_unicast_t *ztu) {
...
             }
         }
-
-        // The keep alive and lease intervals are expressed in milliseconds
-        z_sleep_ms((size_t)interval);
-
-        next_lease = next_lease - interval;
-        next_keep_alive = next_keep_alive - interval;
+        ztu->_common._transmitted = false;
+        return _z_fut_fn_result_wake_up_after(Z_TRANSPORT_LEASE / Z_TRANSPORT_LEASE_EXPIRE_FACTOR);
```
steils: Same question as for :144.

src/runtime/background_executor.c
```diff
@@ -0,0 +1,284 @@
...
+                be->_thread_idx) {  // extra check to allow to stop executor thread even if there are waiters
+                break;              // stop requested, exit the loop and end the thread
+            }
+            // if there are waiters, sleep until they are resumed
+            _Z_CLEAN_RETURN_IF_ERR(_z_condvar_wait(&be->_condvar, &be->_mutex), _z_mutex_unlock(&be->_mutex));
+        }
+        if (thread_idx < be->_thread_idx) {
+            break;  // stop requested, exit the loop and end the thread
+        }
+        _z_executor_spin_result_t res = _z_executor_spin(&be->_executor);
```
steils, line 69: So `be->_mutex` is held here. But if the user calls for example `z_close()` or `zp_stop_read_task()` from the background executor thread, or something that later calls [`_z_background_executor_stop()` => ... => ] `_z_background_executor_inner_suspend_and_lock()` (which takes that mutex ones more) won't it be a deadlock?

src/transport/manager.c
steils, line 45: Shouldn't client-side unicast socket be switched to non-blocking, as it is for peers? I'm afraid in the new runtime model, a blocking client read could stall lease / keep-alive / reconnect.
Maybe that's for a follow-up, not this PR.

steils (changes requested):

src/runtime/background_executor.c
DenisBiryukov91: There indeed will be a deadlock, closing the session from inside of the callback is forbidden, same as it is in the current main.

src/transport/manager.c
DenisBiryukov91: It is exactly the same as in the main - a blocking read will prevent setting `_received = true` which might eventually lead to the failure of the lease task.

src/transport/unicast/lease.c
DenisBiryukov91: fixed

src/transport/unicast/lease.c
DenisBiryukov91: fixed

src/runtime/background_executor.c
DenisBiryukov91: Fixed, now a error is return on attempt to manipulate executor thread from inside itself.

src/transport/manager.c
steils: Yes, I understand that it's the same as in current master. That's a concern for a follow-up, not a blocker for this PR.

src/api/api.c
steils: Now `_z_session_close()` does `_Z_RETURN_IF_ERR(_z_runtime_stop(...))`, but we don't handle it here, always returning _Z_RES_OK.

steils (changes requested):

src/api/api.c
DenisBiryukov91: fixed

steils (approved):

PR #1199, opened 2026-04-10
Fix dead lock at session mutex

riebl:
## Description
With LOCAL_PUBLISHER and LOCAL_QUERYABLE enabled, I noticed a deadlock when publishing from a query callback (using zenoh-cpp with zenoh-pico backend).

### What does this PR do?
Callbacks are no longer invoked while session mutex is still locked. Instead, the callback function pointer and its argument are stored on stack and invoked as soon as session mutex is unlocked.

Publisher blocks infinitely at session mutex when called from a callback and session mutex is already acquired by call stack.

### Related Issues
N/A

---
## 🏷️ Label-Based Checklist

Current labels: _No labels_

### DIFF
--- a/src/session/liveliness.c
+++ b/src/session/liveliness.c
@@ -158,7 +158,11 @@ static z_result_t _z_liveliness_pending_query_reply(_z_session_t *zn, uint32_t i
     _Z_RETURN_IF_ERR(_z_get_keyexpr_from_wireexpr(zn, &ke, wireexpr, peer, true));
     z_result_t ret = _Z_RES_OK;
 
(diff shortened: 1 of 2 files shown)


DenisBiryukov91:
It is generally not advised to do any blocking operation (which the publish is) from callbacks, since this starves zenoh-pico background thread - one can use channels if needed for this purpose.
So it is not entirely clear whether the issue needs fixing.

riebl:
Yes, I also merely schedule the publisher invocation in my callback now. However, it can be a quite unpleasant surprise when switching from zenoh-c to zenoh-pico as backend for zenoh-cpp and suddenly the application is trapped in a deadlock.

Issue #1202, opened 2026-04-13
[Bug] Local Replies are not delivered in peer mode


DenisBiryukov91:
### Describe the bug

When making a local query on session in peer mode, the replies are not delivered (the local queryable still receives a query). 

### To reproduce

Start session in peer mode.
Create queryable on this session.
Make local get from this session to queryable declared above.
Reply to the query received by queryable.
Reply will not be delivered to get callback

### System info

any

DenisBiryukov91:
The issue comes from p2p not properly handling number of awaited response_finals: i.e. it should normally correspond to the number of peers at time of query (and would likely need to somehow account for disconnects ???), while it is always set to 1. In particular in case when no peers are connected to the session, the query simply timeouts while infinitely waiting for an extra response final which never arrives. The local response is not delivered due to default LATEST consolidation, which only delivers consolidated response after all response_finals arrive.

PR #1203, opened 2026-04-13
Fix count of reply_final in p2p mode

DenisBiryukov91:
## Description

Fixes calculation of expected number of ResponseFinal messages when sending a query in p2p mode

### What does this PR do?

Fix calculation of expected number of ResponseFinal messages when sending a query in p2p mode

It fixes a query replies delivery logic
### Related Issues

#1202

---
## 🏷️ Label-Based Checklist

**Labels:** `bug`

## 🐛 Bug Fix Requirements

**Instructions:**

### DIFF
--- a/include/zenoh-pico/session/session.h
+++ b/include/zenoh-pico/session/session.h
@@ -194,7 +194,7 @@ struct _z_pending_query_t {
     z_clock_t _start_time;
     uint64_t _timeout;
     void *_arg;
-    uint8_t _remaining_finals;
+    uint32_t _remaining_finals;
     _z_pending_reply_slist_t *_pending_replies;
     z_query_target_t _target;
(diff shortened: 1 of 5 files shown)


gmartin82 (approved):
This looks like the right fix for the peer-mode case 👍

One thing to flag: `_remaining_finals` is still stored as a `uint8_t`, but `remote_targets` now comes from `_z_transport_get_peers_count()`, so it can exceed 255.

Even if that’s unlikely in typical zenoh-pico deployments, this value is used to determine when a query is complete. If it overflows, the query could complete early (or behave incorrectly) because the expected number of ResponseFinals is truncated.

Might be worth either widening `_remaining_finals` or explicitly checking so the behaviour is well-defined.

Issue #1103, opened 2025-11-29
[Bug] Task Cleanup Race Condition and Reconnection Issues


Vladislavert:
### Summary

When a zenoh-pico client on ESP32 loses connection to the zenohd router (e.g., router stops), the ESP32 crashes with a spinlock assertion failure during task cleanup.

### Crash Log

```
assert failed: spinlock_acquire spinlock.h:142 (lock->count == 0)

Backtrace: 0x40376d92:0x3fcbba80 0x4037e3b1:0x3fcbbaa0 ...

0x4037ae73: xEventGroupSetBits at /idf/components/freertos/FreeRTOS-Kernel/event_groups.c:245
0x42015fda: z_task_wrapper at zenoh-pico/src/system/espidf/system.c:64
```

### Root Cause

In `src/system/espidf/system.c`, `z_task_wrapper` calls `vTaskDelete(NULL)` immediately after `xEventGroupSetBits`.
If `_z_task_free` runs immediately (e.g. on another core), it frees the Event Group while `xEventGroupSetBits` is still using it.
This causes a crash when `xEventGroupSetBits` tries to access the freed memory.

### Workaround

I patched `src/system/espidf/system.c` to use `vTaskSuspend(NULL)` instead of `vTaskDelete(NULL)`, and modified `_z_task_free()` to properly wait for task completion before deleting resources:

```c
// In z_task_wrapper:
vTaskSuspend(NULL);  // Instead of vTaskDelete(NULL)

// In _z_task_free:
xEventGroupWaitBits(ptr->join_event, 1, pdFALSE, pdFALSE, pdMS_TO_TICKS(500));
if (ptr->handle != NULL) { vTaskDelete(ptr->handle); ptr->handle = NULL; }
vEventGroupDelete(ptr->join_event);  // Instead of z_free(ptr->join_event)
```

### Related Issue

This appears related to [#1033](https://github.com/eclipse-zenoh/zenoh-pico/issues/1033).

### To reproduce

1. Flash and run the standard `z_pub` example on ESP32 
2. Start an instance of zenohd(`docker run --rm --init --net host  eclipse/zenoh:latest -l udp/0.0.0.0:7447`)
3. Stop zenohd
4. Start zenohd again
5. Zenoh pico will crash due to FreeRTOS attempting to acquire a spinlock on freed memory 

### System info

- **zenoh-pico version**: 1.6.2
- **zenohd version**: 1.6.2
[...]

PR #1205, opened 2026-04-16
fix: prevent use-after-destroy of TX mutex during transport teardown

Vuk-SFL:
## Description
Adds an atomic _tx_ready guard to prevent use-after-destroy of the TX mutex during transport teardown.

### What does this PR do?

A new _z_atomic_bool_t _tx_ready field is added to _z_transport_common_t. It is set to true on transport creation and atomically set to false in _z_transport_common_clear before the mutex is destroyed. All TX send paths check this flag before acquiring the TX mutex, dropping the message with an info log if the transport is shutting down.

This change adresses race conditions in mutlithreaded app when executor thread is doing reconneciton while tx thread is still sending data.

There is a race between TX send paths and transport teardown: _z_transport_common_clear destroys the TX mutex while a concurrent send may still be holding or about to acquire it. This causes use-after-destroy of the mutex, leading to undefined behavior (segfaults, assertion failures, or silent corruption depending on the platform).

The atomic flag provides a lock-free check that is ordered before the mutex acquisition, so the teardown path can guarantee that no new senders will enter the critical section after _tx_ready is set to false. A subsequent drain (lock + unlock) ensures any in-flight sender has released the mutex before it is destroyed.

### Related Issues
Related to https://github.com/eclipse-zenoh/zenoh-pico/issues/1165 — segmentation fault when calling batching operations during transport reconnect. This PR guards the same TX send paths (including _z_transport_tx_send_n_batch) against concurrent teardown.

Related to https://github.com/eclipse-zenoh/zenoh-pico/issues/1103 — task cleanup race condition. While #1103's root cause is in the ESP-IDF task layer, the underlying pattern (resources accessed after teardown begins) is the same class of bug addressed here for the TX mutex.

---
## 🏷️ Label-Based Checklist

**Labels:** `bug`

## 🐛 Bug Fix Requirements

**Instructions:**

### DIFF
--- a/src/transport/common/tx.c
+++ b/src/transport/common/tx.c
@@ -270,14 +270,14 @@ z_result_t _z_transport_tx_send_t_msg(_z_transport_common_t *ztc, const _z_trans
                                       _z_transport_peer_unicast_slist_t *peers) {
     z_result_t ret = _Z_RES_OK;
     _Z_DEBUG("Send session message");
-
+    // If sending to a peer list, make sure the peer mutex is locked
+    _z_transport_tx_mutex_lock(ztc, true);
     ret = _z_transport_tx_check_ready(ztc);
     if (ret != _Z_RES_OK) {
         _Z_INFO("Dropping zenoh message because transport is not ready");
+        _z_transport_tx_mutex_unlock(ztc);
         return ret;
     }
-    // If sending to a peer list, make sure the peer mutex is locked
-    _z_transport_tx_mutex_lock(ztc, true);
 
     ret = _z_transport_tx_send_t_msg_inner(ztc, t_msg, peers);
 
@@ -295,18 +295,21 @@ static z_result_t _z_transport_tx_send_n_msg(_z_transport_common_t *ztc, const _
     z_result_t ret = _Z_RES_OK;
     _Z_DEBUG("Send network message");
 
-    ret = _z_transport_tx_check_ready(ztc);
-    if (ret != _Z_RES_OK) {
-        _Z_INFO("Dropping zenoh message because transport is not ready");
-        return ret;
-    }
-
     // Acquire the lock and drop the message if needed
     if (!_z_transport_batch_hold_tx_mutex()) {
         ret = _z_transport_tx_mutex_lock(ztc, cong_ctrl == Z_CONGESTION_CONTROL_BLOCK);
     }
     if (ret != _Z_RES_OK) {
         _Z_INFO("Dropping zenoh message because of congestion control");
+        return ret;
+    }
+
+    ret = _z_transport_tx_check_ready(ztc);
+    if (ret != _Z_RES_OK) {
+        _Z_INFO("Dropping zenoh message because transport is not ready");
+        if (!_z_transport_batch_hold_tx_mutex()) {
+            _z_transport_tx_mutex_unlock(ztc);
+        }
         return ret;
     }
     // Process message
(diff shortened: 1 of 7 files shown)


src/transport/common/tx.c
```diff
@@ -270,6 +270,12 @@ z_result_t _z_transport_tx_send_t_msg(_z_transport_common_t *ztc, const _z_trans
...
     z_result_t ret = _Z_RES_OK;
     _Z_DEBUG("Send session message");
+
+    ret = _z_transport_tx_check_ready(ztc);
+    if (ret != _Z_RES_OK) {
+        _Z_INFO("Dropping zenoh message because transport is not ready");
+        return ret;
+    }
     // If sending to a peer list, make sure the peer mutex is locked
     _z_transport_tx_mutex_lock(ztc, true);
```
DenisBiryukov91, line 280: You still have a data race here: you can read tx_ready=true, then this thread gets descheduled, while background one clears the transport.

Vuk-SFL: You are right. In latest commit I added additional (mutex use count based) check to be sure no one is using mutex when it is destroyed.

Vuk-SFL:
With this change following situations are handled in multithread scenario when one thread is performing tx while executor is duing transport cleanup  (during reonnection for example):

- Thread tries to send after cleanup started: _tx_ready is set to false first, so it fails fast with _Z_ERR_TRANSPORT_TX_FAILED.
- Thread is in lock-entry path while cleanup starts: _tx_lock_attempts is incremented before proceeding, and clear waits until this counter reaches 0.
- Thread is blocked on _mutex_tx when cleanup starts: it is still counted in _tx_lock_attempts, so clear cannot drop mutex yet.
- Thread already entered TX critical section when cleanup starts: clear does a drain lock/unlock on _mutex_tx, so it waits for that in-flight critical section to exit.
- Thread acquires _mutex_tx right around shutdown: post-lock _tx_ready recheck forces unlock+abort before touching TX resources.
- Batched path with externally held TX mutex: explicit _z_transport_tx_check_ready in send paths blocks TX work when _tx_ready=false.
- Mutex destruction while someone may still lock it: prevented by (_tx_ready=false) + waiting on _tx_lock_attempts + drain lock.
- Buffer/link/session free while TX still active: free happens only after the above synchronization steps complete.

DenisBiryukov91:
I think destroying tx mutex is unnecessary if we are going to reconnect anyways (since it and other parts of transport_common could be reused after reconnection succeeds), so I believe a more appropriate fix would be to address this issue (and check _state field), rather than introduce more synchronization primitives

Vuk-SFL:
I see your point. I will take a look at reconneciton handling part as well and see how state could be utilized/extended.

Even without auto reconnection taking place (since it is optional feature) race conditions can still exist.
For example in teardown scenario when executor thread lease is closing transport while thread performing tx is still working.

DenisBiryukov91:
> Even without auto reconnection taking place (since it is optional feature) race conditions can still exist. For example in teardown scenario when executor thread lease is closing transport while thread performing tx is still working.

I suppose in this case we should just delegate destruction of transport_common to z_session_drop.

Vuk-SFL:
@DenisBiryukov91 I implemented and validated your proposal and got rid of additional variables for bookeeping of mux access
> I suppose in this case we should just delegate destruction of transport_common to z_session_drop.

Now combination of `_z_transport_tx_mutex_lock` followed `_z_transport_tx_check_ready` in tx methods guards from using resources teared down. Teardown still waits for mux to be released before it deletes data.

 ` _mutex_tx`  is created once in `_z_multicast/unicast_transport_create` and survives reconenects until session is closed.

dusan19:
Tested the change in the context of PX4 flight controller FW on nuttx. 
- Hardfaults are resolved.

Are there more concerns before moving forward to merging this PR?

PetervdPerk-NXP:
@DenisBiryukov91 @gmartin82 Could you help out with this PR, it would be great if this issue can be resolved.

PR #1207, opened 2026-04-17
fix: exchange declarations on both sides of unicast peer connections

YuanYuYuan:
## Summary

In unicast peer mode, declaration exchange was asymmetric: only the **accepting** side of each connection received the remote peer's declarations. The connecting side never pushed its own declarations, and INTEREST messages on unicast transports were silently dropped. In a full mesh of N peers, a publisher on node A would never learn about subscribers declared by node B if B had connected to A rather than the other way around — resulting in silent message loss.

Three code paths are fixed:

- `_z_new_peer` (`src/transport/manager.c`) — after establishing an outbound connection, push the local session's declarations to the newly added peer via `_z_interest_push_declarations_to_peer`, mirroring what the accept task already does.
- `_z_interest_process_interest` (`src/session/interest.c`) — instead of returning immediately on unicast, respond to `CURRENT` interests by pushing matching declarations (resources first, then subscribers/queryables/tokens, then `DeclareFinal`) back to the requesting peer.
- `_z_add_interest` (`src/net/primitives.c`) — include unicast peer mode in the condition that sends an INTEREST on the wire, so that publishers can pull current subscriber state from connected peers when building their write filter.

## Key Changes

- `src/transport/manager.c`: call `_z_interest_push_declarations_to_peer` from `_z_new_peer` after peer is added
- `src/session/interest.c`: handle CURRENT INTEREST on unicast by responding with declarations; always send resource declarations before subscriber/queryable declarations so the receiver can resolve wire-encoded resource IDs
- `src/net/primitives.c`: send INTEREST on unicast peer transport (previously only sent for client mode, router peers, and multicast)
- `tests/z_test_peer_mesh_declare.c`: two new tests — one verifying outbound declare push via `_z_new_peer`, one verifying write filter population via INTEREST using `Z_TEST_HOOKS` to block the accept-path passive push

## Breaking Changes

None

---
## 🏷️ Label-Based Checklist

**Labels:** `bug`

## 🐛 Bug Fix Requirements

**Instructions:**


src/transport/manager.c
```diff
@@ -181,7 +183,12 @@ z_result_t _z_new_peer(_z_transport_t *zt, const _z_id_t *session_id, const _z_s
                 _z_socket_close(&socket);
                 return ret;
             }
-            ret = _z_transport_peer_unicast_add(&zt->_transport._unicast, &tp_param, socket, true, NULL);
+            _z_transport_peer_unicast_t *new_peer = NULL;
```
DenisBiryukov91, line 186: This part does not seem to be necessary, since p2p does not support reconnect as of now, and on the first connection, there should not be any declarations before _z_open returns.
The whole p2p connection procedure should be reworked and likely implemented as a separate async background task that regularly checks for disconnected peers.

DenisBiryukov91 (approved):

src/transport/manager.c
YuanYuYuan: Yes. I'm working on the concurrent version 🙂

DenisBiryukov91: This should be in a separate PR though, since this will likely exceed the scope of this fix

PR #1210, opened 2026-04-20
fix(advanced-subscriber): avoid double free when closing late-join background subscribers

evshary:
## Description

This fixes a flaky failure in `z_api_callback_drop_on_undeclare_test`.

### What was happening

The failing CI run crashes with `free(): double free detected in tcache 2` while running the background late-join advanced-subscriber case in `test_advanced_subscriber_late_join_callback_drop_on_undeclare(true)` from `tests/z_api_callback_drop_on_undeclare_test.c`.

This case enables `detect_late_publishers`, so the advanced subscriber may still be holding internal late-join state and may also be starting history recovery queries while the background session is being closed.

### Root cause

There were two double-free paths in advanced-subscriber teardown.

1. `_ze_advanced_subscriber_subscriber_drop_handler()` was clearing advanced-subscriber internal containers during subscriber drop:
   - `_miss_handlers`
   - `_sequenced_states`
   - `_timestamped_states`

   Later, `_ze_advanced_subscriber_state_clear()` cleared the same containers again when the advanced-subscriber state itself was finally released. In the late-join background case those containers can be populated, so this duplicated cleanup could free the same allocations twice.

2. The late-join query path allocates a per-query `query_ctx` and passes it to `z_get()` through the reply closure.

   During `z_close()`, query startup can fail while `z_get()` is already consuming that closure. In that case, `z_get()` may trigger `_ze_advanced_subscriber_query_drop_handler()`, which frees `query_ctx`.

   But the advanced-subscriber query wrapper error paths were also freeing the same `query_ctx` in:
   - `_ze_advanced_subscriber_initial_query()`
   - `_ze_advanced_subscriber_sequenced_query()`
   - `_ze_advanced_subscriber_timestamped_query()`

   So session teardown could race with late-join query startup and both sides could free the same query context.

### Why it looked flaky

Both problems depend on teardown timing in the late-join background case. The failure only appears when session close overlaps with internal late-join cleanup and/or history query startup, which makes the crash intermittent in CI even though the underlying bugs are real double frees.

### What does this PR do?

This PR makes teardown ownership explicit in both places.

1. `_ze_advanced_subscriber_subscriber_drop_handler()` now only marks the state as undeclaring and cancels outstanding work. It no longer clears `_miss_handlers`, `_sequenced_states`, or `_timestamped_states` itself.
2. `_ze_advanced_subscriber_run_query()` now owns cleanup of `query_ctx` on pre-dispatch failures, and the individual query wrappers no longer free `query_ctx` themselves on error.

This leaves `_ze_advanced_subscriber_state_clear()` as the single owner of container cleanup and keeps query-context ownership consistent with the reply closure drop path, preventing both double-free races during late-join session close.

### Related Issues

 * https://github.com/eclipse-zenoh/zenoh-pico/actions/runs/24320585645/job/71005915340

---
## 🏷️ Label-Based Checklist

**Labels:** `bug`

## 🐛 Bug Fix Requirements

**Instructions:**

### DIFF
--- a/src/api/advanced_subscriber.c
+++ b/src/api/advanced_subscriber.c
@@ -897,11 +897,14 @@ static z_result_t _ze_advanced_subscriber_run_query(_ze_advanced_subscriber_quer
                                                     const z_loaned_keyexpr_t *keyexpr, const char *params) {
     if (_Z_RC_IS_NULL(rc_state)) {
         _Z_ERROR("Failed to run query - state is NULL");
+        // Query context is still locally owned until the reply closure is handed to z_get().
+        _ze_advanced_subscriber_query_ctx_free(ctx);
         _Z_ERROR_RETURN(_Z_ERR_GENERIC);
     }
 
     _ze_advanced_subscriber_state_t *state = _Z_RC_IN_VAL(rc_state);
-    _Z_RETURN_IF_ERR(_ze_advanced_subscriber_state_rc_copy(&ctx->_statesref, rc_state));
+    _Z_CLEAN_RETURN_IF_ERR(_ze_advanced_subscriber_state_rc_copy(&ctx->_statesref, rc_state),
+                           _ze_advanced_subscriber_query_ctx_free(ctx));
 
     z_owned_closure_reply_t callback;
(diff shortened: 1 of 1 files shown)


gmartin82 (approved):
Looks good.

sashacmc (approved):

PR #1213, opened 2026-04-21
Fix z_random_fill: advance pointer across buffer

jerry73204:
Previous loop rewrote byte 0 `len` times instead of filling all `len` bytes — any caller reading past byte 0 got uninitialised memory (typically zeros). Affects zenoh session IDs, ephemeral port seeds, and any other entropy-consuming path on FreeRTOS.

---
## 🏷️ Label-Based Checklist

**Labels:** `bug`

## 🐛 Bug Fix Requirements

**Instructions:**

### DIFF
--- a/src/system/freertos/system.c
+++ b/src/system/freertos/system.c
@@ -57,8 +57,9 @@ uint64_t z_random_u64(void) {
 }
 
 void z_random_fill(void *buf, size_t len) {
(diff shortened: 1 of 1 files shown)


DenisBiryukov91:
Test not included, since we do not run tests on FreeRtos in ci.

DenisBiryukov91 (approved):

PR #1218, opened 2026-05-07
Fix ESP-IDF CI build

gmartin82:
## Description

Fix PlatformIO source filtering so platform-specific backend files are selected consistently with the CMake platform profiles.

PlatformIO was auto-discovering sources outside `src/system/**`, including platform-specific transport backends under `src/link/transport/**`. This meant builds could include sources for the wrong platform, for example ESP-IDF builds attempting to compile Arduino ESP32 C++ transport files.

### What does this PR do?

This PR updates `extra_script.py` to:

- exclude platform-specific system and transport backend directories from PlatformIO’s default source discovery;
- read the selected CMake platform profile from `cmake/platforms/<platform>.cmake`;
- derive PlatformIO source-filter add-backs from the `ZP_PLATFORM_SOURCE_FILES` listed by that CMake profile;
- keep the CMake platform files as the source of truth for platform-specific source composition.

The previous PlatformIO filter treated `src/system/**` as the platform-specific boundary, but the platform-composition rework also has platform-specific files under `src/link/transport/**`, such as Arduino ESP32 Bluetooth and serial sources.

Because PlatformIO starts from `+<*>`, those files could be picked up even when the active platform profile did not select them. This caused incorrect cross-platform compilation, for example ESP-IDF builds trying to compile Arduino-specific C++ files.

Deriving the PlatformIO filter from the CMake platform profile avoids duplicating the source list in `extra_script.py` and keeps PlatformIO aligned with the existing CMake platform selection.

### Related Issues

Related to PlatformIO build failures where ESP-IDF attempted to compile Arduino ESP32 transport sources.

---
## 🏷️ Label-Based Checklist

**Labels:** `internal`

## 🏠 Internal Change

**Instructions:**

### DIFF
--- a/extra_script.py
+++ b/extra_script.py
@@ -12,7 +12,9 @@
 #   ZettaScale Zenoh Team, <zenoh@zettascale.tech>
 #
 import os
+import re
 import subprocess
+import sys
 
 Import('env', 'projenv')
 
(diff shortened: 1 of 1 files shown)


extra_script.py
```diff
@@ -43,17 +91,17 @@ def _platform_src_filter(system_dir):
     BOARD = 'generic'
 
 if FRAMEWORK == 'zephyr':
-    SRC_FILTER = _platform_src_filter("system/zephyr")
     ZP_PLATFORM = "zephyr"
+    SRC_FILTER = _platform_src_filter(ZP_PLATFORM)
```
steils, line 95: If the user passes `-DZP_PLATFORM=...`, line 165 avoids appending the framework default for CMake. However, `SRC_FILTER` has already been taken above from the framework default, so PlatformIO and CMake can still select different platform sources.

To fix it we probably should parse `cmake_extra_args_list` before selecting `SRC_FILTER`, get explicit `-DZP_PLATFORM=` value (if present), and use that value platform for `SRC_FILTER`. Use the framework default only when no explicit platform was passed.

extra_script.py
```diff
@@ -62,28 +110,29 @@ def _platform_src_filter(system_dir):
...
-    SRC_FILTER = _platform_src_filter("system/mbed")
     ZP_PLATFORM = "mbed"
+    SRC_FILTER = _platform_src_filter(ZP_PLATFORM)
     CPPDEFINES = [
         "ZENOH_MBED",
         "ZENOH_C_STANDARD=99",
     ]
+
 elif FRAMEWORK == 'generic':
     SRC_FILTER = BASE_SRC_FILTER
```
steils, line 137: `generic` now uses BASE_SRC_FILTER directly, but BASE_SRC_FILTER now excludes `link/transport/*/**`. Before this PR, generic framework kept all transport backend files.
Maybe generic should also read explicit `-DZP_PLATFORM=...` (if present) from `build.cmake_extra_args` and use that platform source list?

extra_script.py
```diff
@@ -22,16 +24,62 @@
...
+        )
+        return []
+
+    with open(platform_file, "r", encoding="utf-8") as f:
+        content = f.read()
+
+    filters = []
+    seen = set()
+
+    for match in re.finditer(r'"\$\{PROJECT_SOURCE_DIR\}/src/([^"]+)"', content):
```
steils, line 70: This reads every `${PROJECT_SOURCE_DIR}/src/...` path from the file, but it does not handle CMake conditions. For example, files inside `if(ZP_UDP_MULTICAST_ENABLED)` are always added to the PlatformIO filter, even when CMake would skip them.
I'm not sure how to fix it. Maybe make CMake export the final platform source list, and use that source list instead of reading cmake files with regex?
The current source files have compile-time guards, so this may be okay today. However the result is not fully the same CMake' source list.

extra_script.py
```diff
@@ -134,5 +183,6 @@ def _has_cmake_arg(name):
     build_env.Prepend(CPPPATH=[generated_include_dir])
     build_env.Prepend(CCFLAGS=[f"-I{generated_include_dir}"])
 # Run the CMake command with the source and binary directories
+print("PlatformIO SRC_FILTER:", SRC_FILTER)
```
steils, line 186: Isn't this a debug print? Are you sure it would be not too long to print it on every PlatformIO build?

steils (changes requested):

gmartin82:
I don’t have capacity to work on this further. From what I can see, this is a consequence of the platform redesign in #1188, so I think the proper fix should be scheduled as follow-up work to align the PlatformIO integration with the new platform structure.

In the short term, pinning the older `platformio` package should restore CI, but I’d treat that as a temporary workaround rather than the real fix.

gmartin82:
I’ve opened #1219 as a narrower short-term workaround that pins PlatformIO Core to `6.1.18` in the ESP-IDF CI workflow.

That restores the CI behaviour prior to the recent PlatformIO Core `6.1.19` release, without trying to solve the broader PlatformIO/CMake source-selection issue in this PR.

I’m going to close this PR in favour of #1219. The proper fix should still be handled separately as follow-up work to align the PlatformIO integration with the platform restructuring from #1188.

PR #1220, opened 2026-05-12
Reconnect to the last endpoint without restarting tasks

### LINKED ISSUE #1005: The rmw_zenoh on Arduino on ESP32 in unicast of client mode does not reconnect
### Describe the release item

Description:
Currently, I am developing middleware that utilizes zenoh-pico in client mode on arduino on esp32.

I encountered a problem during development.
In rmw_zenoh running on Arduino/ESP32 in client mode, when using unicast, it does not automatically reconnect after the Zenoh server is restarted.

I believe that rmw_zenoh_pico generates two tasks: a read task and a lease task.
When the connected peer terminates, zenoh-pico calls _z_task_detach() on the lease task.
Then, the lease task calls _z_reopen() after returning from _z_task_detach() in _zp_unicast_send_keep_alive().

In environments where POSIX is available, pthread_detach() in _z_task_detach() is called on the lease task.
However, since Arduino (and FreeRTOS) do not implement pthread_detach(), a custom function has been implemented.

The implementation in zenoh-pico/src/system/arduino/esp32/system.c since version 1.4.0 is as follows:

```
z_result_t _z_task_detach(_z_task_t *task) {
    // Note: task/thread detach not supported on FreeRTOS API, so we force its deletion instead.
    return _z_task_cancel(task);
}

z_result_t _z_task_cancel(_z_task_t *task) {
    vTaskDelete(task->handle);
    task->handle = NULL;
    return _Z_RES_OK;
}
```

Issue:
In this implementation, the lease task is terminated by vTaskDelete() before _z_reopen() is executed.
As a result, the zenoh-pico task cannot reconnect to the peer Zenoh server, even if it was connected before termination.

### LINKED ISSUE #1053: [Bug] Reconnection on ThreadX causes crash due to memset()
### Describe the bug

ThreadX's threads, mutexes etc contain pointers to the next and previous thread, mutex etc, forming a linked list. When Zenoh Pico attempts to reconnect it uses memset() to zero these structs rather than deleting them properly. This leads to ThreadX attempting to dereference zeroed memory when going through the linked list, causing the program to crash. An example of this is at the start of the function `_z_unicast_transport_create`.

### To reproduce

1. Run any ThreadX example with `Z_FEATURE_AUTO_RECONNECT` = 1
2. Start an instance of `zenohd`
3. Stop `zenohd`
4. Start `zenohd` again
5. Zenoh pico will crash due to ThreadX attempting to dereference uninitialised memory

### System info

- Platform: STM32H573, ThreadX, NetXduo

steils:
Avoid restarting transport tasks during client reconnect.
Reuse the existing client transport and try the last successful endpoint first.
If that fails, continue with the configured connect locators or scouted locators. Restore the connection without replacing the transport object.

Closes: https://github.com/eclipse-zenoh/zenoh-pico/issues/1005
Closes: https://github.com/eclipse-zenoh/zenoh-pico/issues/1053

---
## 🏷️ Label-Based Checklist

**Labels:** `enhancement`

## ✨ Enhancement Requirements

**Instructions:**

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -492,8 +492,11 @@ jobs:
 
   connection_restore_test:
     needs: zenoh_build
-    name: Connection restore test
+    name: Connection restore test (multi-thread=${{ matrix.multi_thread }})
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        multi_thread: [1, 0]
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
@@ -510,8 +513,14 @@ jobs:
         run: |
           sudo apt install -y ninja-build
           CMAKE_GENERATOR=Ninja ASAN=ON CMAKE_BUILD_TYPE=Debug ZENOH_DEBUG=3 make
-          RUST_LOG=debug sudo python3 ./build/tests/connection_restore.py ./zenoh-standalone/zenohd
+          if [ "$Z_FEATURE_MULTI_THREAD" = "0" ]; then
+            RUST_LOG=debug sudo python3 ./build/tests/connection_restore.py ./zenoh-standalone/zenohd --single-thread
+          else
+            RUST_LOG=debug sudo python3 ./build/tests/connection_restore.py ./zenoh-standalone/zenohd
+          fi
         timeout-minutes: 20
+        env:
+          Z_FEATURE_MULTI_THREAD: ${{ matrix.multi_thread }}
 
   routed_peer_client_test:
     needs: zenoh_build
--- a/src/net/session.c
+++ b/src/net/session.c
@@ -31,6 +31,7 @@
 #include "zenoh-pico/transport/common/lease.h"
 #include "zenoh-pico/transport/common/read.h"
 #include "zenoh-pico/transport/common/tx.h"
+#include "zenoh-pico/transport/manager.h"
 #include "zenoh-pico/transport/multicast.h"
 #include "zenoh-pico/transport/multicast/lease.h"
 #include "zenoh-pico/transport/multicast/read.h"
@@ -39,6 +40,7 @@
 #include "zenoh-pico/transport/unicast.h"
 #include "zenoh-pico/transport/unicast/lease.h"
 #include "zenoh-pico/transport/unicast/read.h"
+#include "zenoh-pico/transport/unicast/transport.h"
 #include "zenoh-pico/utils/config.h"
 #include "zenoh-pico/utils/logging.h"
 #include "zenoh-pico/utils/result.h"
@@ -130,6 +132,34 @@ static z_result_t _z_config_get_mode(const _z_config_t *config, z_whatami_t *mod
     return ret;
 }
 
+typedef enum {
+    _Z_OPEN_CONNECT_DEFAULT,
+#if Z_FEATURE_AUTO_RECONNECT == 1
+    _Z_OPEN_CONNECT_CLIENT_UNICAST_RECONNECT,
+#endif
+} _z_open_connect_mode_t;
+
+typedef struct {
+    const _z_string_t *_preferred_locator;
+    _z_open_connect_mode_t _connect_mode;
+#if Z_FEATURE_AUTO_RECONNECT == 1
+    _z_transport_unicast_t *_reconnect_unicast;
+#endif
+} _z_open_options_t;
+
+#if Z_FEATURE_AUTO_RECONNECT == 1
+static void _z_session_update_last_connect_locator(_z_session_t *s, const _z_string_t *locator) {
+    _z_string_t copy = _z_string_null();
+    if (_z_string_copy(&copy, locator) != _Z_RES_OK) {
+        _Z_DEBUG("Failed to update last successful connect locator");
+        _z_string_clear(&s->_last_connect_locator);
+        return;
+    }
+    _z_string_clear(&s->_last_connect_locator);
+    (void)_z_string_move(&s->_last_connect_locator, &copy);
+}
+#endif
+
 static z_result_t _z_open_inner(_z_session_rc_t *zs, _z_string_t *locator, const _z_id_t *zid, int peer_op,
                                 const _z_config_t *config) {
     z_result_t ret = _Z_RES_OK;
@@ -141,12 +171,87 @@ static z_result_t _z_open_inner(_z_session_rc_t *zs, _z_string_t *locator, const
     }
     _z_transport_get_common(&zn->_tp)->_session = _z_session_rc_clone_as_weak(zs);
     _z_transport_get_common(&zn->_tp)->_state = _Z_TRANSPORT_STATE_OPEN;
+#if Z_FEATURE_AUTO_RECONNECT == 1
+    if ((zn->_mode == Z_WHATAMI_CLIENT) && (peer_op == _Z_PEER_OP_OPEN)) {
+        _z_session_update_last_connect_locator(zn, locator);
+    }
+#endif
 #if Z_FEATURE_MULTICAST_DECLARATIONS == 1
     if (zn->_tp._type == _Z_TRANSPORT_MULTICAST_TYPE) {
         ret = _z_interest_pull_resource_from_peers(zn);
     }
 #endif
     return ret;
+}
+
+static void _z_open_options_default(_z_open_options_t *opts) {
+    opts->_preferred_locator = NULL;
+    opts->_connect_mode = _Z_OPEN_CONNECT_DEFAULT;
+#if Z_FEATURE_AUTO_RECONNECT == 1
+    opts->_reconnect_unicast = NULL;
+#endif
+}
+
+static bool _z_open_has_preferred_locator(const _z_open_options_t *opts) {
+    return (opts != NULL) && (opts->_preferred_locator != NULL) && _z_string_check(opts->_preferred_locator);
+}
+
+static z_result_t _z_open_connect_single_locator(_z_session_rc_t *zs, _z_string_t *locator, const _z_id_t *zid,
+                                                 const _z_config_t *config, const _z_open_options_t *opts) {
+    switch (opts->_connect_mode) {
+        case _Z_OPEN_CONNECT_DEFAULT:
+            return _z_open_inner(zs, locator, zid, _Z_PEER_OP_OPEN, config);
+#if Z_FEATURE_AUTO_RECONNECT == 1
+        case _Z_OPEN_CONNECT_CLIENT_UNICAST_RECONNECT: {
+            if (opts->_reconnect_unicast == NULL) {
+                return _Z_ERR_GENERIC;
+            }
+            z_result_t ret = _z_reopen_client_unicast_transport(opts->_reconnect_unicast, zid, locator, config);
+            if (ret != _Z_RES_OK) {
+                return ret;
+            }
+            _z_session_update_last_connect_locator(_Z_RC_IN_VAL(zs), locator);
+            return _Z_RES_OK;
+        }
+#endif
+        default:
+            return _Z_ERR_GENERIC;
+    }
+}
+
+static z_result_t _z_open_append_locator_copy(_z_string_svec_t *dst, const _z_string_t *locator) {
+    _z_string_t copy = _z_string_null();
+    _Z_RETURN_IF_ERR(_z_string_copy(&copy, locator));
+    z_result_t ret = _z_string_svec_append(dst, &copy, false);
+    if (ret != _Z_RES_OK) {
+        _z_string_clear(&copy);
+    }
+    return ret;
+}
+
+static z_result_t _z_open_order_connect_locators(_z_string_svec_t *ordered_locators,
+                                                 const _z_string_svec_t *connect_locators,
+                                                 const _z_open_options_t *opts) {
+    *ordered_locators = _z_string_svec_null();
+    const _z_string_t *preferred_locator = NULL;
+    if (_z_open_has_preferred_locator(opts)) {
+        preferred_locator = opts->_preferred_locator;
+        _Z_RETURN_IF_ERR(_z_open_append_locator_copy(ordered_locators, preferred_locator));
+    }
+
+    size_t connect_len = _z_string_svec_len(connect_locators);
+    for (size_t i = 0; i < connect_len; i++) {
+        _z_string_t *locator = _z_string_svec_get(connect_locators, i);
+        if ((preferred_locator != NULL) && _z_string_equals(locator, preferred_locator)) {
+            continue;
+        }
+        z_result_t ret = _z_open_append_locator_copy(ordered_locators, locator);
+        if (ret != _Z_RES_OK) {
+            _z_string_svec_clear(ordered_locators);
+            return ret;
+        }
+    }
+    return _Z_RES_OK;
 }
 
 static int32_t _z_get_remaining_timeout_ms(z_clock_t *start, int32_t timeout_ms) {
@@ -188,10 +293,13 @@ typedef struct {
  */
 static z_result_t _z_open_connect_locator(_z_session_rc_t *zn, _z_pending_peers_t *pending_peers, const _z_id_t *zid,
                                           _z_config_t *config, int32_t timeout_ms, bool exit_on_non_retryable_failure,
-                                          _z_open_connect_result_t *out) {
+                                          const _z_open_options_t *opts, _z_open_connect_result_t *out) {
     size_t connect_len = _z_pending_peer_svec_len(&pending_peers->_peers);
     z_result_t last_retryable_ret = _Z_ERR_TRANSPORT_OPEN_FAILED;
     z_result_t last_non_retryable_ret = _Z_RES_OK;
+    if (opts == NULL) {
+        return _Z_ERR_GENERIC;
+    }
 
     out->transport_opened = false;
     out->remaining_timeout_ms = timeout_ms;
@@ -215,7 +323,7 @@ static z_result_t _z_open_connect_locator(_z_session_rc_t *zn, _z_pending_peers_
 
             _z_string_t *locator = &peer->_locator;
 
-            ret = _z_open_inner(zn, locator, zid, _Z_PEER_OP_OPEN, config);
+            ret = _z_open_connect_single_locator(zn, locator, zid, config, opts);
             if (ret == _Z_RES_OK) {
                 out->transport_opened = true;
                 peer->_state = _Z_PENDING_PEER_STATE_DONE;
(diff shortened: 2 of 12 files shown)


.github/workflows/ci.yml
```diff
@@ -352,9 +352,10 @@ jobs:
       - name: Build project and run tests
         run: |
           sudo apt install -y ninja-build
-          CMAKE_GENERATOR=Ninja ASAN=ON make BUILD_TYPE=Debug
+          CMAKE_GENERATOR=Ninja ASAN=ON ZENOH_DEBUG=3 make BUILD_TYPE=Debug
```
gmartin82: Was this an intentional change or part of debugging?

.github/workflows/ci.yml
```diff
@@ -352,9 +352,10 @@ jobs:
...
           ninja -C build/ test
           python3 ./build/tests/single_thread.py
+          python3 -u ./build/tests/connection_restore.py ./zenoh-standalone/zenohd --single-thread
```
gmartin82: As we have a dedicated `connection_restore_test` CI job would it be better having a single thread version of that test instead?

tests/single_thread.py
```diff
@@ -6,6 +6,12 @@
 # Specify the directory for the binaries
 DIR_EXAMPLES = "build/examples"
 
+
+def filter_debug_logs(output):
```
gmartin82: All changes in this file would not be needed if a single threaded version of the connection restore test CI job was added instead instead.

src/transport/transport.c
```diff
@@ -109,7 +109,7 @@ void _z_transport_free(_z_transport_t **zt) {
 #if Z_FEATURE_BATCHING == 1
 z_result_t _z_transport_start_batching(_z_transport_t *zt) {
     _z_transport_common_t *ztc = _z_transport_get_common(zt);
-    if (ztc == NULL) {
+    if ((ztc == NULL) || (ztc->_state != _Z_TRANSPORT_STATE_OPEN) || (ztc->_link == NULL)) {
```
gmartin82: This `_link == NULL` check looks unnecessary to me.

My understanding is that `_state` should be the authoritative transport availability state. If a transport is `_Z_TRANSPORT_STATE_OPEN`, then it should have a valid link. If the link is cleared/freed/moved during reconnect or close, the state should first be changed to `RECONNECTING` or `CLOSED`.

If `OPEN && _link == NULL` is possible, that seems like a lifecycle bug rather than a normal condition the send path should need to handle. Could we rely on `_state` here and make sure the reconnect/clear/replace paths maintain that invariant?

src/transport/transport.c
```diff
@@ -133,14 +133,19 @@ z_result_t _z_transport_stop_batching(_z_transport_t *zt) {
         _Z_ERROR_RETURN(_Z_ERR_TRANSPORT_NOT_AVAILABLE);
     }
 
+    bool is_available = (ztc->_state == _Z_TRANSPORT_STATE_OPEN) && (ztc->_link != NULL);
```
gmartin82: See previous comment about `_state` and `_link`.

src/transport/transport.c
```diff
@@ -133,14 +133,19 @@ z_result_t _z_transport_stop_batching(_z_transport_t *zt) {
...
+    if (ztc->_batch_state != _Z_BATCHING_ACTIVE) {
+        return is_available ? _Z_RES_OK : _Z_ERR_TRANSPORT_NOT_AVAILABLE;
```
gmartin82: Was this change necessary?

`stop_batching()` looks more like cleanup/unwind logic than a transport availability check. If `start_batching()` succeeded, I would expect `stop_batching()` to reliably release the batching state/mutexes even if the transport has since moved to `RECONNECTING` or `CLOSED`.

I can see the value in giving feedback that something has gone wrong with the transport, but is this the right place to do that?

src/transport/transport.c
```diff
@@ -133,14 +133,19 @@ z_result_t _z_transport_stop_batching(_z_transport_t *zt) {
...
+
 #if Z_FEATURE_BATCH_TX_MUTEX == 1
     _z_transport_tx_mutex_unlock(ztc);
 #endif
 #if Z_FEATURE_BATCH_PEER_MUTEX == 1
     _z_transport_peer_mutex_unlock(ztc);
 #endif
     ztc->_batch_state = _Z_BATCHING_IDLE;
-    return _Z_RES_OK;
+    return is_available ? _Z_RES_OK : _Z_ERR_TRANSPORT_NOT_AVAILABLE;
```
gmartin82: See previous comment.

src/net/session.c
```diff
@@ -542,6 +559,116 @@ z_result_t _z_open(_z_session_rc_t *zn, _z_config_t *config, const _z_id_t *zid)
...
+    if (ret != _Z_RES_OK) {
+        _z_string_svec_clear(candidates);
+        return ret;
+    }
+    return _z_string_svec_is_empty(candidates) ? _Z_ERR_CONFIG_LOCATOR_INVALID : _Z_RES_OK;
+}
+
+static z_result_t _z_client_reopen_unicast(_z_session_t *s) {
+    _z_string_svec_t candidates = _z_string_svec_null();
+    _Z_RETURN_IF_ERR(_z_client_reconnect_candidates(s, &candidates));
```
gmartin82: I’m going to pause detailed review here because I have a broader concern about the approach.

I appreciate this change was introduced while you were already working on this PR, but the client connect path was recently refactored so that configured connect locators are handled consistently as a list of alternatives. That gives us one place for locator ordering, retryable vs non-retryable errors, timeout behaviour, configured locators, and scouting fallback semantics.

This PR appears to introduce a separate client reconnection path, with its own locator candidate construction and connection establishment logic. I’m concerned that initial connect and reconnect could drift semantically over time.

I think reconnect should build on the existing client connect/open path rather than adding a second implementation. The “try the last successful locator first” behaviour feels like an extension of the existing locator ordering logic, not a separate connection establishment model.

Could we adapt the existing client connect/open path to accept an optional preferred reconnect locator, try that first while avoiding duplicates, and then fall back to the normal configured/scouted alternatives? The reconnect-specific part should ideally be limited to what happens after a successful connection: for client unicast reconnect, install the new connection into the existing transport object rather than replacing/reinitialising the whole transport.

Given that this affects the structure of the change, I think it would be better to resolve this design question before reviewing the rest of the implementation in detail.

gmartin82 (changes requested):

src/net/session.c
```diff
@@ -555,47 +682,65 @@ _z_fut_fn_result_t _z_client_reopen_task_fn(void *ztc_arg, _z_executor_t *execut
...
     if (_z_config_is_empty(&s->_config)) {
         _z_session_rc_drop(&zs);
         return _z_fut_fn_result_ready();
     }
+
     _z_session_transport_mutex_lock(s);
-    z_result_t ret = _z_open(&zs, &s->_config, &s->_local_zid);
-    _z_session_transport_mutex_unlock(s);
+    z_result_t ret =
+        s->_mode == Z_WHATAMI_CLIENT ? _z_client_reopen_unicast(s) : _z_open(&zs, &s->_config, &s->_local_zid);
```
gmartin82: Minor point: you're tying client reconnect to unicast here.

That may be fine for the currently supported behaviour, since `_z_multicast_open_client()` returns `_Z_ERR_CONFIG_UNSUPPORTED_CLIENT_MULTICAST`, but `_z_new_transport_client()` still has explicit handling for `Z_LINK_CAP_TRANSPORT_RAWETH` and `Z_LINK_CAP_TRANSPORT_MULTICAST`.

Could we either make the unicast-only assumption explicit here, or ensure the reconnect path will not become a trap if client multicast/raweth support is implemented later?

Vuk-SFL:
Does this change make https://github.com/eclipse-zenoh/zenoh-pico/pull/1205 unnecessary since that PR is about adressing tx resource usage while reconenction is active?

steils:
@Vuk-SFL 

> Does this change make #1205 unnecessary since that PR is about adressing tx resource usage while reconenction is active?

For the reconnect case is covered in this PR, I think yes, it's no longer needed. But as I see #1205 touches more general clear paths, so that still should be relevant after this PR.

src/net/session.c
```diff
@@ -466,6 +555,74 @@ z_result_t _z_open_locators(_z_session_rc_t *zn, const _z_string_svec_t *listen_
...
+    _Z_RC_IN_VAL(zn)->_mode = mode;
+
+    if ((_z_string_svec_len(&listen_locators) > 0) || (_z_string_svec_len(&connect_locators) > 0)) {
+        ret = _z_open_locators(zn, &listen_locators, &connect_locators, zid, config, mode, opts);
+        _z_string_svec_clear(&listen_locators);
+        _z_string_svec_clear(&connect_locators);
+        return ret;
+    }
+
+    bool fallback_to_scout = true;
```
gmartin82: I think this could be clearer.

My understanding is that this covers the reconnect case where the initial open used a scouted locator, so there are no configured connect locators to try, but we may still have a last successful locator stored.

Could we add a comment explaining that? I also think this block and the following `fallback_to_scout` check could be condensed into a single block, avoiding the extra boolean.

src/net/session.c
```diff
@@ -466,6 +555,74 @@ z_result_t _z_open_locators(_z_session_rc_t *zn, const _z_string_svec_t *listen_
     }
 }
 
+static z_result_t _z_open_common(_z_session_rc_t *zn, _z_config_t *config, const _z_id_t *zid,
```
gmartin82, line 584: One broader concern here is that this changes the structure of existing code more than seems necessary for the behaviour being added.

This is a shared codebase, so I think we should avoid unnecessary restructuring based on personal style preference. It makes the change harder to review, especially in C code where cleanup and ownership paths are important.

This also introduces quite a lot of duplicated cleanup, which makes the control flow harder to maintain.

src/net/session.c
```diff
@@ -130,6 +131,27 @@ static z_result_t _z_config_get_mode(const _z_config_t *config, z_whatami_t *mod
...
+                                         const _z_config_t *config, void *ctx);
+
+typedef struct {
+    const _z_string_t *_preferred_locator;
+    _z_open_connect_fn _open_connect;
+    void *_open_connect_ctx;
+} _z_open_options_t;
+
+#if Z_FEATURE_AUTO_RECONNECT == 1
+static void _z_session_update_last_connect_locator(_z_session_t *s, const _z_string_t *locator) {
```
gmartin82, line 151: If `_z_string_copy()` fails here, the failure is silent from the caller’s perspective and the previous `_last_connect_locator` is retained.

That means a successful connection may not update the stored “last successful” locator, and a future reconnect may prefer a stale locator. If this state is only best-effort, should we clear `_last_connect_locator` on failure instead of leaving the old value in place?

src/net/session.c
```diff
@@ -188,10 +266,15 @@ typedef struct {
...
 static z_result_t _z_open_connect_locator(_z_session_rc_t *zn, _z_pending_peers_t *pending_peers, const _z_id_t *zid,
                                           _z_config_t *config, int32_t timeout_ms, bool exit_on_non_retryable_failure,
-                                          _z_open_connect_result_t *out) {
+                                          const _z_open_options_t *opts, _z_open_connect_result_t *out) {
     size_t connect_len = _z_pending_peer_svec_len(&pending_peers->_peers);
     z_result_t last_retryable_ret = _Z_ERR_TRANSPORT_OPEN_FAILED;
     z_result_t last_non_retryable_ret = _Z_RES_OK;
+    _z_open_connect_fn open_connect = (opts != NULL) ? opts->_open_connect : NULL;
+    void *open_connect_ctx = (opts != NULL) ? opts->_open_connect_ctx : NULL;
+    if (open_connect == NULL) {
```
gmartin82: This fallback feels wrong to me.

We have `_z_open_options_default()`, so the expected pattern should be: use default options when `opts == NULL`, otherwise assume the provided options are valid. If a caller provides `_z_open_options_t`, it should have been initialised correctly, usually by starting from `_z_open_options_default()` and then overriding the required fields.

Silently falling back to `_z_open_connect_default` here risks hiding a partially initialised or invalid `opts` value.

src/net/session.c
```diff
@@ -392,11 +480,12 @@ static inline const char *_z_open_connect_exit_on_failure_default(z_whatami_t mo
...
 z_result_t _z_open_locators(_z_session_rc_t *zn, const _z_string_svec_t *listen_locators,
                             const _z_string_svec_t *connect_locators, const _z_id_t *zid, _z_config_t *config,
-                            z_whatami_t mode) {
+                            z_whatami_t mode, const _z_open_options_t *opts) {
     size_t listen_len = _z_string_svec_len(listen_locators);
     size_t connect_len = _z_string_svec_len(connect_locators);
+    bool has_preferred_locator = (mode == Z_WHATAMI_CLIENT) && _z_open_has_preferred_locator(opts);
 
-    if ((listen_len == 0) && (connect_len == 0)) {
+    if ((listen_len == 0) && (connect_len == 0) && !has_preferred_locator) {
```
gmartin82: This condition feels wrong to me.

This path is for opening configured locators. If there are no configured listen or connect locators, this function should not be treating a preferred reconnect locator hidden in `opts` as if it were part of the configured locator set.

In that case, `_z_open_common()` should be taking the scouting path. If reconnect needs to try the last successful locator before scouting, that should be handled explicitly in the scouting/reconnect flow rather than weakening the contract of `_z_open_locators()`.

src/net/session.c
```diff
@@ -508,40 +665,54 @@ z_result_t _z_open_locators(_z_session_rc_t *zn, const _z_string_svec_t *listen_
...
+    _z_client_reopen_unicast_ctx_t *ctx = (_z_client_reopen_unicast_ctx_t *)ctx_arg;
+    _z_session_t *s = _Z_RC_IN_VAL(zs);
+    _z_transport_t next = {0};
+
+    z_result_t ret = _z_new_transport(&next, zid, locator, Z_WHATAMI_CLIENT, _Z_PEER_OP_OPEN, config, &s->_runtime);
+    if (ret != _Z_RES_OK) {
+        _z_transport_clear(&next);
+        return ret;
+    }
+    if (next._type != _Z_TRANSPORT_UNICAST_TYPE) {
```
gmartin82: This feels like the type check is happening too late.

`_z_client_reopen_unicast_connection()` can only transplant unicast connection state into an existing unicast transport, but it calls the generic `_z_new_transport()` first and only then rejects non-unicast transports. That means we may open/create a transport needlessly before discovering this reconnect path cannot use it.

Should we either reject incompatible locators before opening, use a unicast-specific open helper, or record enough information with the last successful locator to know whether it is valid for this in-place unicast reconnect path?

src/net/session.c
```diff
@@ -508,40 +665,54 @@ z_result_t _z_open_locators(_z_session_rc_t *zn, const _z_string_svec_t *listen_
...
+
+static z_result_t _z_client_reopen_unicast(_z_session_rc_t *zs) {
+    _z_session_t *s = _Z_RC_IN_VAL(zs);
+    _z_client_reopen_unicast_ctx_t ctx = {0};
+    ctx._target = &s->_tp._transport._unicast;
+
+    _z_open_options_t opts;
+    _z_open_options_default(&opts);
+    opts._preferred_locator = &s->_last_connect_locator;
+    opts._open_connect = _z_client_reopen_unicast_connection;
```
gmartin82: I’m not sure this needs to be modelled as a function pointer in `_z_open_options_t`.

This is all internal code, so if normal open and reconnect need different entry points, could they be regular functions which call into the shared common code directly? Hiding the reconnect-specific open/install behaviour behind an options callback makes the control flow harder to follow.

src/net/session.c
```diff
@@ -555,47 +726,66 @@ _z_fut_fn_result_t _z_client_reopen_task_fn(void *ztc_arg, _z_executor_t *execut
...
+                } else {
+                    _z_transport_clear(&s->_tp);
+                    tc->_session = _z_session_rc_clone_as_weak(&zs);
+                    tc->_state = _Z_TRANSPORT_STATE_RECONNECTING;
+                }
+                tc->_tasks = tasks_handles;
+                _z_session_transport_mutex_unlock(s);
                 _z_session_rc_drop(&zs);
-                return _z_fut_fn_result_continue();
+                return _z_fut_fn_result_wake_up_after(1000);
```
gmartin82: This looks like an unrelated change in behaviour.

A delayed retry may make sense if declaration replay fails, but previously this path returned `continue()` and now it waits 1 second before retrying. If we want that behaviour change, I think it should be made explicit and avoid using a magic number here.

src/net/session.c
```diff
@@ -555,47 +726,66 @@ _z_fut_fn_result_t _z_client_reopen_task_fn(void *ztc_arg, _z_executor_t *execut
...
     _z_transport_tasks_t tasks_handles = tc->_tasks;
     _z_session_rc_t zs = _z_session_weak_upgrade(&tc->_session);  // should not fail
+    if (_Z_RC_IS_NULL(&zs)) {
+        return _z_fut_fn_result_ready();
+    }
     _z_session_t *s = _Z_RC_IN_VAL(&zs);
-    _z_session_weak_drop(&tc->_session);
 
+    _z_session_transport_mutex_lock(s);
+    bool in_place_reconnect = (s->_mode == Z_WHATAMI_CLIENT) && (s->_tp._type == _Z_TRANSPORT_UNICAST_TYPE);
```
gmartin82: I’m not keen on this transport-specific check at the session/reopen-task level.

Ideally this layer should not need to special-case client unicast. It would be cleaner to ask the transport layer whether the current transport supports in-place reconnect, and let the lower-level transport code handle the transport-specific details.

This kind of branching makes the transport abstraction harder to evolve.

gmartin82 (changes requested):

src/net/session.c
```diff
@@ -508,40 +665,54 @@ z_result_t _z_open_locators(_z_session_rc_t *zn, const _z_string_svec_t *listen_
...
+#if Z_FEATURE_AUTO_RECONNECT == 1
+typedef struct {
+    _z_transport_unicast_t *_target;
+} _z_client_reopen_unicast_ctx_t;
+
+static z_result_t _z_client_reopen_unicast_connection(_z_session_rc_t *zs, _z_string_t *locator, const _z_id_t *zid,
+                                                      const _z_config_t *config, void *ctx_arg) {
+    _z_client_reopen_unicast_ctx_t *ctx = (_z_client_reopen_unicast_ctx_t *)ctx_arg;
+    _z_session_t *s = _Z_RC_IN_VAL(zs);
+    _z_transport_t next = {0};
```
gmartin82: `next` should be explicitly initialised as `_Z_TRANSPORT_NONE` before calling `_z_new_transport()`.

As written, `_z_transport_t next = {0}` makes `_type` equal to `_Z_TRANSPORT_UNICAST_TYPE`, because unicast is enum value 0. If `_z_new_transport()` fails before setting a valid transport type, `_z_transport_clear(&next)` can treat the zero-initialised stack object as a unicast transport and clear invalid/uninitialised state.

src/net/session.c
steils: Fixed.

src/net/session.c
steils: Agreed about the duplicated cleanup. I would prefer not to restore the deeply nested code from the previous version. Since this code already had to be moved into `_z_open_common()`, I took the opportunity to flatten the control flow a bit, and lessen the number of indentation levels, without changing the behaviour.
I agree that the duplicated cleanup made that harder to review and maintain, so I changed it to use a single `exit` cleanup block.

src/net/session.c
steils: Fixed.

src/net/session.c
steils: Fixed.

src/net/session.c
steils: Fixed.

src/net/session.c
steils: Fixed.

src/net/session.c
steils: Fixed.

src/net/session.c
steils: Right. Forgot from a debug version. Fixed.

src/net/session.c
steils: Made a simple helper on the transport level. Hope it's fixed.

src/net/session.c
steils: Fixed.

PR #1221, opened 2026-05-12
Ensure generated timestamps are session-monotonic

### LINKED ISSUE #1102: Ensure z_timestamp_new returns unique (and monotonically increasing) timestamp values
### Describe the feature

Currently z_timestamp_new use gettimeofday function on unix systems (and probably even less precise functions for other platforms). For certain hardware this is not precise enough to guarantee unique timestamps - which leads to message loss for advanced_pub_sub when publishing at high frequency, since advanced subscriber retains at most one message per timestamp value when fetching history.

gmartin82:
## Description

Makes generated timestamps monotonic within a session, even when the platform clock returns the same value repeatedly or moves backwards slightly.

### What does this PR do?

- Tracks the last generated timestamp on each session.
- Updates `z_timestamp_new()` to bump repeated or older clock values by one NTP tick.
- Switches Unix timestamp retrieval from `gettimeofday()` to `clock_gettime(CLOCK_REALTIME)`.
- Adds timestamp tests for both real-clock behavior and deterministic repeated-clock behavior using `Z_TEST_HOOKS`.

Some platforms or clock sources can return the same timestamp for multiple rapid calls. This could produce duplicate timestamps from the same session. The session-local monotonic guard ensures each generated timestamp is strictly increasing.

### Related Issues

Closes #1102.

---
## 🏷️ Label-Based Checklist

**Labels:** `bug`

## 🐛 Bug Fix Requirements

**Instructions:**


src/system/unix/system.c
```diff
@@ -329,9 +329,11 @@
 }
 
 z_result_t _z_get_time_since_epoch(_z_time_since_epoch *t) {
-    z_time_t now;
-    gettimeofday(&now, NULL);
+    struct timespec now;
```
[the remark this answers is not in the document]
gmartin82: Switched to `clock_gettime` for nanosecond-level accuracy (up from microseconds). The Y2038 risk remains identical to the previous `gettimeofday` implementation on 32-bit systems, but the new interface provides better precision and retains compatibility across Unix environments.

src/system/unix/system.c
```diff
@@ -329,9 +329,11 @@
...
+    if (clock_gettime(CLOCK_REALTIME, &now) != 0) {
```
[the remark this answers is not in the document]
gmartin82: See above.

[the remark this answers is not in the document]
gmartin82: False positive as `clock_gettime` is included via `time.h`.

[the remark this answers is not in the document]
gmartin82: See above.

src/api/api.c
```diff
@@ -403,13 +403,48 @@ z_result_t z_bytes_writer_append(z_loaned_bytes_writer_t *writer, z_moved_bytes_
...
+        _Z_RETURN_IF_ERR(_z_get_time_since_epoch(&t));
+    }
+#else
     _Z_RETURN_IF_ERR(_z_get_time_since_epoch(&t));
+#endif
+
+    _z_session_t *s = _Z_RC_IN_VAL(zs);
+    _z_ntp64_t time = _z_timestamp_ntp64_from_time(t.secs, t.nanos);
+    _Z_RETURN_IF_ERR(_z_session_mutex_lock(s));
+    if (time <= s->_last_timestamp) {
```
DenisBiryukov91: I'm not sure we should consider < case, since clock can not normally go backward, unless user messed with it himself, and in this case - we can not do a lot here (except probably issuing a warning)

src/api/api.c
```diff
@@ -403,13 +403,48 @@ z_result_t z_bytes_writer_append(z_loaned_bytes_writer_t *writer, z_moved_bytes_
...
+    } else {
+        _Z_RETURN_IF_ERR(_z_get_time_since_epoch(&t));
+    }
+#else
     _Z_RETURN_IF_ERR(_z_get_time_since_epoch(&t));
+#endif
+
+    _z_session_t *s = _Z_RC_IN_VAL(zs);
+    _z_ntp64_t time = _z_timestamp_ntp64_from_time(t.secs, t.nanos);
+    _Z_RETURN_IF_ERR(_z_session_mutex_lock(s));
```
DenisBiryukov91: Could we rather do it using atomics, to improve performance and avoid possible deadlocks when z_timestamp_new is called within session-locked context ?

src/api/api.c
gmartin82: Zenoh uses `uhlc::HLC` for timestamp generation, and `uhlc` has essentially the same behaviour: it uses the physical clock value when it is greater than the last generated timestamp, otherwise it returns the previous timestamp + 1.

So I think zenoh-pico should keep the `<` handling. A system clock should not normally move backwards, but it can happen, so timestamp generation needs to handle it defensively. In that case, preserving session-local monotonicity seems preferable to returning a non-increasing timestamp or failing timestamp generation.

I agree that this behaviour needs to be documented clearly in `z_timestamp_new()`, so users understand that generated timestamps are wall-clock based but have a session-local monotonicity guard.

I’m less convinced we should print a warning every time this path is hit. If the system clock is adjusted backwards by a significant amount, every generated timestamp may hit this case until the clock catches up, which could flood logs. Zenoh/uhlc handles this as part of HLC behaviour without logging on every generated timestamp.

src/api/api.c
gmartin82: Good point, I'll change it to use an atomic.

src/protocol/core.c
```diff
@@ -32,6 +33,8 @@
 
 const _z_id_t empty_id = {0};
 
+_Z_ATOMIC_NUMERIC_IMPL(ntp64, _z_ntp64_t)
```
DenisBiryukov91: You can not define atomics this way, since it might break if zenoh-pico built using c-compiler is included inside cpp project (or project built by other c-compiler enabling different atomic implementation like pre- c11 gcc vs c11) since atomics are not size-compatible between different c and cpp compilers and normally sizeof(atomic<type>) != sizeof(type). This is the exact reason why only size_t atomic type was provided.

src/api/api.c
gmartin82: I looked into using atomics here, but `_z_ntp64_t` is a 64-bit value and zenoh-pico needs to work on platforms where native 64-bit atomics may not be available or may be expensive/problematic.

Instead of using the main session mutex, I’ve changed this to use a dedicated mutex that only protects `_last_timestamp`. That should avoid the possible deadlock/re-entrancy issue with session-locked contexts, while still keeping timestamp generation thread-safe.

I also did local throughput checks with and without the mutex, and I couldn’t measure a meaningful performance difference on my test machine.

src/api/api.c
```diff
@@ -403,13 +403,50 @@ z_result_t z_bytes_writer_append(z_loaned_bytes_writer_t *writer, z_moved_bytes_
...
+
+    _z_session_t *s = _Z_RC_IN_VAL(zs);
+    _z_ntp64_t time = _z_timestamp_ntp64_from_time(t.secs, t.nanos);
+    _Z_RETURN_IF_ERR(_z_session_last_timestamp_mutex_lock(s));
+    if (time > s->_last_timestamp) {
+        s->_last_timestamp = time;
+    } else {
+        if (s->_last_timestamp == UINT64_MAX) {
+            _z_session_last_timestamp_mutex_unlock(s);
+            _Z_ERROR_RETURN(_Z_ERR_GENERIC);
```
DenisBiryukov91: maybe it is worth creating a dedicated error-code for clock issue here ?

DenisBiryukov91 (approved):

src/api/api.c
gmartin82: Thanks Denis. I'll make the change before we merge.

gmartin82: I've added a dedicated `_Z_ERR_TIMESTAMP_GENERATION_FAILED` error code for this.

Issue #1217, opened 2026-05-05
[Bug] Broken build with esp-idf, missing include?


xopxe:
### Describe the bug

Building a platform.io project for an ESP32 board, using `platform = espressif32 @ 7.0.0`

* Espressif IoT Development Framework (ESP-IDF) 6.0.0
* zenoh-pico @ 1.9.0+sha.3b3ab65

I get the following error:

```
In file included from .pio/libdeps/pico32/zenoh-pico/src/link/transport/udp/udp_multicast_lwip_common.c:15:
.pio/libdeps/pico32/zenoh-pico/src/link/transport/udp/udp_multicast_lwip_common.h:25:69: error: 'struct sockaddr' declared inside parameter list will not be visible outside of this definition or declaration [-Werror]
   25 |                                                              struct sockaddr **lsockaddr);
      |                                                                     ^~~~~~~~
cc1: all warnings being treated as errors
```
Adding a `#include "lwip/sockets.h"` somewhere at the top fixes it.

### To reproduce

Use the fololwing platformio.ini
```
[env:pico32]
platform = espressif32 @ 7.0.0
board = pico32
framework = espidf

monitor_speed = 115200

board_build.cmake_extra_args = 
    -DZ_FEATURE_LINK_SERIAL=1
    -DZ_FEATURE_LINK_TCP=0 
    -DZ_FEATURE_LINK_UDP_UNICAST=0 
    -DZ_TRANSPORT_LEASE=60000 
    -DZ_TRANSPORT_LEASE_EXPIRE_FACTOR=2

build_flags =
    -DZENOH_C_STANDARD=99
    -DZENOH_COMPILER_GCC
    -Wno-missing-field-initializers
lib_deps =
    https://github.com/eclipse-zenoh/zenoh-pico
```

### System info

Ubuntu 26.04
PlatformioIDE 3.3.4
zenoh-pico from git main: 3b3ab65

xopxe:
I see the same problem with esp-idf 5.5:

framework-espidf @ 3.50500.0 (5.5.0) 
tool-xtensa-esp-elf-gdb @ 12.1.0+20221002 
toolchain-esp32ulp @ 1.23800.240113 (2.38.0) 
toolchain-xtensa-esp-elf @ 14.2.0+20241119
zenoh-pico @ 1.9.0+sha.6d91aec

xopxe:
[...]

Issue #1184, opened 2026-03-13
[Bug] Build error if Z_FEATURE_SESSION_CHECK is set as 0


giafranchini:
### Describe the bug

zenoh-pico fails to build when setting `Z_FEATURE_SESSION_CHECK` to 0. This is due to wrong usage of `sub` in:

https://github.com/eclipse-zenoh/zenoh-pico/blob/b8f785621a0d0bf5e3ab461843d3303045fab9bf/src/api/api.c#L2110

while should probably be:

```c
_z_session_t *zn = _z_session_weak_as_unsafe_ptr(&queryable->_zn);
```

and in:

https://github.com/eclipse-zenoh/zenoh-pico/blob/b8f785621a0d0bf5e3ab461843d3303045fab9bf/src/net/primitives.c#L392

while should probably be:

```c
_z_session_t *zn = _z_session_weak_as_unsafe_ptr(&qle->_zn);
```

### To reproduce

1. set `Z_FEATURE_SESSION_CHECK` to 0
2. build zenoh-pico

### System info

- Platform: Ubuntu 22.04
- Zenoh version: branch `main` on commit: b8f785621a0d0bf5e3ab461843d3303045fab9bf

Issue #1135, opened 2026-01-14
[Bug] Build errors when Z_FEATURE_BATCHING=0


ivis-Hyojun:
### Describe the bug

When `Z_FEATURE_BATCHING` is set to `0` (disabled), the project fails to build with undefined reference errors for batch mutex helper functions.

### To reproduce

`cmake -DZ_FEATURE_BATCHING=0 -DCMAKE_BUILD_TYPE=Debug -B build`
`cmake --build build`

```
[build] src/transport/common/tx.c: undefined reference to `_z_transport_batch_hold_tx_mutex'
[build] src/transport/common/tx.c: undefined reference to `_z_transport_batch_hold_peer_mutex'
[build] ld returned 1 exit status
```

### System info

OS: Windows/Linux/macOS (affects all platforms)
Compiler: GCC/Clang/MSVC (affects all compilers)
CMake version: Any
zenoh-pico version: Current main branch

PR #1225, opened 2026-05-19
Fix feature guards and warning for modular builds

DenisBiryukov91:
## Description

Fix feature guards and warning for modular builds

### What does this PR do?

Fix feature guards and warning for modular builds

To make build pass

### Related Issues

#1217
#1184 
#1135 

---
## 🏷️ Label-Based Checklist

**Labels:** `internal`

## 🏠 Internal Change

**Instructions:**

### DIFF
--- a/src/api/api.c
+++ b/src/api/api.c
@@ -2415,7 +2415,7 @@ const z_loaned_keyexpr_t *z_subscriber_keyexpr(const z_loaned_subscriber_t *sub)
     }
     _z_session_t *zn = _Z_RC_IN_VAL(&sess_rc);
 #else
-    _z_session_t *zn = _z_session_weak_as_unsafe_ptr(&sub->_zn);
+    _z_session_t *zn = _z_session_weak_as_unsafe_ptr((_z_session_weak_t *)&sub->_zn);
 #endif
     if (_z_session_mutex_lock_if_open(zn) != _Z_RES_OK) {
         _Z_WARN("Failed to lock session for subscriber keyexpr retrieval - session may be closing");
--- a/src/api/api.c
+++ b/src/api/api.c
@@ -2119,7 +2119,7 @@ const z_loaned_keyexpr_t *z_queryable_keyexpr(const z_loaned_queryable_t *querya
     }
     _z_session_t *zn = _Z_RC_IN_VAL(&sess_rc);
 #else
-    _z_session_t *zn = _z_session_weak_as_unsafe_ptr(&sub->_zn);
+    _z_session_t *zn = _z_session_weak_as_unsafe_ptr((_z_session_weak_t *)&queryable->_zn);
 #endif
     if (_z_session_mutex_lock_if_open(zn) != _Z_RES_OK) {
         _Z_WARN("Failed to lock session for queryable keyexpr retrieval - session may be closing");
--- a/src/link/transport/udp/udp_multicast_lwip_common.h
+++ b/src/link/transport/udp/udp_multicast_lwip_common.h
@@ -15,6 +15,7 @@
 #ifndef ZENOH_PICO_LINK_TRANSPORT_DATAGRAM_UDP_MULTICAST_LWIP_COMMON_H
 #define ZENOH_PICO_LINK_TRANSPORT_DATAGRAM_UDP_MULTICAST_LWIP_COMMON_H
 
+#include "lwip/sockets.h"
 #include "zenoh-pico/link/transport/udp_multicast.h"
 
 #ifdef __cplusplus
--- a/src/net/filtering.c
+++ b/src/net/filtering.c
@@ -309,7 +309,7 @@ z_result_t _z_write_filter_ctx_clear(_z_write_filter_ctx_t *ctx) {
 #endif
     return res;
 }
-#if Z_FEATURE_MATCHING == 1
+
 z_result_t _z_write_filter_clear(_z_write_filter_t *filter) {
     if (_Z_RC_IS_NULL(&filter->ctx)) {
         return _Z_RES_OK;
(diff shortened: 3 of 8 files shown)


src/link/transport/udp/udp_multicast_lwip_common.h
```diff
@@ -15,6 +15,7 @@
 #ifndef ZENOH_PICO_LINK_TRANSPORT_DATAGRAM_UDP_MULTICAST_LWIP_COMMON_H
 #define ZENOH_PICO_LINK_TRANSPORT_DATAGRAM_UDP_MULTICAST_LWIP_COMMON_H
 
+#include "lwip/sockets.h"
```
gmartin82, line 18: This seems to have caused a failure in the zephyr build.

src/session/interest.c
```diff
@@ -535,7 +535,9 @@ z_result_t _z_interest_process_interest(_z_session_t *zn, const _z_wireexpr_t *w
         return _Z_RES_OK;  // Nothing to do on unicast
     }
     // Push a join in case it's a new node
+#if Z_FEATURE_MULTICAST_TRANSPORT == 1
```
gmartin82, line 538: Should this guard include `Z_FEATURE_RAWETH_TRANSPORT` as well?

Raweth appears to use the same multicast join machinery: `_zp_start_transport_tasks()` installs `_zp_multicast_send_join_task_fn` for `_Z_TRANSPORT_RAWETH_TYPE`, and `_zp_multicast_send_join()` itself is declared/compiled for `Z_FEATURE_MULTICAST_TRANSPORT || Z_FEATURE_RAWETH_TRANSPORT`.

As written, a raweth-only build would compile but skip this interest-triggered join path, which looks like a behaviour change rather than just a feature-guard fix.

src/api/api.c
```diff
@@ -2119,7 +2119,7 @@ const z_loaned_keyexpr_t *z_queryable_keyexpr(const z_loaned_queryable_t *querya
     }
     _z_session_t *zn = _Z_RC_IN_VAL(&sess_rc);
 #else
-    _z_session_t *zn = _z_session_weak_as_unsafe_ptr(&sub->_zn);
+    _z_session_t *zn = _z_session_weak_as_unsafe_ptr((_z_session_weak_t *)&queryable->_zn);
```
gmartin82, line 2122: The `queryable->_zn` fix looks correct, but can we avoid casting away const here?

If `_z_session_weak_as_unsafe_ptr()` only reads the weak reference, it should probably accept `const _z_session_weak_t *` or have a const variant. The cast fixes the warning locally but weakens the const contract in this public API path.

src/api/api.c
```diff
@@ -2415,7 +2415,7 @@ const z_loaned_keyexpr_t *z_subscriber_keyexpr(const z_loaned_subscriber_t *sub)
     }
     _z_session_t *zn = _Z_RC_IN_VAL(&sess_rc);
 #else
-    _z_session_t *zn = _z_session_weak_as_unsafe_ptr(&sub->_zn);
+    _z_session_t *zn = _z_session_weak_as_unsafe_ptr((_z_session_weak_t *)&sub->_zn);
```
gmartin82, line 2418: As above.

src/net/filtering.c
```diff
@@ -309,7 +309,7 @@ z_result_t _z_write_filter_ctx_clear(_z_write_filter_ctx_t *ctx) {
 #endif
     return res;
 }
-
+#if Z_FEATURE_MATCHING == 1
```
gmartin82: This looks inconsistent with the current write-filter lifecycle.

`_z_write_filter_clear()` is now guarded by `Z_FEATURE_MATCHING`, but write filters are created and used in other paths too: publishers/queriers create them, and the put/delete/get paths consult `_z_write_filter_active()`.

So a build with matching disabled can still create/use the filter, but then skip cleanup on undeclare. Could we either keep cleanup available for those paths, or consistently compile out the create/use paths as well?

gmartin82 (changes requested):

src/api/api.c
DenisBiryukov91: I would need to make the same const cast inside _z_session_weak_as_unsafe_ptr which would simply hide the fact that I'm getting a non-const pointer from a const one.
Normally this part should be completely removed (in a separate PR?) since we should not allow bypassing alive session check. The fix here is purely to silence compiler warnings/errors.

src/api/api.c
DenisBiryukov91: Same as above

src/session/interest.c
DenisBiryukov91: this explicitly calls send join on a `_multicast` transport and not the `_raweth`, so indeed this line should be skipped for raweth transport (otherwise it will be UB). Please let me know if raweth indeed requires a send_join to be called from `_raweth`transport in this case

src/net/filtering.c
DenisBiryukov91: fixed

src/link/transport/udp/udp_multicast_lwip_common.h
gmartin82: Fixed by guarding include.

src/session/interest.c
gmartin82: Yeah sorry my mistake. Yes, agreed — I conflated two different paths.

I was looking at raweth using the same join function via `_zp_multicast_send_join_task_fn` in `_zp_start_transport_tasks()`, but this code is in `_z_interest_process_interest()` and explicitly calls `_zp_multicast_send_join()` through `zn->_tp._transport._multicast`.

So I agree this specific call should not run for a raweth transport.

gmartin82 (approved):
Looks good.

src/link/transport/udp/udp_multicast_lwip_common.h
DenisBiryukov91: fixed

src/session/interest.c
gmartin82: Looks good.

PR #1236, opened 2026-06-11
fix(codec): make locator decode cleanup safe

evshary:
## Description

This PR makes locator decoding safer when a `HELLO` message contains invalid locator data. It ensures partially decoded locator arrays are left in a safe state and are cleaned up correctly on failure.

### What does this PR do?

- Initializes locator array entries before decoding them so cleanup can safely handle partially decoded arrays.
- Clears partially decoded locator arrays when `HELLO` locator decoding fails.
- Removes the old partial-length failure handling in the locator decode loop.

Before this change, a malformed locator payload could leave a partially built locator array behind. That made failure handling unsafe and could lead to leaks or cleanup problems. This PR makes the failure path clean and safe by preparing every locator slot before decode and explicitly clearing the array when decoding fails.

### Related Issues

---
## 🏷️ Label-Based Checklist

**Labels:** `bug`

## 🐛 Bug Fix Requirements

**Instructions:**


evshary:
The PR should not be merged while @DenisBiryukov91's refactor later will fix the issue.

PR #1238, opened 2026-06-11
fix(codec): reject malformed metadata without separators

evshary:
## Description

This PR makes locator metadata parsing fail cleanly when the input does not contain a valid `key=value` separator. It prevents malformed `HELLO` locator metadata from getting stuck in the parser.

### What does this PR do?

- Detects metadata entries that do not contain a `=` separator.
- Returns a normal invalid-config error for malformed metadata instead of looping forever.

A malformed locator metadata string could previously enter `_z_str_intmap_from_strn()` and never make progress because the parser did not handle the "no separator found" case. This change turns that path into a normal parse failure so invalid inputs are rejected cleanly instead of causing fuzzing timeouts.

### Related Issues

---
## 🏷️ Label-Based Checklist

**Labels:** `bug`

## 🐛 Bug Fix Requirements

**Instructions:**


evshary:
This is included in https://github.com/eclipse-zenoh/zenoh-pico/pull/1248

PR #1239, opened 2026-06-11
fix(codec): bound locator metadata to the metadata section

evshary:
## Description

This PR fixes locator metadata parsing so the metadata end is searched only within the metadata portion of the locator string. It prevents malformed locator strings from producing an invalid length and crashing the decoder.

### What does this PR do?

- Limits the `#` terminator search to the metadata substring after `?`.
- Uses the end of the metadata substring when no `#` terminator is present.
- Prevents wrapped metadata lengths from being passed into the key-value parser.

Before this change, locator metadata parsing could find a `#` that appeared earlier in the locator string, before the metadata section. That made the computed metadata length underflow and turn into a huge unsigned size, which could later crash parsing. This PR ensures the metadata bounds are computed from the metadata section itself, so malformed inputs fail safely instead of causing invalid memory access.

### Related Issues

---
## 🏷️ Label-Based Checklist

**Labels:** `bug`

## 🐛 Bug Fix Requirements

**Instructions:**

### DIFF
--- a/src/link/endpoint.c
+++ b/src/link/endpoint.c
@@ -154,9 +154,10 @@ z_result_t _z_locator_metadata_from_string(_z_str_intmap_t *strint, const _z_str
         return _Z_RES_OK;
     }
 
-    const char *p_end = (char *)memchr(_z_string_data(str), ENDPOINT_CONFIG_SEPARATOR, _z_string_len(str));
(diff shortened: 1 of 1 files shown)


evshary:
The PR should not be merged while @DenisBiryukov91's refactor later will fix the issue.

PR #1242, opened 2026-06-11
fix: clear LRU cache values on eviction

gmartin82:
## Description

Fixes a memory leak in the generic LRU cache eviction path used by the RX cache.

### What does this PR do?

When the LRU cache reaches capacity, it evicts the least-recently-used node before inserting the new entry. The eviction path previously removed and freed the cache node, but did not clear the value stored inside it.

This PR updates the LRU insertion/eviction path so evicted values are cleared with the cache element clear callback before the node is freed. This matches the existing full-cache clear/delete behavior.

A regression test was added to verify that evicting an entry invokes the value clear callback.

With `Z_FEATURE_RX_CACHE` enabled, subscription/queryable cache entries own copied key expressions and cloned ref-counted vectors. If traffic uses more distinct key expressions than `Z_RX_CACHE_SIZE`, the RX cache churns and evicts old entries.

Before this fix, those evicted entries leaked their owned contents. In a scenario with 23 topics total, 12 received topics, and the default cache size of 10, memory usage grew significantly during RX cache churn.

After this fix, RX cache memory remains bounded by the configured cache capacity instead of growing with the number of evictions.

### Related Issues

N/A

---
## 🏷️ Label-Based Checklist

**Labels:** `bug`

## 🐛 Bug Fix Requirements

**Instructions:**

### DIFF
--- a/include/zenoh-pico/collections/lru_cache.h
+++ b/include/zenoh-pico/collections/lru_cache.h
@@ -45,7 +45,8 @@ typedef struct _z_lru_cache_t {
 
 _z_lru_cache_t _z_lru_cache_init(size_t capacity);
 void *_z_lru_cache_get(_z_lru_cache_t *cache, void *value, _z_lru_val_cmp_f compare);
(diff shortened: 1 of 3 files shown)


steils (approved):
LGTM, thanks.

PR #1255, opened 2026-06-26
Optimize CI

OlivierHecart:
## Description
Optimize CI to avoid unneeded runner consumption.

### What does this PR do?
Don't run CI on push.
Download Zenoh instead of building it.
Several minor optimizations.

Decrease CI cost.

---
## 🏷️ Label-Based Checklist

**Labels:** `internal`

## 🏠 Internal Change

**Instructions:**


.github/workflows/ci.yml
```diff
@@ -15,13 +15,16 @@ name: CI
...
-    branches: ["**"]
+    branches: ["main"]
   pull_request:
     branches: ["**"]
   schedule:
     - cron: "0 6 * * 1-5"
   workflow_dispatch:
 
+env:
+  ZENOH_URL: https://github.com/eclipse-zenoh/zenoh/releases/download/1.9.0/zenoh-1.9.0-x86_64-unknown-linux-gnu-standalone.zip
```
evshary: Instead of using the fixed version, we can probably use gh to get the latest artifacts, or probably reuse others' workflow. Then we don't need to update the version at every release.
https://github.com/robinraju/release-downloader#download-asset-from-the-latest-release-in-the-current-repository

.github/workflows/integration.yaml
```diff
@@ -16,6 +16,10 @@ name: integration
 on:
   workflow_call:
 
+env:
+  ZENOH_LINUX_URL: https://github.com/eclipse-zenoh/zenoh/releases/download/1.9.0/zenoh-1.9.0-x86_64-unknown-linux-gnu-standalone.zip
```
evshary: Same here.

.github/workflows/ci.yml
```diff
@@ -15,13 +15,16 @@ name: CI
 
 on:
   push:
-    branches: ["**"]
+    branches: ["main"]
   pull_request:
     branches: ["**"]
   schedule:
     - cron: "0 6 * * 1-5"
```
diogomatsubara: We can probably remove the scheduled CI run. Since the CI workflow will be run on main when changes are merged that should suffice to keep it green and development on pico is active enough nowadays that we'll get it fixed during development.

.github/workflows/ci.yml
```diff
@@ -15,13 +15,16 @@ name: CI
 
 on:
   push:
-    branches: ["**"]
+    branches: ["main"]
   pull_request:
     branches: ["**"]
```
diogomatsubara, line 20: I'd suggest add `[ready_for_review]` here instead of `[**]` as any push to an open PR would trigger the `synchronize` event

.github/workflows/ci.yml
```diff
@@ -15,13 +15,16 @@ name: CI
 
 on:
   push:
-    branches: ["**"]
+    branches: ["main"]
```
diogomatsubara: Could you add:
```
    branches:
      - "main"
      - "release/*"
      - "!release/dry-run/*"
``` instead?

File: .github/workflows/ci.yml [hunk 6]
```diff
@@ -33,10 +36,12 @@ jobs:
           mbedtls-version: mbedtls-3.6.5
       - name: Checkout code
         uses: actions/checkout@v4
+        with:
+          fetch-depth: 1
 
       - name: Build & run tests
         run: |
-          sudo apt install -y ninja-build
+          sudo apt update && sudo apt install -y ninja-build
```
diogomatsubara, line 46: Maybe it's worth caching the apt installs in a github cache as well as described in here: https://skorski.info/2023/03/25/effective-caching-with-github-actions/

Or even have one initial job that installs all the tooling and cache them and in the individual jobs/steps you would re-use the cache to avoid a apt-update and apt installing the tool directly from the cache.

diogomatsubara (changes requested):
@OlivierHecart thanks for working on this, it's much appreciated. The comments about the events that trigger the CI workflow I'd like addressed but the comments on the apt-update/apt-install are more general non-blocking comments.

.github/workflows/ci.yml
OlivierHecart: Done

.github/workflows/integration.yaml
OlivierHecart: Done

.github/workflows/ci.yml
OlivierHecart: Done

.github/workflows/ci.yml
OlivierHecart: Done

.github/workflows/ci.yml
OlivierHecart: I'm not sure about this one. Devs still want to have regular CI runs during their development to track any regressions. Since we disabled CI on push, creating a (draft) PR is a way for devs to run CI.

.github/workflows/ci.yml
OlivierHecart: I'm not sure if GitHub caching would be relevant for self hosted runners. And that's what we want to optimize for. Maybe this needs a bit more investigations in another PR ?

.github/workflows/ci.yml
diogomatsubara: Fair enough. If we keep the pull_request triggers for any event, then we could add the concurrency group like this: 

```
concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: true
``` 

to the workflow so on synchronize, the previous jobs gets automatically cancelled. What do you think?

diogomatsubara (changes requested):

.github/workflows/ci.yml
OlivierHecart: Done

gmartin82: I actually think Diogo was right on this one. Some devs commit very frequently at times and CI is triggered when it obviously isn't ready for CI to be run.

CI is expensive, it's reasonable to offer a way to trigger a CI run and expect devs to exercise judgement and choose when to run it rather than indiscriminately triggering CI on every push to an open PR.

Regardless I think adding the cancel in progress below was a good choice.

diogomatsubara (approved):
LGTM

evshary (approved):

.github/workflows/ci.yml
```diff
@@ -15,13 +15,18 @@ name: CI
 
 on:
   push:
-    branches: ["**"]
+    branches: ["main", "release/*", "!release/dry-run/*"]
```
DenisBiryukov91, line 18: I would rather prefer to able to run ci without the need to create PR to have immediate feedback regarding my work in progress without the need to flood PR tab with draft PRs

OlivierHecart: You can run the CI on any branch by hand with workflow_dispatch.

PR #1256, opened 2026-06-27
collections: fix variant_template move leak, macro cleanup

steils:
<variant>_move template did not destroy the previous destination value overwriting, contradicting its documented contract.

- Document that *dst is overwritten and its previous contents are not destroyed.
- Fix variant_template move tests to use a destination that is safe to overwrite.
- Remove the unused _ZP_VARIANT_FN_VISIT / _ZP_VARIANT_FN_VISIT_CTX defines and two #undefs.
- Wrap the _ZP_FIND* helper macros in do { ... } while (0).

---
## 🏷️ Label-Based Checklist

**Labels:** `internal`

## 🏠 Internal Change

**Instructions:**

### DIFF
--- a/include/zenoh-pico/collections/variant_template.h
+++ b/include/zenoh-pico/collections/variant_template.h
@@ -107,7 +107,7 @@
 //   void    NAME_destroy(NAME_t *v)
 //
 // Move (variant-to-variant)  — omitted when _ZP_VARIANT_TEMPLATE_NO_MOVE_FN is defined:
-//   void    NAME_move(NAME_t *dst, NAME_t *src)
+//   void    NAME_move(NAME_t *dst, NAME_t *src)  — dst must be uninitialized or safe to overwrite
 //
 // Inspection:
 //   bool       NAME_is_none(const NAME_t *v)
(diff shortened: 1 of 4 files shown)


include/zenoh-pico/collections/variant_template.h
```diff
@@ -672,6 +670,10 @@ static inline void _ZP_CAT(_ZP_VARIANT_TEMPLATE_NAME, destroy)(_ZP_VARIANT_TEMPL
 #ifndef _ZP_VARIANT_TEMPLATE_NO_MOVE_FN
 static inline void _ZP_CAT(_ZP_VARIANT_TEMPLATE_NAME, move)(_ZP_VARIANT_TEMPLATE_TYPE *dst,
                                                             _ZP_VARIANT_TEMPLATE_TYPE *src) {
+    if (dst == src) {
```
DenisBiryukov91: Move is considered to be a constructor, rather than an assignment across zenoh-pico code base (same as copy btw), which commonly assumes that dst is not initialized and is safe to overwrite. So it would be more preferable to just update the docs to this function, rather than change the code, to keep behavior in sync with other `move` constructors.

tests/z_variant_template_test.c
```diff
@@ -772,6 +810,8 @@ int main(void) {
     test_default_from_a();
     test_default_from_b();
     test_default_move();
+    test_default_move_destroys_old_dst();
```
DenisBiryukov91: As I mentioned above, it is not advisable to change the variant's move behavior, and thus these tests are not needed.

DenisBiryukov91 (approved):

DenisBiryukov91 (changes requested):
Could please you also update the void NAME_move(NAME_t *dst, NAME_t *src) description in the readme here https://github.com/eclipse-zenoh/zenoh-pico/blob/main/include/zenoh-pico/collections/README.md#api-name--configured-name-xn--alternative-name . Otherwise LGTM

DenisBiryukov91 (approved):

PR #1258, opened 2026-06-29
link: abstract peer handling and open/listen dispatch

gmartin82:
## Description

This PR refactors internal link and unicast transport handling so that common transport code no longer depends on concrete socket-shaped peer state, and generic link open/listen dispatch no longer hard-codes the built-in transport selection chain.

There are no public API, protocol, locator, or configuration additions in this PR.

### What does this PR do?

* Introduces an internal `_z_link_peer_t` handle for per-peer link state and lifecycle.
* Makes concrete link-peer state opaque to common unicast transport code.
* Routes peer read, write, endpoint lookup, blocking mode, readiness waiting, outbound peer creation, accept handling, close, and cleanup through link/link-peer operations.
* Keeps link-owned transmission separate from transmission to an explicitly selected peer.
* Removes generic unicast transport dependencies on raw sockets, TLS internals, TCP accept, and socket wait APIs.
* Splits peer accept from accept completion so capacity checks can happen before expensive link-specific completion work such as TLS handshaking.
* Adds link driver descriptors with validate, create, open, and listen callbacks.
* Adds a compiled, ordered driver registry for built-in link drivers.
* Refactors generic link open/listen dispatch to iterate the registry instead of hard-coding transport-specific selection chains in `link.c`.
* Keeps built-in driver selection feature-gated and deterministic.
* Adds focused tests for TX destination behaviour and the compiled driver registry shape.

The previous structure made common transport and generic link dispatch depend too directly on concrete built-in link implementations.

Common unicast transport code previously depended on details of the built-in peer representation. This made the transport layer more tightly coupled to the current link implementations than necessary.

Generic link dispatch also knew which built-in transports existed and selected them through hard-coded transport-specific branches. That made link construction and selection harder to extend internally without editing the central dispatch path.

After this change:

* common unicast transport code does not need to know how a concrete link represents a peer;
* common unicast transport code does not need to know whether a peer is socket-backed;
* peer open, accept, read, write, readiness, endpoint lookup, close, and cleanup are owned by the relevant link implementation;
* generic link open/listen dispatch no longer needs transport-specific selection branches in `link.c`;
* concrete link implementations remain responsible for endpoint validation, construction, activation, embedded link/peer state, and link/peer operations.

This keeps the common session and transport layers responsible for Zenoh behaviour, while link implementations own link-specific state and I/O details.

### Related Issues

ZEN-700

---
## 🏷️ Label-Based Checklist

**Labels:** `internal`

## 🏠 Internal Change

**Instructions:**


src/transport/common/tx.c
```diff
@@ -61,12 +61,89 @@ static _z_zint_t _z_transport_tx_get_sn(_z_transport_common_t *ztc, z_reliabilit
...
+            _Z_ERROR_LOG(_Z_ERR_TRANSPORT_TX_FAILED);
+            ret = _Z_ERR_TRANSPORT_TX_FAILED;
+            break;
+    }
+
+    if (sent) {
+        ztc->_transmitted = true;  // Tell session we transmitted data
+    }
+    return ret;
+}
```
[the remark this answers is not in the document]
gmartin82: Fixed

src/link/common/socket_ops.c
```diff
@@ -0,0 +1,135 @@
...
+z_result_t _z_link_socket_peer_get_endpoints(const _z_link_peer_t *peer, char *local, size_t local_len, char *remote,
+                                             size_t remote_len) {
+    const _z_sys_net_socket_t *socket = _z_link_socket_peer_get_socket_const(peer);
+    if (socket == NULL) {
+        _Z_ERROR_RETURN(_Z_ERR_INVALID);
+    }
+    return _z_socket_get_endpoints(socket, local, local_len, remote, remote_len);
+}
+
+void _z_link_socket_peer_close(_z_link_peer_t *peer) {
```
DenisBiryukov91: Is there a reason why functions operating on _z_link_peer_t have  _z_link_socket_peer prefix instead of just  _z_link_peer one ?

src/link/common/socket_ops.c
```diff
@@ -0,0 +1,135 @@
...
+    _z_link_socket_peer_state_t *state = _z_link_peer_socket_state(peer);
+    if ((state == NULL) || state->_closed || (state->_close_f == NULL)) {
```
DenisBiryukov91: Is it intended that if state->_close_f == NULL the _closed field is never set ? If the flag is used purely to track calls to _close_f, would not it make more sense to reset _close_f to NULL instead ?

include/zenoh-pico/link/link.h
```diff
@@ -82,64 +73,112 @@ typedef struct _z_link_capabilities_t {
...
+    z_result_t (*_get_endpoints_f)(const struct _z_link_peer_t *peer, char *local, size_t local_len, char *remote,
+                                   size_t remote_len);
+    void (*_close_f)(struct _z_link_peer_t *peer);
+} _z_link_peer_ops_t;
+
+typedef void (*_z_link_peer_drop_f)(void *state);
+
+typedef struct _z_link_peer_impl_t {
+    const _z_link_peer_ops_t *_ops;
+    void *_state;
```
DenisBiryukov91: do you really need another heap-allocated abstraction for state, given that it resides within another heap-allocated abstraction _z_link_peer_impl_t which is in turn leaves inside another heap-allocated abstraction _z_link_t ?  Unless I miss something, it looks like a single heap-allocated (or probably even statically-allocated union-based one as it was previously) abstraction would do the job.

src/transport/common/tx.c
```diff
@@ -61,12 +61,90 @@ static _z_zint_t _z_transport_tx_get_sn(_z_transport_common_t *ztc, z_reliabilit
     return sn;
 }
 
+typedef enum {
```
DenisBiryukov91: We have a typesafe z_variant_template which would automatically generate most of the boiler plate code here.

src/link/common/socket_ops.c
gmartin82: Yes. `socket_ops.c` contains common helpers for socket-backed link-peer implementations. Hence the choice of prefix.

src/link/common/socket_ops.c
gmartin82: Fixed, I've removed the `_closed` flag and instead reset `_close_f` to `NULL`.

src/transport/common/tx.c
```diff
@@ -61,12 +61,90 @@ static _z_zint_t _z_transport_tx_get_sn(_z_transport_common_t *ztc, z_reliabilit
...
+                curr_list = _z_transport_peer_unicast_slist_next(curr_list);
+            }
+            break;
+        }
+        case _Z_TRANSPORT_TX_DEST_LINK_PEER:
+            ret = _z_link_peer_send_wbuf(ztc->_link, &ztc->_wbuf, dest->_peer);
+            sent = (ret == _Z_RES_OK);
+            break;
+        default:
+            _Z_ERROR_LOG(_Z_ERR_TRANSPORT_TX_FAILED);
```
DenisBiryukov91: should this be _Z_ERR_INVALID (or something similar), rather than _Z_ERR_TRANSPORT_TX_FAILED ?

include/zenoh-pico/link/link.h
gmartin82: The indirection here is intentional. `_z_link_peer_impl_t` is the generic ref-counted peer handle; `_state` is opaque driver-owned peer state.

I don't think returning to a union is the right direction. Part of the reason for this shape is that generic link/transport code should not need to know the concrete peer layout or lifetime rules. TLS is a good example: it is still socket-backed, but it is not identical to the other socket-backed links. It needs TLS-specific read/write behaviour and different close/lifetime handling, and different TLS backends may need different per-peer state. That belongs in the TLS link driver, not in the generic peer representation.

I agree allocation overhead matters, but these allocations are tied to link/peer instantiation, not to the packet/message path. That means they can be predicted from the configured number of links and peers.

I have also switched the peer implementation to use the simple refcount. It is not a huge saving, but link peers only need strong shared ownership here, not weak references, so the simple refcount is all that is strictly needed.

src/transport/common/tx.c
gmartin82: Fixed. I've changed it to `Z_ERR_INVALID` as the code should never really reach here.

src/transport/common/tx.c
gmartin82: I don't think `z_variant_template` is the right fit here.

This is a small file-local dispatch descriptor, not a general-purpose variant with ownership semantics. It only distinguishes the default link destination, a peer-list destination, and an explicitly selected link peer.

The previous code overloaded `peers == NULL` for the default destination and used a temporary one-element peer list for selected-peer sends. That was not a good fit once peers carry `_z_link_peer_t` ownership/lifetime semantics.

Keeping this as an explicit enum/struct avoids faking a peer list or copying peer state, and makes the batching rule clear: selected-peer sends first flush any pending shared batch to the normal destination, then send directly to the selected peer.

Using `z_variant_template` would require wrapper types for borrowed pointers and either `none` or a dummy type for the default destination. That feels heavier than this local dispatch helper.

DenisBiryukov91: I'm not sure why would you need  wrapper types for borrowed pointers. If you do not want move/destroy semantics in the variant - you just leave the corresponding functions unspecified. You also get `none` variant for free as the one corresponding to non-initialized/empty state.

gmartin82: Fair point, I misunderstood how `z_variant_template` would apply here.

I have changed the local TX destination dispatch helper to use `z_variant_template`. The variant now represents an optional destination override: `none` is the normal/default TX destination, and the alternatives are borrowed peer-list and link-peer pointers.

include/zenoh-pico/link/link.h
DenisBiryukov91: Still, it is not entirely clear why state should necessarily be allocated separately on heap and stored as void*. If _z_link_peer_t owns state, its subclasses can surely own it by value:

```c
typedef struct _z_link_peer_t {
    _z_link_peer_impl_t ops;
} _z_link_peer_t;

void _z_link_peer_clear(_z_link_peer_t *l) {
    l->ops.drop_f(l);
}

z_result_t _z_link_peer_do_something(_z_link_peer_t *l, args_t args) {
    return l->ops.do_something_f(l, args);
}

// _z_link_peer_subclass_t 'inherits' _z_link_peer_t
typedef struct _z_link_peer_subclass_t {
    _z_link_peer_t _base; // the address of the first field always corresponds to the address of the whole struct
    _z_link_peer_subclass_state_t _s;
} _z_link_peer_subclass_t;

void _z_link_peer_subclass_clear(_z_link_peer_subclass_t* s) {
   ...
    _z_link_peer_subclass_state_clear(&s->state);
}

void _z_link_peer_subclass_clear_from_base(_z_link_peer_t* base) {
    _z_link_peer_subclass_clear((_z_link_peer_subclass_t*)base);
}

z_result_t _z_link_peer_subclass_do_something(_z_link_peer_subclass *s, args_t args) {
    return do_something(&s->_s, args);
}

z_result_t _z_link_peer_subclass_do_something_from_base(_z_link_peer_t *base, args_t args) {
    return _z_link_peer_subclass_do_something((_z_link_peer_subclass_t*)base, args);
}

void _z_link_peer_t* _z_link_peer_subclass_create() {
    _z_link_peer_subclass_t* s = (_z_link_peer_subclass_t*)malloc(sizeof(_z_link_peer_subclass_t)); 
    s->base.ops = {.do_something_f = _z_link_peer_subclass_do_something_from_base, .drop_f = _z_link_peer_subclass_clear_from_base };
    s->state = ...;
    ...;
    return (_z_link_peer_t*)(&s->_base);
}

int main() {
    _z_link_peer_rc_t link_peer_rc = _z_link_peer_rc_new(_z_link_peer_subclass_create());
}
```

This might probably require adding some code generation templates for automatically creating `from_base` functions and constructor filling ops vtable, but looks way more typesafe (i.e. there is no chance to provide wrong state to vtable functions given that now they can not be constructed separately from each other) and removes the need for extra allocations.

src/transport/common/tx.c
```diff
@@ -61,12 +73,76 @@ static _z_zint_t _z_transport_tx_get_sn(_z_transport_common_t *ztc, z_reliabilit
...
 }
 
+/*
+ * The shared TX buffer does not retain a destination. Buffered data belongs to the
+ * transport's normal destination; a selected-peer send first flushes any pending
+ * shared batch there, then sends its message directly without entering the shared batch.
+ */
+static _z_transport_tx_dest_t _z_transport_tx_dest_default(void) { return _z_transport_tx_dest_none(); }
+
+static _z_transport_tx_dest_t _z_transport_tx_dest_peer_list(_z_transport_peer_unicast_slist_t *peers) {
```
DenisBiryukov91: Any reason why generated function names cannot be used as is without wrapping them ?

src/transport/common/tx.c
```diff
@@ -77,38 +153,35 @@ static z_result_t _z_transport_tx_send_fragment_inner(_z_transport_common_t *ztc
...
-        } else {
-            _z_transport_peer_unicast_slist_t *curr_list = peers;
-            while (curr_list != NULL) {
-                _z_transport_peer_unicast_t *curr_peer = _z_transport_peer_unicast_slist_value(curr_list);
-                // Send on peer socket
-                _z_link_send_wbuf(ztc->_link, &ztc->_wbuf, &curr_peer->_socket);
-                curr_list = _z_transport_peer_unicast_slist_next(curr_list);
+        z_result_t send_ret = _z_transport_tx_send_wbuf(ztc, dest);
+        if (send_ret != _Z_RES_OK) {
+            if (_z_transport_tx_dest_is_default(dest)) {
```
steils: For a LINK_PEER destination, this keeps fragmenting and sending subsequent fragments after a send failure.
DEFAULT and LINK_PEER should both abort on `send_ret != _Z_RES_OK`.

It should be something like:

```
if (send_ret != _Z_RES_OK) {
    if (!_z_transport_tx_dest_is_peer_list(dest)) {
        return send_ret;
    }
    if (ret == _Z_RES_OK) {
        ret = send_ret;
    }
}
```

src/api/admin_space.c
```diff
@@ -310,31 +310,7 @@ static z_result_t _ze_admin_space_encode_link(_z_json_encoder_t *je, const _z_li
...
-        case _Z_LINK_TYPE_TLS:
-            _Z_RETURN_IF_ERR(_z_json_encoder_write_string(je, "tls"));
-            break;
-        case _Z_LINK_TYPE_RAWETH:
-            _Z_RETURN_IF_ERR(_z_json_encoder_write_string(je, "raweth"));
-            break;
-        default:
-            return _Z_ERR_INVALID;
-    }
+    _Z_RETURN_IF_ERR(_z_json_encoder_write_z_string(je, &link->_endpoint._locator._protocol));
```
steils, line 313: This makes admin-space `link.type` report the locator protocol instead of the link kind. For raweth, the locator protocol is "reth", while the link/transport kind exposed elsewhere is "raweth". I think we should preserve the old raweth spelling here, e.g write "raweth" when `link->_cap._transport == Z_LINK_CAP_TRANSPORT_RAWETH`, unless the admin-space payload compatibility change is intentional.

include/zenoh-pico/link/link.h
```diff
@@ -82,64 +73,112 @@ typedef struct _z_link_capabilities_t {
...
+
+static inline const struct _z_link_peer_t *_z_link_peer_iter_get_peer(const _z_link_peer_iter_t *iter) {
+    return iter->_get_peer_f(iter);
+}
+
+static inline void _z_link_peer_iter_set_ready(_z_link_peer_iter_t *iter, bool ready) {
+    iter->_set_ready_f(iter, ready);
+}
+
+typedef struct _z_link_peer_ops_t {
```
steils, line 152: Please add a comment above `_z_link_peer_ops_t` documenting the current contract for `_read_f` and `_write_f`, including whether `SIZE_MAX` means "fatal error" or "would block", and what would mean `0`.

src/transport/unicast/accept.c
```diff
@@ -65,83 +60,43 @@ static void _zp_unicast_dispatch_connected_event(_z_transport_unicast_t *ztu, co
...
-    if (socket_ptr == NULL) {
-        _Z_ERROR_LOG(_Z_ERR_INVALID);
-        return _z_fut_fn_result_ready();
-    }
 
-    _z_sys_net_socket_t listen_socket = *socket_ptr;
-    _z_sys_net_socket_t con_socket = {0};
-    z_result_t ret = _z_tcp_accept(&listen_socket, &con_socket);
+    _z_link_peer_t link_peer = _z_link_peer_null();
+    z_result_t ret = _z_link_accept_peer(ztu->_common._link, &link_peer);
```
steils, line 65: For TLS listeners, `_z_link_accept_peer()` performs the TLS accept/handshake before this code checks `Z_LISTEN_MAX_CONNECTION_NB`. When the listener is already at the connection limit, every refused TLS connection still costs a full TLS handshake.
The limit decision should happen after accepting the socket but before the expensive TLS handshake.

steils (changes requested):

include/zenoh-pico/link/link.h
DenisBiryukov91: maybe it is better to avoid using magic numbers at all and return z_result_t directly ?

include/zenoh-pico/link/link.h
```diff
@@ -82,64 +73,112 @@ typedef struct _z_link_capabilities_t {
...
-        _z_ws_socket_t _ws;
-#endif
-#if Z_FEATURE_LINK_TLS == 1
-        _z_tls_socket_t _tls;
-#endif
-#if Z_FEATURE_RAWETH_TRANSPORT == 1
-        _z_raweth_socket_t _raweth;
-#endif
-    } _socket;
+    void *_state;
```
DenisBiryukov91: Same concerns as for _z_link_peer_t regarding using void* heap-allocated pointers for something that can be easily achieved without it.

gmartin82: I've explained my reasons. This is no longer constructive review feedback.

include/zenoh-pico/link/link.h
```diff
@@ -82,64 +73,112 @@ typedef struct _z_link_capabilities_t {
...
+static inline const struct _z_link_peer_t *_z_link_peer_iter_get_peer(const _z_link_peer_iter_t *iter) {
+    return iter->_get_peer_f(iter);
+}
+
+static inline void _z_link_peer_iter_set_ready(_z_link_peer_iter_t *iter, bool ready) {
+    iter->_set_ready_f(iter, ready);
+}
+
+typedef struct _z_link_peer_ops_t {
+    size_t (*_read_f)(const struct _z_link_t *link, const struct _z_link_peer_t *peer, uint8_t *ptr, size_t len);
```
DenisBiryukov91, line 153: Is there a reason why both _z_link_t and _z_link_peer_t are required here, given that _z_link_t always contains its _z_link_peer_t

include/zenoh-pico/link/link.h
DenisBiryukov91: Please read the suggestion regarding subclassing that I left and reconsider again.

gmartin82: No.

include/zenoh-pico/link/link.h
gmartin82: I've already explained my reasoning. I will not enter into further discussions over this.

src/transport/common/tx.c
gmartin82: The generated functions can be used directly, but I would prefer to keep these local wrappers.

The wrapper names make the intent clearer than the generated variant names. In particular, the variant `none` state represents the default TX destination here, not an absence of a destination.

I have marked the helpers `inline` though.

src/transport/common/tx.c
gmartin82: Fixed.

`DEFAULT` and `LINK_PEER` destinations now abort immediately on `send_ret != _Z_RES_OK`.

For `PEER_LIST`, this preserves the current fan-out behaviour: continue attempting delivery to the remaining peers while still returning the first failure.

Strictly speaking, fragmented peer-list sends should probably also stop sending later fragments to any peer that has already failed, since that peer can no longer reconstruct the message. I’ve left that out because it is not really part of this refactoring, but it is worth considering as a follow-up improvement.

src/api/admin_space.c
gmartin82: This was intentional. The admin-space code now reports the locator protocol directly instead of maintaining a separate link-type mapping, and I considered the `"raweth"` → `"reth"` change an acceptable consequence of that simplification.

However, I’m happy to restore the previous `"raweth"` spelling if preserving the existing admin-space payload is preferred.

include/zenoh-pico/link/link.h
gmartin82: Added a comment documenting the '_z_link_peer_ops_t' callback contract.

The 'size_t' byte-count return convention, including use of 'SIZE_MAX' as a sentinel, is pre-existing in the link layer rather than introduced by this refactor. For example, '_z_link_send_wbuf()' already treats 'wb == SIZE_MAX' as a transport TX failure:

https://github.com/eclipse-zenoh/zenoh-pico/blob/0ffe44a0502fd2cebeb974147239d4bca85b2ca7/src/link/link.c#L214

Changing this to a 'z_result_t' plus out-parameter API may be cleaner, but that would change the existing link I/O API shape and is outside the scope of this refactor.

src/transport/unicast/accept.c
gmartin82: Thanks, good catch — this was a regression from the refactor.

I’ve added a split between accepting a peer and completing the accept. `_z_link_accept_peer()` now creates a provisional accepted peer using the minimum work needed to own the incoming connection. The transport can then apply `Z_LISTEN_MAX_CONNECTION_NB` before calling `_z_link_accept_peer_complete()`.

For TLS, the completion step performs `_z_tls_accept()`, so we preserve the previous ordering:

```text
TCP accept → connection-limit check → TLS accept → Zenoh handshake
```

For links that do not need a completion step, the completion callback is left unset and `_z_link_accept_peer_complete()` succeeds as a no-op.

Do you think this is adequate for possible future transports as well? My intent is that transports with an expensive protocol-specific accept step can defer that work to `_accept_peer_complete_f`, while simpler transports can just leave it unset.

include/zenoh-pico/link/link.h
gmartin82: Yes. The `_z_link_peer_t` owned by `_z_link_t` is the default/link-owned peer; it is not necessarily the peer being operated on.

For example, on a listening unicast link, `link->_peer` represents the listener/default peer. Accepted peers are owned by the unicast transport, and each accepted transport peer has its own `_z_link_peer_t`. So the peer to read from or write to cannot be inferred from `link->_peer`.

The two arguments serve different purposes:

```text
link = shared link-level context/state/configuration/capabilities
peer = the concrete peer instance being read from or written to
```

The `link` argument is also needed by current code. For example, UDP unicast peer write uses the peer to get the socket, but uses the link to access the configured UDP endpoint/state needed by `_z_udp_unicast_write()`.

I’ve extended the `_z_link_peer_ops_t` comment to make this distinction explicit.

DenisBiryukov91: Very confusing. Following your explanation, it seems the cleaner design would be _z_link_peer_t containing a (shared reference to ?) _z_link_t, rather than _z_link_t owning a so-called 'default' _z_link_peer_t. It is not even clear what default means in this context. Should not be all _z_link_peers equal ? How for example you envision a situation where you would have both tcp and udp listeners (or just a tcp listener, with udp connect links) - what would be the 'default' link_peer here ?

gmartin82: This is not part of this PR’s scope.

The goal of this PR is not to redesign the link/transport ownership model. The goal is to remove common unicast transport assumptions about socket-shaped peer state, while preserving the existing behaviour and ownership shape as far as possible.

On `main`, `_z_link_t` already represents one concrete opened/listening link instance. It is not a collection of all peers or all possible links. This PR keeps that model. The difference is that peer/socket state is now accessed through an opaque `_z_link_peer_t`, rather than common transport code reaching into concrete socket/TLS fields.

Please keep review comments focused on issues within the stated scope of this change: preserving existing behaviour while hiding concrete peer representation from common unicast transport code. Larger ownership redesigns should be handled separately if we decide they are worth pursuing.

gmartin82:
The follow-on link driver registry work is now ready, so I have folded it into this PR for review alongside the peer abstraction changes.

The combined PR now covers the internal link/transport boundary as a single coherent refactoring: common unicast transport no longer depends on concrete peer representation details, and generic link open/listen dispatch no longer hard-codes the built-in transport selection chain.

include/zenoh-pico/link/link.h
gmartin82: Sorry, I understand what you mean now.

I have reworked this to use an embedded-base layout instead of storing separately allocated concrete state behind a `void *`.

Concrete link and peer implementations now own their state by value alongside the common base object. That removes the extra heap-state indirection introduced by the refactor, while keeping common link/transport code opaque to the concrete implementation layout.

The callback/vtable setup remains explicit and local to the concrete implementation. I have not introduced new code-generation templates here, as I think that would be a broader codebase pattern change rather than something to add as part of this refactor.

The remaining peer allocation is the refcount control block needed for `_z_link_peer_t` shared ownership.

include/zenoh-pico/link/link.h
gmartin82: Applied the same embedded-base pattern here as well.

Concrete link implementations now embed `_z_link_t` as their first member and own their driver-specific state by value alongside it, rather than storing a separately allocated state object behind `void *`.

This removes the extra heap-state indirection while keeping the common link/transport code operating on `_z_link_t *` without knowing the concrete implementation layout.

src/api/admin_space.c
steils: Okay, I don't have a strong opinion here. Maybe it's worth asking Sasha or Olivier.

src/link/multicast/udp.c
```diff
@@ -16,120 +16,169 @@
...
     const char *iface = _z_str_intmap_get(&endpoint->_config, UDP_CONFIG_IFACE_KEY);
     if (iface == NULL) {
-        _Z_ERROR_LOG(_Z_ERR_CONFIG_LOCATOR_INVALID);
         return _Z_ERR_CONFIG_LOCATOR_INVALID;
     }
 
     return _Z_RES_OK;
 }
 
 z_result_t _z_f_link_open_udp_multicast(_z_link_t *self) {
```
steils: `_z_f_link_open_udp_multicast` is not used anywhere anymore.

steils (approved):
`_z_f_link_open_udp_multicast` is not used anymore, and may be removed. Otherwise approved.

src/link/common/driver_registry.c
```diff
@@ -0,0 +1,80 @@
...
+static const _z_link_driver_t *const _z_link_drivers[] = {
+#if Z_FEATURE_LINK_TCP == 1
+    &_z_link_driver_tcp,
+#endif
+#if Z_FEATURE_LINK_TLS == 1
+    &_z_link_driver_tls,
+#endif
+#if Z_FEATURE_LINK_UDP_UNICAST == 1
+    &_z_link_driver_udp_unicast,
+#endif
```
DenisBiryukov91, line 59: I'm not sure it is such a good idea to mix multicast with unicast links and deserves a separate discussion. Now you hide the only way to understand whether a multicast transport or a unicast transport would be created (which before could be easily done by analyzing session mode + a single udp multicast locator validity check). Ideally we would want to first understand the transport to be used and then validate that all locators correspond to it, or alternatively being able to run both unicast and multicast in parallel, but this is also infeasible with the current design.

src/transport/common/tx.c
```diff
@@ -137,50 +211,40 @@ static inline bool _z_transport_tx_batch_has_data(_z_transport_common_t *ztc) {
...
 #else
-    return _z_transport_tx_flush_buffer(ztc, peers);
+    return _z_transport_tx_flush_buffer(ztc, dest);
 #endif
 }
 
 static z_result_t _z_transport_tx_batch_overflow(_z_transport_common_t *ztc, const _z_network_message_t *n_msg,
                                                  z_reliability_t reliability, _z_zint_t sn, size_t prev_wpos,
-                                                 _z_transport_peer_unicast_slist_t *peers) {
+                                                 const _z_transport_tx_dest_t *dest, bool force_flush) {
```
DenisBiryukov91, line 242: I'm not sure that fixing shortcoming of the current batching implementation belongs to this PR. It is clear that batching was broken for p2p, but it is probably better to discuss whether we want to address it by completely disabling it in p2p (and returning a error on attempt to enable it), using a separate batch buffer per peer, or the way you proposed (which essentially boils down to silently disabling batching for p2p without informing the user about it).

src/link/common/driver_registry.c
gmartin82: This is intentionally not different from the previous dispatch behaviour on main.

Before this PR, the common link code already selected between the built-in link implementations by validating the locator against each supported transport path. This PR moves that built-in dispatch policy out of the common code and into an ordered driver registry, but it does not try to redesign transport selection.

That was the stated aim of this work: keep the common link/transport code from knowing the concrete built-in link implementations, while preserving the existing behaviour as far as possible.

It was also stated that this work should be minimal. Changing the model so the transport type is selected first, validating all locators against a chosen transport type, or supporting unicast and multicast in parallel would all be larger design changes. Those may be worth discussing, but I do not think they belong in this PR.

src/transport/common/tx.c
gmartin82: I agree this should not become a broader batching redesign in this PR.

This change is not intended to fix batching for P2P generally. The existing behaviour was already not ideal.

The reason I touched it here is that this refactor makes the TX destination explicit. Once there is a distinct selected-peer destination, allowing that message to enter the shared batch would risk flushing it later to an unintended destination. This seemed like an easy, pragmatic fix.

A full design discussion about P2P batching, per-peer buffers, disabling batching for P2P, or making this configurable/visible, should happen separately.

I do not think we should reject a cheap incremental improvement here just because it is not the full batching redesign.

src/link/multicast/udp.c
gmartin82: Thanks for spotting this. I've removed it.

src/transport/common/tx.c
DenisBiryukov91: It just feels unnecessary, together with introduction of _z_transport_tx_dest_t enum, and creates conflicts with the proposal to pass destination peers via bit masks in #1265,  which looks to be more efficient in terms of time and space.

gmartin82: As I've stated the intention of this PR was to be a minimal refactor of what's there and this was clearly communicated at the time. I was not aware that another developer was re-writing parts of the code base which conflicts with this work because that information was not communicated to me.

DenisBiryukov91: Unfortunately it turned out to be the case. So the question is whether you plan to continue building something on your fixes for batching and _z_transport_tx_dest_t usage (i.e. most of the changes in transport/common/tx.c except _z_link_send_wbuf calls) or whether these changes are not essential and can be removed from this PR, probably in favor of ones introduced in #1265 ?

gmartin82: I'm not planning to make any further changes as the PR complies with my stated design intent.

steils: @gmartin82 I think the question was if these changes in `transport/commont/tx.c` Denis is talking about - are they needed in the future stages of ZEN-700, as you see it? Not in the current PR.

gmartin82: No, these specific `transport/common/tx.c` changes are not required for future work.

They were added as part of this PR because the existing TX path overloaded `peers == NULL` for the default destination and used peer-list handling even for directed peer sends. Making the destination explicit was intended to make the existing behaviour clearer and avoid copying/faking peer-list state now that peers carry link-owned lifetime semantics.

That also exposed an existing shared-batch limitation: the shared TX buffer has no destination identity. So the selected-peer path was kept out of the shared batch to avoid destination confusion for directed sends.

I treated that as a pragmatic local correction to behaviour that was already not ideal. It is not intended to be the final or ideal P2P batching design, because that is not what this PR is about.

That said, I have already reworked `_z_transport_tx_dest_t` once in response to review feedback, including changing it to use `z_variant_template`. I do not think it is reasonable to keep reopening this part of the PR at this stage, especially around separate work that I was not aware of when implementing this.

I don't have time to make further structural changes before my holiday. If the team prefers to remove or adapt these TX changes to fit around #1265, or make other changes while I am away, I am happy for that to happen.

PR #1259, opened 2026-06-30
Improve fuzz coverage workflow and add transport round-trip encoding

evshary:
## Description

This PR improves the fuzzing workflow by documenting how to collect and inspect LLVM coverage, and by extending the transport fuzz target to exercise both decode and encode paths.

### What does this PR do?

- Documents how to inspect coverage for each fuzz target, including focused reporting for the most relevant files.
- Extends `fuzz/fuzz_transport_message_decode.c` to round-trip valid decoded transport messages through encode and decode again, increasing coverage of the transport codec encode path.

The existing transport fuzz target only exercised decoding, so the encode side of the transport codec was not being covered.
At the same time, the fuzzing documentation did not clearly explain how to build with coverage instrumentation or how to inspect the resulting reports.

The coverage improvement:

```
Filename                                                     Regions    Missed Regions     Cover   Functions  Missed Functions  Executed       Lines      Missed Lines     Cover    Branches   Missed Branches     Cover
src/protocol/codec/transport.c                                   582               367    36.94%          23                11    52.17%         451               217    51.88%         236               138    41.53%
src/protocol/codec/transport.c                                   582               148    74.57%          23                 2    91.30%         451                56    87.58%         236                48    79.66%
```

### Related Issues

---
## 🏷️ Label-Based Checklist

**Labels:** `internal`

## 🏠 Internal Change

**Instructions:**

### DIFF
--- a/fuzz/fuzz_transport_message_decode.c
+++ b/fuzz/fuzz_transport_message_decode.c
@@ -25,7 +25,24 @@ int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
     _z_zbuf_t zbf = _z_slice_as_zbuf(&slice);
     _z_transport_message_t msg = {0};
 
-    (void)_z_transport_message_decode(&msg, &zbf);
+    if (_z_transport_message_decode(&msg, &zbf) == _Z_RES_OK) {
+        // Re-encode valid decoded messages to exercise the transport encode path.
+        _z_wbuf_t wbf = _z_wbuf_null();
+
+        if (_z_wbuf_init(&wbf, size + 64, true) == _Z_RES_OK) {
+            if (_z_transport_message_encode(&wbf, &msg) == _Z_RES_OK) {
+                _z_zbuf_t encoded = _z_wbuf_to_zbuf(&wbf);
+                _z_transport_message_t roundtrip = {0};
+
+                (void)_z_transport_message_decode(&roundtrip, &encoded);
+                _z_zbuf_clear(&encoded);
+            }
+
+            _z_wbuf_clear(&wbf);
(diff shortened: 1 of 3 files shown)


fuzz/fuzz_transport_message_decode.c
```diff
@@ -25,7 +25,24 @@ int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
     _z_zbuf_t zbf = _z_slice_as_zbuf(&slice);
     _z_transport_message_t msg = {0};
 
-    (void)_z_transport_message_decode(&msg, &zbf);
+    if (_z_transport_message_decode(&msg, &zbf) == _Z_RES_OK) {
+        // Re-encode valid decoded messages to exercise the transport encode path.
+        _z_wbuf_t wbf = _z_wbuf_null();
+
+        if (_z_wbuf_init(&wbf, size + 64, true) == _Z_RES_OK) {
+            if (_z_transport_message_encode(&wbf, &msg) == _Z_RES_OK) {
```
DenisBiryukov91, line 33: What is the point of exercising encoding in the test, if we allow it to fail ?

evshary: In fact, this is also to cover the encoding path and improve the test coverage. We can find some crashes in the encoding.
We can report errors if our goal is to make every decodable message encodable. However, it might be too strict. I did some experiments, and I found that there are several inconsistent issues for encoding and decoding. I would recommend focusing on catching the vulnerabilities first. As for the protocol inconsistency, this can be the next step.

DenisBiryukov91 (approved):

PR #1261, opened 2026-07-01
Do not run ubuntu integration tests without ASAN

DenisBiryukov91:
## Description

Only run ubuntu integration tests with ASAN (since those without ASAN are identical, but do not catch memory errors)

### What does this PR do?

Only run ubuntu integration tests with ASAN (since those without ASAN are identical, but do not catch memory errors)

Reduce CI time

### Related Issues

---
## 🏷️ Label-Based Checklist

**Labels:** `internal`

## 🏠 Internal Change

**Instructions:**

### DIFF
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -84,9 +84,9 @@ jobs:
         uses: actions/cache@v4
         with:
           path: ~/.ccache
-          key: ${{ runner.os }}-ccache-integration-${{ hashFiles('CMakeLists.txt', 'GNUmakefile', 'cmake/**', 'src/**', 'include/**', 'tests/**') }}
+          key: ${{ runner.os }}-asan${{ matrix.asan }}-ccache-integration-${{ hashFiles('CMakeLists.txt', 'GNUmakefile', 'cmake/**', 'src/**', 'include/**', 'tests/**') }}
           restore-keys: |
-            ${{ runner.os }}-ccache-integration-
+            ${{ runner.os }}-asan${{ matrix.asan }}-ccache-integration-
 
       - name: Compile debug
         run: make all
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -22,12 +22,18 @@ env:
 
 jobs:
   build:
-    name: Build on ${{ matrix.os }}
+    name: Build on ${{ matrix.name }}
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macOS-latest]
+        include:
+        - name: ubuntu-latest with ASAN
+          os: ubuntu-latest
+          asan: 1
+        - name: macOS-latest
+          os: macOS-latest
+          asan: 0
 
     steps:
       - name: Clone this repository
@@ -100,6 +106,7 @@ jobs:
           Z_FEATURE_ADVANCED_PUBLICATION: 1
           Z_FEATURE_ADVANCED_SUBSCRIPTION: 1
           Z_FEATURE_ADMIN_SPACE: 1
+          ASAN: ${{ matrix.asan }}
 
       - name: Download and extract Zenoh
         if: runner.os == 'Linux'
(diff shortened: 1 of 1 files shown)


gmartin82 (changes requested):
I do not think the non-ASAN Ubuntu integration run is redundant.

ASAN is a materially different executable configuration: it adds `-fsanitize=address` at compile and link time, changing allocator behaviour, memory layout, binary instrumentation, and execution timing. It is valuable for finding memory errors, but it is not a substitute for exercising the normal Linux binary. A race, timeout, scheduling-sensitive, layout-sensitive, or optimisation-sensitive issue can pass under ASAN and fail without it, or vice versa.

There is also already a separate coverage gap here: `GNUmakefile` defaults to Release, while these integration jobs explicitly build Debug. Release uses `-O3`, whereas Debug uses `-O0`, so neither the old nor proposed integration matrix exercises the default optimised configuration.

Rather than dropping the ordinary Ubuntu job, I think we should repurpose it as:

* Ubuntu, Release, ASAN=OFF
* Ubuntu, Debug, ASAN=ON
* macOS, Debug, ASAN=OFF

That keeps the current job count while covering the default Linux behaviour and retaining ASAN diagnostics.

There is also a cache-key issue here: the unified job now uses the same ccache key as the non-ASAN integration build, rather than retaining a distinct ASAN-specific key.

DenisBiryukov91:
> I do not think the non-ASAN Ubuntu integration run is redundant.
> 
> ASAN is a materially different executable configuration: it adds `-fsanitize=address` at compile and link time, changing allocator behaviour, memory layout, binary instrumentation, and execution timing. It is valuable for finding memory errors, but it is not a substitute for exercising the normal Linux binary. A race, timeout, scheduling-sensitive, layout-sensitive, or optimisation-sensitive issue can pass under ASAN and fail without it, or vice versa.
> 
> There is also already a separate coverage gap here: `GNUmakefile` defaults to Release, while these integration jobs explicitly build Debug. Release uses `-O3`, whereas Debug uses `-O0`, so neither the old nor proposed integration matrix exercises the default optimised configuration.
> 
> Rather than dropping the ordinary Ubuntu job, I think we should repurpose it as:
> 
> * Ubuntu, Release, ASAN=OFF
> * Ubuntu, Debug, ASAN=ON
> * macOS, Debug, ASAN=OFF
> 
> That keeps the current job count while covering the default Linux behaviour and retaining ASAN diagnostics.
> 
> There is also a cache-key issue here: the unified job now uses the same ccache key as the non-ASAN integration build, rather than retaining a distinct ASAN-specific key.

@gmartin82 

As far as I understand all behavior changes that you mentioned for non-ASAN run, such as allocator behavior, memory layout optimizations and execution time scheduling do not have much to do with ASAN, which does not change it to a greater extent than enabling/disabling certain feature flags or adding an extra compile option or simply changing machine configuration / compiler version.
Additionally release mode tends to only occasionally expose code issues - and in this case we just consider the test to be flaky and simply rerun it until it pass - **which brings the value of this test to 0**.

Also tests failed in release mode on GitHub runner tend to produce very little value for us since in most case due to the reasons cited above - they are sporadic and often non-reproducible locally, nor provide any useful information on the origin of the error.

In addition we already run all the tests in release mode on all 3 OSes both with and without unstable API through cpp wrapper.

We also run zenoh-pico examples in release mode as a part of the CI.

I think that running tests with TSAN would produce way more value in detecting scheduling issues or data races. 

Also, I'm not sure I understood why macOS does not deserve a run in release mode ?

gmartin82:
I do not think the conclusion follows.

The current integration workflow is already testing Debug builds. Debug and instrumented builds are materially different configurations: they change code generation, optimisation, execution overhead and timing; ASAN also changes memory-access and allocator behaviour. In my two decades working on ZettaScale codebases, Release testing has repeatedly exposed issues which were not present in Debug or instrumented test runs.

Additional configurations, such as TSAN or further macOS coverage, may well be useful. However, that is separate from the point here: this PR removes a supposedly redundant configuration, but a normal non-instrumented Linux build is not redundant with an ASAN Debug build.

If we are changing the matrix, the useful change would be to make the non-ASAN Ubuntu integration run a Release build, while retaining the ASAN Debug run.

DenisBiryukov91:
> I do not think the conclusion follows.
> 
> The current integration workflow is already testing Debug builds. Debug and instrumented builds are materially different configurations: they change code generation, optimisation, execution overhead and timing; ASAN also changes memory-access and allocator behaviour. In my two decades working on ZettaScale codebases, Release testing has repeatedly exposed issues which were not present in Debug or instrumented test runs.
> 
> Additional configurations, such as TSAN or further macOS coverage, may well be useful. However, that is separate from the point here: this PR removes a supposedly redundant configuration, but a normal non-instrumented Linux build is not redundant with an ASAN Debug build.
> 
> If we are changing the matrix, the useful change would be to make the non-ASAN Ubuntu integration run a Release build, while retaining the ASAN Debug run.

@gmartin82 
Any change in configuration/compiler/os can potentially detect errors otherwise non-detectable without it - yet we do not exercises dozens of differently configured tests to catch these errors.

And I'm not sure I've ever seen the case in zenoh-pico code base where Debug would repeatedly detect something what Debug with ASAN would not - which is sufficient for me to consider Debug - only tests redundant.

Why not simply flip all integration tests (including ASAN one) to RELEASE with debug info then ?

gmartin82:
I think that, if we are going to change the build matrix, it needs careful consideration to make sure it is a genuine improvement rather than simply exchanging one form of coverage for another.

My concern is that, historically, testing Release versions of our products has surfaced issues that were not found in Debug or instrumented builds. I also think there is value in testing the product in the form in which it is built and supplied to end users.

So I do not think we should assume that a non-instrumented configuration is redundant without considering what coverage would be lost.

If comprehensive CI coverage is too expensive to run on every change, I do not think the answer should be to treat distinct configurations as redundant and remove them. We should instead consider which higher-cost or lower-probability configurations can run less frequently—for example on a schedule, before release, or when relevant files change—while making that coverage and its cadence explicit.

DenisBiryukov91:
> I think that, if we are going to change the build matrix, it needs careful consideration to make sure it is a genuine improvement rather than simply exchanging one form of coverage for another.
> 
> My concern is that, historically, testing Release versions of our products has surfaced issues that were not found in Debug or instrumented builds. I also think there is value in testing the product in the form in which it is built and supplied to end users.
> 
> So I do not think we should assume that a non-instrumented configuration is redundant without considering what coverage would be lost.
> 
> If comprehensive CI coverage is too expensive to run on every change, I do not think the answer should be to treat distinct configurations as redundant and remove them. We should instead consider which higher-cost or lower-probability configurations can run less frequently—for example on a schedule, before release, or when relevant files change—while making that coverage and its cadence explicit.

@gmartin82 
I still do not get what your concern regarding Release versions has to do with the subject of this PR: namely removal of clearly redundant test in Debug version ? You already explicitly suggested in your very first comment, that Debug version with ASAN is the only Debug version on ubuntu that should stay - this is exactly what this PR does.

If you think that we are missing certain configurations in our test coverage, you can always open a separate PR to address this issue.

PR #1264, opened 2026-07-06
Add hashset support + iteration optimization

DenisBiryukov91:
## Description

Add hashset support.
Merge static and dynamic hashmap template code into a single header file.
Optimize iteration over large-size maps containing small number of elements by introducing chaining of live nodes (this change does not increase storage requirements for most cases, but significantly improves iteration speed for large size maps, including static ones, with small number of elements)

### What does this PR do?

Add hashset support.
Merge static and dynamic hashmap template code into a single header file.
Optimize iteration over large-size maps containing small number of elements by introducing chaining of live nodes (this change does not increase storage requirements for most cases, but significantly improves iteration speed for large size maps, including static ones, with small number of elements)

Allow using hash sets (i.e. hash maps without value type).
Reduce code duplication.
Optimize iteration for large maps with small number of elements (this mostly concerns static hashmap uses).

### Related Issues

---
## 🏷️ Label-Based Checklist

**Labels:** `internal`

## 🏠 Internal Change

**Instructions:**


include/zenoh-pico/collections/hashmap_template.h
```diff
@@ -287,22 +325,27 @@ static inline bool _ZP_CAT(_ZP_HASHMAP_TEMPLATE_NAME, grow)(_ZP_HASHMAP_TEMPLATE
...
-            _ZP_HASHMAP_TEMPLATE_KEY_MOVE_FN(&new_slots[i]._node.key, &map->_slots[i]._node.key);
+    // MOVE_FN, and keep live links
+    for (size_t i = 0; i < map->_capacity; i++) {
+        new_slots[i]._next_live = map->_slots[i]._next_live;
+        new_slots[i]._prev_live = map->_slots[i]._prev_live;
+        if (map->_slots[i]._next_live != _ZP_HASHMAP_TEMPLATE_INDEX_NONE ||
+            map->_slots[i]._prev_live != _ZP_HASHMAP_TEMPLATE_INDEX_NONE) {
+            _ZP_HASHMAP_TEMPLATE_KEY_MOVE_FN(_ZP_HASHMAP_TEMPLATE_NODE_KEY(&new_slots[i]._node),
+                                             _ZP_HASHMAP_TEMPLATE_NODE_KEY(&map->_slots[i]._node));
+#ifndef _ZP_HASHMAP_TEMPLATE_IS_SET
```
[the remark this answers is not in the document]
DenisBiryukov91: fixed

include/zenoh-pico/collections/hashmap_template.h
```diff
@@ -193,19 +227,21 @@ typedef _ZP_HASHMAP_TEMPLATE_ITER_TYPE _ZP_HASHMAP_TEMPLATE_ITER_TYPEDEF;
...
-// of the free list, again chained through _next. The _bucket field is always
-// the head of bucket i, independent of whether slot i itself holds a live entry.
+//   _node      : key/value payload of node i
+//   _next      : index of the next node in the bucket chain, or the next free slot if _bucket == INDEX_NONE
+//   _bucket    : index of the first node in bucket i, INDEX_NONE = empty
+//   _next_live : an index of the live node preceeding this one in the iteration order, or INDEX_NONE if this is the
+//   first live node.
+//   _prev_live : an index of the live node following this one in the iteration order, or INDEX_NONE if
+//   this is the last live node.
+//   If both _prev_live and _next_live are INDEX_NONE, then this node is not live (not present in the map).
```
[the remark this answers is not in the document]
DenisBiryukov91: fixed

include/zenoh-pico/collections/hashmap_template.h
```diff
@@ -329,8 +373,10 @@ static inline bool _ZP_CAT(_ZP_HASHMAP_TEMPLATE_NAME, grow)(_ZP_HASHMAP_TEMPLATE
     }
     // Re-hash every live entry into the new bucket heads (indices are preserved).
     for (size_t i = 0; i < map->_capacity; i++) {
-        if (new_slots[i]._present) {
-            size_t b = _ZP_HASHMAP_TEMPLATE_KEY_HASH_FN(&new_slots[i]._node.key) % new_capacity;
+        if (new_slots[i]._next_live != _ZP_HASHMAP_TEMPLATE_INDEX_NONE ||
+            new_slots[i]._prev_live != _ZP_HASHMAP_TEMPLATE_INDEX_NONE) {
+            size_t b =
```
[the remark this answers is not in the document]
DenisBiryukov91: fixed

include/zenoh-pico/collections/static_hashmap_template.h
```diff
@@ -148,16 +173,21 @@ typedef _ZP_STATIC_HASHMAP_TEMPLATE_ITER_TYPE _ZP_STATIC_HASHMAP_TEMPLATE_ITER_T
...
-//   _bucket   : index of the first node in bucket i, INDEX_NONE = empty
-//   _present  : whether node i currently holds a live entry
+//   _node      : key/value payload of node i
+//   _next      : index of the next node in the bucket chain, or the next free slot if _bucket == INDEX_NONE
+//   _bucket    : index of the first node in bucket i, INDEX_NONE = empty
+//   _next_live : an index of the live node preceeding this one in the iteration order, or INDEX_NONE if this is the
+//   first live node.
+//   _prev_live : an index of the live node following this one in the iteration order, or INDEX_NONE if
+//   this is the last live node.
+//   If both _prev_live and _next_live are INDEX_NONE, then this node is not live (not present in the map).
```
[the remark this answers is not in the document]
DenisBiryukov91: fixed

include/zenoh-pico/collections/hashmap_template.h
```diff
@@ -624,53 +695,32 @@ static inline bool _ZP_CAT(_ZP_HASHMAP_TEMPLATE_NAME, remove)(_ZP_HASHMAP_TEMPLA
...
-        if (map->_slots[i]._present) {
-            _ZP_HASHMAP_TEMPLATE_NODE_TYPE *n = &map->_slots[i]._node;
-            _ZP_HASHMAP_TEMPLATE_KEY_DESTROY_FN(&n->key);
-            _ZP_HASHMAP_TEMPLATE_VAL_DESTROY_FN(&n->val);
-        }
-    }
+    _ZP_CAT(_ZP_HASHMAP_TEMPLATE_NAME, clear)(map);
     _ZP_HASHMAP_TEMPLATE_FREE_FN(map->_slots);
     map->_slots = NULL;
     map->_capacity = 0;
```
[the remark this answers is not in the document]
DenisBiryukov91: fixed

include/zenoh-pico/collections/hashmap_template.h
```diff
@@ -624,53 +701,33 @@ static inline bool _ZP_CAT(_ZP_HASHMAP_TEMPLATE_NAME, remove)(_ZP_HASHMAP_TEMPLA
...
-    for (size_t i = 0; i < map->_capacity; i++) {
-        if (map->_slots[i]._present) {
-            _ZP_HASHMAP_TEMPLATE_NODE_TYPE *n = &map->_slots[i]._node;
-            _ZP_HASHMAP_TEMPLATE_KEY_DESTROY_FN(&n->key);
-            _ZP_HASHMAP_TEMPLATE_VAL_DESTROY_FN(&n->val);
-        }
-    }
+    _ZP_CAT(_ZP_HASHMAP_TEMPLATE_NAME, clear)(map);
     _ZP_HASHMAP_TEMPLATE_FREE_FN(map->_slots);
     map->_slots = NULL;
```
evshary: Do we need to initialize _size and _live_head here? I know we've cleared it in the clear function. Maybe it's worth doing it explicitly here?

include/zenoh-pico/collections/hashmap_template.h
```diff
@@ -91,9 +98,10 @@
 #error "_ZP_HASHMAP_TEMPLATE_KEY_TYPE must be defined before including hashmap_template.h"
 #define _ZP_HASHMAP_TEMPLATE_KEY_TYPE int
```
evshary: I guess we don't need to define _ZP_HASHMAP_TEMPLATE_KEY_TYPE after `#error`

evshary (commented):
I'm not sure whether this is a good idea to have hashset inside `hashmap_template.h`. It's not intuitive to me to find hashset in hashmap. Maybe having another general name (`hashmap_hashset_template.h`) is a good idea.
On the other side, I agree that if we rename the file, it will be harder to trace the changes, so I don't have strong preference on this.

gmartin82:
## Codex review comments for consideration

The current head fixes the earlier singleton-entry relocation and destroyed-map reset problems, and CI is green.

I would still request changes for these points:

1. **Do not bundle a global hashmap representation/iteration rewrite into “add hashset collection”.**

   Both templates replace the existing `_present` flag with `_live_head`, `_next_live`, and `_prev_live`. That adds two index fields to **every slot of every existing hashmap**, increasing static and dynamic memory use even where no hash set is used. It also rewrites iteration and destruction mechanics across the board.

   The hash-set feature does not inherently require this. Retaining `_present` and the existing scan-based iteration would keep the feature much more contained. At minimum, this should be split into a separate, benchmarked optimisation PR with a stated memory trade-off.

2. **This changes observable iteration and destruction order for existing maps.**

   Previously `begin()` / `iter_next()` scanned slots in ascending index order. The new implementation inserts every allocated slot at `_live_head`, then iterates via `_next_live`; a freshly inserted sequence is therefore visited in reverse allocation order. `clear()` and `destroy()` now use that same order.

   Even if ordering was never promised as a formal API contract, it is externally observable through iteration callbacks and custom destructors. That directly conflicts with the checked “No behavior changes” checklist.

   Please preserve the existing iteration order, or make this a deliberate documented behavioural change rather than incidental fallout from adding hash sets.

3. **Hash-set selection should be explicit, not inferred from an omitted required macro.**

   `_ZP_*_HASHMAP_TEMPLATE_VAL_TYPE` used to be mandatory. Now an omitted value type silently produces a materially different collection: different element type, different `insert` and `remove` signatures, and missing `get` APIs.

   That makes a configuration error produce a different API rather than a clear compile-time diagnostic. I would prefer an explicit opt-in, such as a dedicated `hashset_template.h` wrapper or a clearly named `*_TEMPLATE_IS_SET` macro, while retaining the error when neither a value type nor set mode is specified.

4. **Add a genuine map-mode custom-move growth test.**

   The new set test covers a custom **key** mover, but the map test called `strmap` only provides a value destructor; it leaves the default value mover in place. Consequently, the template marks the slot trivially movable and takes the memcpy/realloc path, despite the test claiming to exercise the per-entry move path.

   Please add a map with a custom `VAL_MOVE_FN` that clears the source, force `reserve()` or growth on a partially filled map, and verify retained entries plus clean destruction. That specifically covers the new conditional map-only move logic.

Minor: the `_next_live` / `_prev_live` comments are backwards. `_next_live` is followed by `iter_next()`, so it is absent for the last item, not the first.

The hash-set API itself is reasonable, but this PR currently combines it with an unrelated rewrite of hashmap occupancy and iteration. That increases the footprint of every existing map and changes iteration/destruction order, despite the PR declaring no behavioural changes. Please keep the set feature narrowly scoped, use an explicit set opt-in rather than treating a missing value type as a different container, and add coverage for the map-mode custom-move growth path.

gmartin82:
Other than what Codex found I would agree with @evshary that it may not be straight forward for developers to see that there is a hashset implementation if they don't know it's there.

DenisBiryukov91:
> Other than what Codex found I would agree with @evshary that it may not be straight forward for developers to see that there is a hashset implementation if they don't know it's there.

Debatable, but this the same approach as one that was taken for a bit vector/bit set - so this would need to be changed too.
Otherwise it allows direct code reuse, given that hashmap/hashset are not really the different implementation-wise.
One possible approach is to have internal_hashmap_template.h in a separate file and just stub hashmap/hashset headers - the disadvantage is that it would require providing almost identical doc strings/ readme description - which would hinder maintaining it up to date (since we will now need to keep in sync 2 almost identical blobs of text) - @gmartin82  let me know if you really prefer this approach.

Also, next time you do a codex review, please provide a single (or better an inline) comment per codex finding, otherwise it is a bit difficult to reply to them.

gmartin82:
> Also, next time you do a codex review, please provide a single (or better an inline) comment per codex finding, otherwise it is a bit difficult to reply to them.

If Codex access is enabled on the repository then I will let it post in the format it posts reviews in. Otherwise when I choose to request an AI review from a tool I have access to I will post it's feedback on the PR as is.

Given that you are a user of AI tools for review purposes yourself then I expect you won't have a problem with this?

DenisBiryukov91:
> I'm not sure whether this is a good idea to have hashset inside `hashmap_template.h`. It's not intuitive to me to find hashset in hashmap. Maybe having another general name (`hashmap_hashset_template.h`) is a good idea. On the other side, I agree that if we rename the file, it will be harder to trace the changes, so I don't have strong preference on this.

@evshary Yes, it might make sense. The issue is given that both have almost identical methods - this will lead to having 2 almost identical doc-string blobs in headers and 2 even bigger text blobs in readme.md. And in the end these 2 headers would just include the current hashmap one, renamed into something like internal_hashmap_template.h - due to very similar hashmap/hashset implementations. So the question do we really want this, given that we keep template collections only for internal usage ?

include/zenoh-pico/collections/hashmap_template.h
DenisBiryukov91: No, clear is enough, it resets map to zero-size with live_head pointing to INDEX_NONE.

include/zenoh-pico/collections/hashmap_template.h
DenisBiryukov91: fixed

DenisBiryukov91:
> ## Codex review comments for consideration
> The current head fixes the earlier singleton-entry relocation and destroyed-map reset problems, and CI is green.
> 
> I would still request changes for these points:
> 
> 1. **Do not bundle a global hashmap representation/iteration rewrite into “add hashset collection”.**
>    Both templates replace the existing `_present` flag with `_live_head`, `_next_live`, and `_prev_live`. That adds two index fields to **every slot of every existing hashmap**, increasing static and dynamic memory use even where no hash set is used. It also rewrites iteration and destruction mechanics across the board.
>    The hash-set feature does not inherently require this. Retaining `_present` and the existing scan-based iteration would keep the feature much more contained. At minimum, this should be split into a separate, benchmarked optimisation PR with a stated memory trade-off.

Updated PR name/description to reflect this change. 
For most use cases this change does not provide any storage increase given that struct size is **rounded up to the multiple of word size** : 
- word size of 8. With default uint32_t index, we go from (2 * uint32_t + bool) to (4 * uint32_t) which both are equal to 2 words. With index uint8_t / uint16_t index for both cases we end up with single word-size node metadata (2 * uint8_t + bool) to (4 * uint8_t) / (2 * uint16_t + bool) to (4 * uint16_t). 
- word size 4 - due to similar reasons the only increase due extra metadata is for uint32_t index (which normally should not be used in zenoh-pico, since we do not expect any map to contain more than 65534 elements, especially on 32-bit controllers). 

> 2. **This changes observable iteration and destruction order for existing maps.**
>    Previously `begin()` / `iter_next()` scanned slots in ascending index order. The new implementation inserts every allocated slot at `_live_head`, then iterates via `_next_live`; a freshly inserted sequence is therefore visited in reverse allocation order. `clear()` and `destroy()` now use that same order.
>    Even if ordering was never promised as a formal API contract, it is externally observable through iteration callbacks and custom destructors. That directly conflicts with the checked “No behavior changes” checklist.
>    Please preserve the existing iteration order, or make this a deliberate documented behavioural change rather than incidental fallout from adding hash sets.

Hash map iteration order is an internal detail, and thus this is not relevant.

> 3. **Hash-set selection should be explicit, not inferred from an omitted required macro.**
>    `_ZP_*_HASHMAP_TEMPLATE_VAL_TYPE` used to be mandatory. Now an omitted value type silently produces a materially different collection: different element type, different `insert` and `remove` signatures, and missing `get` APIs.

Made dedicated header for hash sets.

>    That makes a configuration error produce a different API rather than a clear compile-time diagnostic. I would prefer an explicit opt-in, such as a dedicated `hashset_template.h` wrapper or a clearly named `*_TEMPLATE_IS_SET` macro, while retaining the error when neither a value type nor set mode is specified.
> 4. **Add a genuine map-mode custom-move growth test.**
>    The new set test covers a custom **key** mover, but the map test called `strmap` only provides a value destructor; it leaves the default value mover in place. Consequently, the template marks the slot trivially movable and takes the memcpy/realloc path, despite the test claiming to exercise the per-entry move path.
>    Please add a map with a custom `VAL_MOVE_FN` that clears the source, force `reserve()` or growth on a partially filled map, and verify retained entries plus clean destruction. That specifically covers the new conditional map-only move logic.
> 
Done (although this is not really related to the PR).

> Minor: the `_next_live` / `_prev_live` comments are backwards. `_next_live` is followed by `iter_next()`, so it is absent for the last item, not the first.
> 
Fixed.

> The hash-set API itself is reasonable, but this PR currently combines it with an unrelated rewrite of hashmap occupancy and iteration. That increases the footprint of every existing map and changes iteration/destruction order, despite the PR declaring no behavioural changes. Please keep the set feature narrowly scoped, use an explicit set opt-in rather than treating a missing value type as a different container, and add coverage for the map-mode custom-move growth path.

evshary (approved):
I don't have a strong opinion on this since this is internal usage.
We can rename the header to `hashmap_hashset_template.h` and `static_hashmap_hashset_template.h` to make it explicit.
Otherwise, it's also okay for me to keep the current implementation (as far as we all know this).

DenisBiryukov91:
> I don't have a strong opinion on this since this is internal usage. We can rename the header to `hashmap_hashset_template.h` and `static_hashmap_hashset_template.h` to make it explicit. Otherwise, it's also okay for me to keep the current implementation (as far as we all know this).

Made dedicated headers for hash sets. Also merged static/dynamic hash maps (and hash sets) code into a single internal template header.

include/zenoh-pico/collections/internal/hashmap_template_internal.h
```diff
@@ -0,0 +1,853 @@
...
+    }
+    for (_ZP_HASHMAP_TEMPLATE_ITER_TYPE i = 0; i + 1 < _ZP_HASHMAP_TEMPLATE_CAPACITY; i++) {
+        map->_slots[i]._next = (_ZP_HASHMAP_TEMPLATE_ITER_TYPE)(i + 1);
+    }
+    map->_slots[_ZP_HASHMAP_TEMPLATE_CAPACITY - 1]._next = _ZP_HASHMAP_TEMPLATE_INDEX_NONE;  // end of free list
+#else
+    map->_slots = NULL;
+    map->_capacity = 0;
+#endif
+    map->_free_head = 0;
```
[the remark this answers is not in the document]
DenisBiryukov91: Fixed

include/zenoh-pico/collections/internal/hashmap_template_internal.h
```diff
@@ -0,0 +1,853 @@
...
+#undef _ZP_HASHMAP_TEMPLATE_IS_STATIC
+#else
+#undef _ZP_HASHMAP_TEMPLATE_INITIAL_CAPACITY
+#undef _ZP_HASHMAP_TEMPLATE_INDEX_TYPE
+#undef _ZP_HASHMAP_TEMPLATE_ALLOC_FN
+#undef _ZP_HASHMAP_TEMPLATE_FREE_FN
+#ifdef _ZP_HASHMAP_TEMPLATE_REALLOC_FN
+#undef _ZP_HASHMAP_TEMPLATE_REALLOC_FN
+#endif
+#ifdef _ZP_HASHMAP_TEMPLATE_KEY_TRIVIALLY_MOVABLE
```
steils, line 844: These three undefs (838-846) are inside the #else branch (dynamic mode only, line 830), but the macros themselves are defined unconditionally at lines 128, 133, 167, 171 - in both modes. So any include of static_hashmap_template.h/static_hashset_template.h leaves `_*_KEY/VAL/SLOT_TRIVIALLY_MOVABLE` defined.
After static_hashmap_template.h or static_hashset_template.h, a later dynamic map/set with custom move callbacks can be misclassified as trivially movable and grow via memcpy/realloc instead of the configured move function.

include/zenoh-pico/collections/internal/hashmap_template_internal.h
```diff
@@ -0,0 +1,853 @@
...
+#endif
+#ifdef _ZP_HASHMAP_TEMPLATE_VAL_TRIVIALLY_MOVABLE
+#undef _ZP_HASHMAP_TEMPLATE_VAL_TRIVIALLY_MOVABLE
+#endif
+#ifdef _ZP_HASHMAP_TEMPLATE_SLOT_TRIVIALLY_MOVABLE
+#undef _ZP_HASHMAP_TEMPLATE_SLOT_TRIVIALLY_MOVABLE
+#endif
+#undef _ZP_HASHMAP_TEMPLATE_MAX_CAPACITY
+#undef _ZP_HASHMAP_TEMPLATE_MAX_ALLOC_SIZE
+#undef _ZP_HASHMAP_TEMPLATE_NODE_IS_LIVE
```
steils: It undefines `_ZP_HASHMAP_TEMPLATE_NODE_IS_LIVE`, but the macro is named `_ZP_HASHMAP_NODE_IS_LIVE` (if they are not different macros, but then the latter is never undefined).

tests/z_static_hashset_template_test.c
```diff
@@ -0,0 +1,459 @@
+//
```
steils, line 1: No test combines a static and a dynamic template in the same .c file (which is exactly what triggers the mentioned macro leak).

include/zenoh-pico/collections/README.md
steils: I think this section should be updated to include heap/static hashsets. New hashset tests use _ZP_FOREACH, _ZP_CONST_FIND, and _ZP_REMOVE, please add them to the supported list.

steils (changes requested):

include/zenoh-pico/collections/internal/hashmap_template_internal.h
DenisBiryukov91: fixed

include/zenoh-pico/collections/README.md
DenisBiryukov91: fixed

include/zenoh-pico/collections/internal/hashmap_template_internal.h
DenisBiryukov91: fixed

include/zenoh-pico/collections/README.md
steils: hash maps, hash sets and both vectors

steils: *hash maps, hash sets and both vectors

steils (approved):

include/zenoh-pico/collections/README.md
DenisBiryukov91: Fixed

DenisBiryukov91: Fixed

PR #1272, opened 2026-07-26
fix: parse standard UUID text representation

### LINKED ISSUE #1229: [Bug] UUID bytes from string doesn't follow the standard UUID textual representation
### Describe the bug

The [UUID bytes from string function](https://github.com/eclipse-zenoh/zenoh-pico/blob/07c84ebcf926114bffbf70edd82f3d71919c3868/src/utils/uuid.c#L26), used in different places, e.g. when reading the config to open a session, doesn't parse correctly the standard UUID text representation.

In particular, it checks for the last dash 2 indexes too early
- RFC 9562:             `XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX` (8-4-4-4-12)
- zenoh pico uuid: `XXXXXXXX-XXXX-XXXX-XX-XXXXXXXXXXXXXX` (8-4-4-2-14)

This prevents compatibility with other UUID libraries, and in general I can't find a valid reason to use another format different from the standard (in case there is, sorry!)

> [!NOTE]
> I can't find the exact section of the RFC that talks about textual representation, but [here is the Wikipedia one](https://en.wikipedia.org/wiki/Universally_unique_identifier#Textual_representation), and also here is an [example from the RFC](https://datatracker.ietf.org/doc/html/rfc9562#name-uuid-format)

### To reproduce

1. Try to set in the config a standard 8-4-4-12 uuid as the ID
2. You will see extra zeros because of the missed characters

### System info

- Platform: ESP32C3, but shouldn't matter
- Zenoh pico version 1.9.0

tianrking:
### What does this PR do?

- Correct the final dash position used when parsing the standard UUID text representation (`8-4-4-4-12`).
- Add a regression test that converts `00112233-4455-6677-8899-aabbccddeeff` to the expected 16 bytes.

The parser skipped the last dash two characters too early, so standard UUID strings were decoded with incorrect bytes. This restores compatibility with RFC-style UUID strings used by other libraries and configuration producers.

### Validation

- Built the `z_utils_test` target with CMake/Ninja and MSVC.
- Ran `z_utils_test.exe` successfully.
- `ctest -R '^z_utils_test$' --output-on-failure` — 1/1 passed.

### Related Issues

Fixes #1229.

### Development disclosure

OpenAI Codex assisted with the implementation and test development. The patch was independently cross-reviewed and validated with MSVC as well as a GCC/AddressSanitizer test run.

---
## 🏷️ Label-Based Checklist

**Labels:** `bug`

## 🐛 Bug Fix Requirements

**Instructions:**

### DIFF
--- a/src/utils/uuid.c
+++ b/src/utils/uuid.c
@@ -26,7 +26,7 @@
 void _z_uuid_to_bytes(uint8_t *bytes, const char *uuid_str) {
     uint8_t n_dash = 0;
     for (uint8_t i = 0; i < 32; i += 2) {
-        if (i == 8 || i == 12 || i == 16 || i == 18) {
+        if (i == 8 || i == 12 || i == 16 || i == 20) {
             n_dash += 1;
         }
         char val[5] = {'0', 'x', uuid_str[i + n_dash], uuid_str[i + 1 + n_dash], '\0'};
--- a/tests/z_utils_test.c
+++ b/tests/z_utils_test.c
@@ -20,9 +20,21 @@
 #include "zenoh-pico/utils/pointers.h"
 #include "zenoh-pico/utils/query_params.h"
 #include "zenoh-pico/utils/time_range.h"
+#include "zenoh-pico/utils/uuid.h"
 
 #undef NDEBUG
 #include <assert.h>
+
+static void test_uuid_to_bytes(void) {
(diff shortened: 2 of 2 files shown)


DenisBiryukov91:
Hello @tianrking . Thanks for your contribution.  Could you please sign eclipse eca https://www.eclipse.org/legal/eca/?

tianrking:
done

PR #1278, opened 2026-07-28
fix(threadx): fix clock scaling math and TCB memory leak in system.c

### LINKED ISSUE #1233: [Bug] ThreadX __z_clock_gettime has broken time logic
### Describe the bug

in src/app/tx_user.h there is:
    #define TX_TIMER_TICKS_PER_SECOND                2000         
then there is this in stm32/system.c
```
    void __z_clock_gettime(z_clock_t *ts) {                                                                                                                                          
        uint64_t ms = tx_time_get() * (TX_TIMER_TICKS_PER_SECOND / 1000); // <--ms now is 2x too fast                                                                                          
        ts->tv_sec = ms / (uint64_t)1000;                             // <-- sec is consequentially wrong, so is ns also                                                                                                              
        ts->tv_nsec = (ms % (uint64_t)1000) * (uint64_t)1000;                                                                                                                        
    } 
```
Should be like this
```

void __z_clock_gettime(z_clock_t *ts) {
    uint64_t ticks = tx_time_get();
    ts->tv_sec = ticks / TX_TIMER_TICKS_PER_SECOND;
    ts->tv_nsec = (ticks % TX_TIMER_TICKS_PER_SECOND) * 1000000000ULL / TX_TIMER_TICKS_PER_SECOND;
}

```

### To reproduce

1. Build zenoh-pico app on STM32 with ThreadX and serial.
2. Connect to a zenohd router over UART

### System info

• Platform: STM32G474QE, ThreadX 6.x (Azure RTOS), serial transport over UART
• CPU: ARM Cortex-M4
• zenoh-pico version: https://github.com/eclipse-zenoh/zenoh-pico/commit/47022da1cee2570910adb17643cab429f3563ee8

### LINKED ISSUE #1273: [Bug] ThreadX PAL _z_task_join does not delete completed thread, causing tx_thread_create failure on reconnect
### Affected Component
`src/system/threadx/stm32/system.c`

### Summary
`_z_task_join()` polls until thread state is `TX_COMPLETED`, but fails to call `tx_thread_delete()`. ThreadX retains the thread control block in its active system list, causing subsequent `tx_thread_create()` calls on session reconnect to fail with error `0x0E` (`TX_THREAD_ERROR`).

### Observed Log / Firmware Symptom
When the serial link drops and reconnects, PyOCD / SWD trace shows:
`_z_task_init` returns `-1` (`_Z_ERR_GENERIC`) because `tx_thread_create` returned `0x0E`. Reconnection loop hangs in `picoros_init()`.

### Root Cause
ThreadX requires `tx_thread_delete()` to be explicitly called on completed or terminated threads before their `TX_THREAD` control block memory can be reused for a new thread.

### Proposed Fix
```c
z_result_t _z_task_join(_z_task_t *task) {
    while (1) {
        UINT state;
        UINT status = tx_thread_info_get(&(task->threadx_thread), NULL, &state, NULL, NULL, NULL, NULL, NULL, NULL);
        if (status != TX_SUCCESS) _Z_ERROR_RETURN(_Z_ERR_GENERIC);

        if ((state == TX_COMPLETED) || (state == TX_TERMINATED)) break;

        tx_thread_sleep(1);
    }

    tx_thread_delete(&(task->threadx_thread));
    return _Z_RES_OK;
}
```

### Why it fixes it
Calling `tx_thread_delete()` unlinks the completed thread from ThreadX kernel data structures, allowing `tx_thread_create()` to successfully re-initialize the thread control block upon session reconnect.

michael545:
### What does this PR do?
This PR resolves two critical ThreadX PAL issues in `src/system/threadx/stm32/system.c`:

1. **Clock Math Integer Truncation**: `__z_clock_gettime()` and `_z_get_time_since_epoch()` evaluated `(TX_TIMER_TICKS_PER_SECOND / 1000)` first. For 2000Hz tick rates, `2000 / 1000 = 2`, accelerating reported time by 4x. For <=500Hz tick rates, `500 / 1000 = 0`, freezing time at 0. This PR updates `__z_clock_gettime()` to calculate seconds by `ticks / TX_TIMER_TICKS_PER_SECOND` and nanoseconds via remainder modulus.
2. **Task Control Block Memory Leak**: `_z_task_join()` waited for task completion (`TX_COMPLETED`), but omitted `tx_thread_delete()`. On session reconnection, `_z_task_init()` invoked `tx_thread_create()` on the existing TCB, failing with `0x0E` (`TX_THREAD_ERROR`). This PR adds `tx_thread_delete(&(task->threadx_thread))` in `_z_task_join()`.

### Related Issues
- Fixes #1233
- Fixes #1273

Signed-off-by: michael545 <michael.valand@gmail.com>

---
## 🏷️ Label-Based Checklist

**Labels:** `bug`

## 🐛 Bug Fix Requirements

**Instructions:**

### DIFF
--- a/src/system/threadx/stm32/system.c
+++ b/src/system/threadx/stm32/system.c
@@ -83,6 +83,7 @@ z_result_t _z_task_join(_z_task_t *task) {
         tx_thread_sleep(1);
     }
 
+    tx_thread_delete(&(task->threadx_thread));
     return _Z_RES_OK;
 }
 
@@ -222,7 +223,7 @@ z_result_t _z_condvar_wait_until(_z_condvar_t *cv, _z_mutex_t *m, const z_clock_
     }
 
     ULONG now = tx_time_get();
-    ULONG target_time = (abstime->tv_sec * 1000 + abstime->tv_nsec / 1000000) * (TX_TIMER_TICKS_PER_SECOND / 1000);
+    ULONG target_time = ((abstime->tv_sec * 1000ULL + abstime->tv_nsec / 1000000ULL) * TX_TIMER_TICKS_PER_SECOND) / 1000ULL;
     ULONG block_duration = (target_time > now) ? (target_time - now) : 0;
 
     tx_mutex_get(&cv->mutex, TX_WAIT_FOREVER);
@@ -265,9 +266,9 @@ z_result_t z_sleep_s(size_t time) {
 /*------------------ Clock ------------------*/
 
 void __z_clock_gettime(z_clock_t *ts) {
-    uint64_t ms = tx_time_get() * (TX_TIMER_TICKS_PER_SECOND / 1000);
-    ts->tv_sec = ms / (uint64_t)1000;
-    ts->tv_nsec = (ms % (uint64_t)1000) * (uint64_t)1000;
+    uint64_t ticks = (uint64_t)tx_time_get();
+    ts->tv_sec = ticks / TX_TIMER_TICKS_PER_SECOND;
+    ts->tv_nsec = (ticks % TX_TIMER_TICKS_PER_SECOND) * 1000000000ULL / TX_TIMER_TICKS_PER_SECOND;
 }
 
 z_clock_t z_clock_now(void) {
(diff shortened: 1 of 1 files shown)


src/system/threadx/stm32/system.c
```diff
@@ -73,16 +73,8 @@ z_result_t _z_task_init(_z_task_t *task, z_task_attr_t *attr, void *(*fun)(void
...
-        UINT state;
-        UINT status = tx_thread_info_get(&(task->threadx_thread), NULL, &state, NULL, NULL, NULL, NULL, NULL, NULL);
-        if (status != TX_SUCCESS) _Z_ERROR_RETURN(_Z_ERR_GENERIC);
-
-        if ((state == TX_COMPLETED) || (state == TX_TERMINATED)) break;
-
-        tx_thread_sleep(1);
-    }
-
+    tx_thread_terminate(&(task->threadx_thread));
```
DenisBiryukov91: As far as I understand tx_thread_terminate forcefully stops a thread, which is different from join semantics which should merely wait until thread runs to completion. This looks to be more suitable for z_task_cancel

michael545: Yes, most definitely, it was just a quick-and-drity hack, totally violated join! Now it gracefully waits and always **deletes** after completion which was the cause of the bug in hte first place

PR #1284, opened 2026-08-05
Continue scouting after empty Hello messages

### LINKED ISSUE #1283: Continue scouting when Zenoh sends an empty Hello
## Summary

Zenoh-pico stops scouting after the first decoded Hello message, even when that Hello contains no locators. This causes default session opening to fail with `_Z_ERR_SCOUT_NO_RESULTS` when used with Zenoh commit `3ed025b7c373fea81e169f67832e0b606ca57396` and later `main` revisions.

## Affected Versions

- Zenoh-pico: `9fb469aa072007e7f9b410e25d06057518892c30`
- Zenoh before regression: `ad5e077a80841f4fa7f6cb7eb7b13fa509a17784`
- Zenoh regression commit: `3ed025b7c373fea81e169f67832e0b606ca57396`
- Also reproduced with Zenoh `db859f420737f82518fe8158172c13087d3b51f9`

## Reproduction

Run Zenoh with only a loopback TCP listener:

```bash
zenohd -l tcp/127.0.0.1:7447
```

Open a Zenoh-pico session with the default configuration:

```c
z_config_default(&config);
z_open(z_loan(config), NULL);
```

The session uses multicast scouting because no explicit connect locator is configured.

## Observed Behavior

Zenoh-pico fails with:

```text
Failed to open session(Error code: -87 )
```

`-87` is `_Z_ERR_SCOUT_NO_RESULTS`.

The issue is reproducible in the zenoh-cpp integration tests, including:

```text
source_info_zenohpico
pub_sub_zenohpico
queryable_get_zenohpico
cancellation_zenohpico
```

The same tests pass with Zenoh `ad5e077a`.

## Root Cause

Zenoh commit `3ed025b7` introduced loopback-aware locator filtering. For a non-loopback Scout source, Zenoh now uses `get_locators_noloopback()`. If the router only has loopback locators available, it can send a valid Hello with an empty locator list.

Zenoh-pico currently exits on the first Hello in `src/session/scout.c`:

```c
if (!_z_hello_slist_is_empty(ret) && exit_on_first) {
    break;
}
```

It does not check whether the Hello contains locators. `_z_locators_by_scout()` then copies the empty locator list, and session opening returns `_Z_ERR_SCOUT_NO_RESULTS`.

Rust Zenoh already ignores empty Hello messages during automatic connection discovery and continues scouting.

## Expected Behavior

When opening a session, Zenoh-pico should continue scouting after a Hello with no locators and stop only after receiving a Hello containing at least one usable locator or after the scouting timeout expires.

## Proposed Fix

Change the early-exit condition in `src/session/scout.c` to require a non-empty locator list:

```c
if (!_z_hello_slist_is_empty(ret) && exit_on_first &&
    (_z_string_svec_len(&_z_hello_slist_value(ret)->_locators) > 0)) {
    break;
}
```

This preserves empty Hello messages for general scouting APIs while preventing automatic session opening from terminating prematurely.

## Validation

With this local Zenoh-pico change and unpatched Zenoh `db859f42`:

```text
15/15 Zenoh-pico integration tests passed
```

The affected source-info, pub/sub, queryable/get, liveliness, cancellation, and connectivity tests all pass.

## CI Coverage

The standard Zenoh-pico CI does not enable `BUILD_INTEGRATION`, which defaults to `OFF`. Therefore, tests such as `z_api_source_info_test` are not run. The modular CI uses a released Zenoh standalone binary rather than the moving Zenoh `main` revision, so this cross-repository compatibility case is not covered.

diogomatsubara:
## Summary

When automatic session scouting receives a Hello message with no locators, continue scouting instead of treating that Hello as the first usable result.

This affects interoperability with Zenoh `3ed025b7` and later revisions, which can send an empty Hello after filtering loopback locators for a non-loopback Scout source.

Fixes eclipse-zenoh/zenoh-pico#1283

## Changes

- Add `_z_scout_should_exit_on_first()` to centralize the usable-Hello condition.
- Keep the early-exit behavior only when the Hello contains at least one locator.
- Add `z_scout_test` covering empty and non-empty Hello messages.

## Validation

- `cmake -S . -B build -DCMAKE_C_STANDARD=11 -DBUILD_TESTING=ON -DBUILD_INTEGRATION=OFF ...`
- `cmake --build build --target z_scout_test z_open_test`
- `ctest --test-dir build --output-on-failure -R '^z_scout_test$|^z_open_test$'`
- Full local unit suite excluding unrelated macOS packaging tests: `39/39` passed.

## Notes

An empty Hello does not itself provide enough information to establish a transport. This change ensures that pico does not stop before a later Hello with usable locators can be processed; it still returns `_Z_ERR_SCOUT_NO_RESULTS` when the scouting timeout expires without receiving usable locators.

---
## 🏷️ Label-Based Checklist

**Labels:** `bug`

## 🐛 Bug Fix Requirements

**Instructions:**


diogomatsubara:
zenoh-cpp still fails because the CI checks out the current zenoh-cpp which still has the zenoh-pico submodule pointing to an older version (9fb469aa072007e7f9b410e25d06057518892c30)

fuzzypixelz (approved):
Zenoh (Rust) was always susceptible to producing `Hello` messages  with empty `locators` (in version up until 1.9), in particular when only the loopback interface is available on the host. So it's safer for Pico to ignore such messages. Ideally, they should not be sent at all but that would be merit discussion about the scouting protocol.

PR #1288, opened 2026-08-12
freertos: configure TCP send timeout

### LINKED ISSUE #1267: [Bug] Auto-reconnect can be broken due to FreeRTOS network.c missing send timeout
### Describe the bug

`_z_socket_set_blocking()` configures a receive timeout, but no send timeout:
`BaseType_t result = FreeRTOS_setsockopt(sock->_socket, 0, FREERTOS_SO_RCVTIMEO, &option_value, sizeof(option_value));`

If the router is ungracefully disconnected (via a power blip or some other condition which causes no RST/FIN to be sent upon disconnect), z_publisher_put()) will block indefinitely waiting for the TCP send buffer to drain, permanently blocking the lease task. 

### To reproduce

The condition can be simulated by the following steps:
1. Setup: zenoh router and zenoh-pico device running FreeRTOS (TCP, client mode, unicast, Z_FEATURE_AUTO_RECONNECT enabled)
2. Establish the connection
3. Silence the router via iptables
```
sudo iptables -A INPUT -s 192.168.0.100 -j DROP
sudo iptables -A OUTPUT -d 192.168.0.100 -j DROP
```
4. After 30 seconds, reenable traffic:
```
sudo iptables -D INPUT -s 192.168.0.100 -j DROP
sudo iptables -D OUTPUT -d 192.168.0.100 -j DROP
```
5. observe device doesn't auto-reconnect

### System info

- FreeRTOS Plus TCP
- zenoh-pico 1.7.2

Daniiiil1:
## Description

Configure both receive and send timeouts when changing the FreeRTOS+TCP socket blocking mode.

### What does this PR do?

- Applies `Z_CONFIG_SOCKET_TIMEOUT` to `FREERTOS_SO_SNDTIMEO` as well as `FREERTOS_SO_RCVTIMEO`.
- Checks `FreeRTOS_setsockopt()` against its documented success return value (`0`). The pinned FreeRTOS+TCP implementation follows the Berkeley convention and returns `0` on success, while the previous `pdPASS` comparison treated a successful call as an error.

If a router disappears without sending RST/FIN, the TCP send buffer can fill and leave a publisher blocked indefinitely. The lease task then cannot make progress and auto-reconnect never gets a chance to recover the session. A send timeout bounds that wait and lets the existing error/reconnect path run.

### Related Issues

Fixes #1267

### Validation

- `clang-format -n -Werror src/system/freertos/freertos_plus_tcp/network.c`
- `git diff --check`
- Confirmed against the pinned FreeRTOS+TCP source (`34148c3`) that `FREERTOS_SO_RCVTIMEO` and `FREERTOS_SO_SNDTIMEO` return `0` on success.
- The official FreeRTOS+TCP example configured successfully on macOS. Its build requires Linux's `bits/pthread_stack_min.h`, so the full Debug/Release build is left to the repository's Ubuntu CI.

---
## 🏷️ Label-Based Checklist

Current labels: _No labels_

### DIFF
--- a/src/system/freertos/freertos_plus_tcp/network.c
+++ b/src/system/freertos/freertos_plus_tcp/network.c
@@ -28,7 +28,12 @@ z_result_t _z_socket_set_blocking(const _z_sys_net_socket_t *sock, bool blocking
     BaseType_t result =
         FreeRTOS_setsockopt(sock->_socket, 0, FREERTOS_SO_RCVTIMEO, &option_value, sizeof(option_value));
 
-    if (result != pdPASS) {
+    if (result != 0) {
+        _Z_ERROR_RETURN(_Z_ERR_GENERIC);
+    }
+
+    result = FreeRTOS_setsockopt(sock->_socket, 0, FREERTOS_SO_SNDTIMEO, &option_value, sizeof(option_value));
+    if (result != 0) {
         _Z_ERROR_RETURN(_Z_ERR_GENERIC);
     }
     return _Z_RES_OK;


DenisBiryukov91:
@Daniiiil1 Hello, is it still work in progress, or do you consider it finished ?

PR #1298, opened 2026-08-27
feat(link): add CAN / CAN FD multicast and ISO-TP unicast links

jerry73204:
### What does this PR do?

Adds two CAN links, both off by default, mirroring the pair proposed for
zenoh in eclipse-zenoh/zenoh#2757.

**`Z_FEATURE_LINK_CAN`** -- `can`, a **multicast** link over CAN / CAN FD. A CAN
frame names a message rather than a destination and every node hears every
frame, which is what a zenoh multicast transport already expects. MTU is one
frame; zenoh fragments on top of it and the link adds no second segmentation
layer. Where a payload does not land on a DLC step the frame is padded up to
the next one and the length is recovered from zenoh's own headers.

**`Z_FEATURE_LINK_ISOTP`** -- `isotp`, a **unicast** link over ISO 15765-2. The
address is a directed identifier pair plus flow control, so exactly one peer
owns the other end. Segmentation moves below zenoh, giving a 4095-byte MTU.

CAN is the bus these targets are already wired to. Reaching a zenoh network
from an MCU on that bus should not require a gateway.

Both links exist because the multicast one is not sufficient: zenoh routes
queries and liveliness to unicast faces only, so over a multicast-only link a
peer can publish and subscribe and nothing more. ISO-TP supplies a real unicast
face on a bus with no addresses, and everything built on a query then works.

### Design notes worth a reviewer's attention

**ISO-TP is never implemented here -- it is a platform capability.** Linux has
it in the kernel (`CAN_ISOTP`), Zephyr in `subsys/canbus/isotp`, and
`third_party/isotp-c` (MIT, no allocator, ~1.2k lines) serves platforms with
neither.

**The vendored copy comes from a fork, and I want to flag that explicitly.**
It is `SimonCahill/isotp-c` at `abb9e55` plus **one commit** adding an
`ISO_TP_NO_FORMATTED_ERRORS` build flag -- 18 added lines, nothing else.
`snprintf` is that library's only libc dependency beyond `memcpy`/`memset`,
and it exists for two formatted diagnostics that each write into a 128-byte
stack buffer, which makes it a hard requirement of the whole library on
precisely the targets this link is for. The flag omits both; the errors still
go through `isotp_user_debug()` and the return codes are unchanged. It is off
by default, so the tree builds identically to upstream unless a port asks.
On Cortex-M4 with `-Os -DNDEBUG -ffreestanding` it takes `.text` from 2235 to
2091 bytes and the largest stack frame from 160 to 32, after which the object
needs nothing from libc but `memcpy` and `memset`.

The fork is staging, not a destination -- I would rather this pointed at
`SimonCahill/isotp-c` directly, and `VENDOR.md` says so. If you would prefer
the PR wait until the flag lands upstream, or prefer the flag dropped and the
copy taken verbatim from `abb9e55`, say which and I will do that. On `unix` **both** backends are implemented, selected by
`Z_FEATURE_LINK_ISOTP_VENDORED`: the kernel is the oracle the portable library
is judged against, on the one platform where both exist. `third_party/isotp-c/`
carries a `VENDOR.md` recording provenance, licence, the survey behind the
choice, and the update procedure.

**Flow control is configurable, and this is the end that needs it.** The
endpoint takes `stmin` and `bs`, the two knobs ISO 15765-2 gives a receiver:
how long the peer must leave between the ConsecutiveFrames it sends us, and how
many it may send before waiting for another flow control frame. They travel in
the flow control *this* peer transmits, so they constrain the other side. A
zenoh-pico node is usually the small end of the link, and these are its only
defence against a Linux peer that would otherwise push 4 KiB as fast as the
controller takes it.

Measured on `vcan0`, zenoh-rs publishing to a pico subscriber: frames arrive
0.05 ms apart by default and 5.06 ms apart at `stmin=5`, and `bs=4` turns 29
flow control frames into 351.

The backends differ, and one cannot do it. The kernel socket takes
`CAN_ISOTP_RECV_FC` before bind and Zephyr takes an `isotp_fc_opts` into
`isotp_bind`. The vendored `isotp-c` reads its flow control from compile-time
constants with no per-link field, so it **refuses** an endpoint that sets
either rather than accepting one and pacing nothing. A peer that asked to be
protected from a burst should find out at open, not when it drops frames on a
real bus.

**The ISO-TP link registers in `_z_open_link` only, deliberately.** A peer that
dials out never reaches `_zp_unicast_accept_task`, which wants a socket and an
`accept()` that no bus has.

**Additive.** One line of pre-existing code changes: a capability guard that
gains `|| Z_FEATURE_LINK_CAN == 1`. Everything else is new files, new enum
arms, and build plumbing.

The PR is **two commits, one per link**, so the CAN link can be taken and
ISO-TP deferred. The first commit adds no ISO-TP files at all.

### Platforms

`unix` and `zephyr`. The Zephyr module gains `CONFIG_ZENOH_PICO_LINK_CAN` and
`CONFIG_ZENOH_PICO_LINK_ISOTP` (the latter depends on `CAN && ISOTP`). Note
Zephyr marks its ISO-TP subsystem `[EXPERIMENTAL]`, and its own test suite
skips the `STmin` timing cases -- worth knowing before a constrained node
depends on services over this link.

### Footprint

31 files, +4326/-1.

### Testing

`tests/z_can_link_test.c` covers every DLC boundary, both directions, and an
over-MTU write being refused.

Built and exercised in all five feature combinations -- default, `can`,
`isotp` on the kernel socket, `isotp` on the vendored library, and both links
together.

Verified against zenoh-rs over `vcan0`: session establishment, pub/sub, and a
query that returns a reply, with FirstFrame / FlowControl / ConsecutiveFrame on
the wire as ISO 15765-2 specifies. Also run on Zephyr (`native_sim`) talking to
a Linux peer over a shared CAN bus.

**Not covered:** ISO-TP timing parameters (`STmin`, `BS`) -- `vcan` has no bit
rate.


jerry73204:
### CI triage

77 checks pass. The remaining failures are pre-existing on `main`, not
introduced here -- comparing against the last `main` CI run
([33098050489 -> 32268465376](https://github.com/eclipse-zenoh/zenoh-pico/actions/runs/32268465376)):

| Check | On this PR | On `main` |
|---|---|---|
| `Run integration tests` (ubuntu / macOS / ASAN) | fail | **fail** |
| `Codacy Security Scan with cppcheck` | fail | **fail** |
| `CI status checks` (aggregate of the above) | fail | **fail** |

The integration-test failures are all the same error: `z_open(...) returned -87`
(`_Z_ERR_SCOUT_NO_RESULTS`) -- multicast scouting finding nothing on the runner.
`main` fails tests **10--18** this way; this branch fails **11--18**, one fewer.
Both links are off by default, so none of the code added here is even compiled
in that configuration.

Two remaining items:

- **`Generate / main`** wants one of the required labels. I do not have
  permission to add one -- could a maintainer apply `new feature`?
- **Flawfinder** reports 30 alerts in the new code: `memcpy`, `char` buffers,
  `strlen`, `getenv`, `strncpy`. These are the same patterns the existing
  sources use (`src/link/multicast/udp.c` trips the same rules); Codacy only
  reports alerts in changed code, so existing ones are not re-listed. The one
  worth a second look is the `strncpy` in `src/system/unix/isotp_vendored.c`:
  the `struct ifreq` is `memset` to zero first and the copy is bounded to
  `sizeof(ifr.ifr_name) - 1`, so NUL termination is guaranteed. Happy to
  restructure any of them if you would rather they were silenced.

`Check codebase format with clang-format` passes.

Issue #1290, opened 2026-08-17
[Bug]  _z_wbuf_reset() heap-buffer-overflow and dropped io-slices when pruning borrowed (zero-copy) buffers


olegk-sd:
### Describe the bug

`_z_wbuf_reset()` corrupts adjacent heap memory (confirmed under ASAN) and
can silently skip elements when it prunes "borrowed" io-slices added by
`_z_wbuf_wrap_bytes()` (zero-copy sends). Two distinct bugs combine to
produce the same crash signature:

**Bug A — double-multiplication overflow in `_z_svec_remove()`**
(`src/collections/vec.c:256`):

```c
void _z_svec_remove(_z_svec_t *v, size_t pos, z_element_clear_f clear, z_element_move_f move, size_t element_size,
                    bool use_elem_f) {
    assert(pos < v->_len);
    clear((uint8_t *)v->_val + pos * element_size);
    __z_svec_move_inner((uint8_t *)v->_val + pos * element_size, (uint8_t *)v->_val + (pos + 1) * element_size, move,
                        (v->_len - pos - 1) * element_size, element_size, use_elem_f);   // <-- bug
    v->_len--;
}
```

`__z_svec_move_inner`'s `num_elements` parameter (`src/collections/vec.c:146`)
is an **element count** — it multiplies by `element_size` internally on the
`memmove` path:

```c
memmove(dst, src, num_elements * element_size);
```

`_z_svec_remove` passes `(v->_len - pos - 1) * element_size` as
`num_elements`, so the actual `memmove` size becomes
`(v->_len - pos - 1) * element_size^2` — an overflow by a factor of
`element_size` whenever a *non-last* element is removed from a
multi-byte-element vector.

Confirmed under ASAN:
```
ERROR: AddressSanitizer: heap-buffer-overflow, READ of size 1600 out of a 160-byte region
  #1 __z_svec_move_inner   vec.c:146   (memmove)
  #2 _z_svec_remove        vec.c:260
  #4 _z_wbuf_reset         iobuf.c:523
```

**Bug B — loop skips an element after removal in `_z_wbuf_reset()`**
(`src/protocol/iobuf.c:515`):

```c
void _z_wbuf_reset(_z_wbuf_t *wbf) {
    ...
    for (size_t i = 0; i < _z_iosli_svec_len(&wbf->_ioss); i++) {
        _z_iosli_t *ios = _z_wbuf_get_iosli(wbf, i);
[...]

PR #1302, opened 2026-08-31
Fix heap buffer overflow in _z_svec_remove

hyunuk-zz:
### What does this PR do?

Fixes the element count passed by `_z_svec_remove()` to `__z_svec_move_inner()` and adds a regression test.

`__z_svec_move_inner()` expects a number of elements and multiplies it by `element_size` internally.

`_z_svec_remove()` was passing an already byte-scaled value, causing an oversized `memmove()` when removing a non-last element.

### Testing

Added `svec_remove_non_last_element_test()` in `z_collections_test.c`.

The test verifies that:
- the remaining elements are shifted correctly;
- memory beyond the logical vector range is not modified.

The test fails with the previous implementation and passes with this fix.

### Related Issues

Addresses Bug A in #1290.

---
## 🏷️ Label-Based Checklist

**Labels:** `bug`

## 🐛 Bug Fix Requirements

**Instructions:**

### DIFF
--- a/src/collections/vec.c
+++ b/src/collections/vec.c
@@ -258,7 +258,7 @@ void _z_svec_remove(_z_svec_t *v, size_t pos, z_element_clear_f clear, z_element
     assert(pos < v->_len);
     clear((uint8_t *)v->_val + pos * element_size);
     __z_svec_move_inner((uint8_t *)v->_val + pos * element_size, (uint8_t *)v->_val + (pos + 1) * element_size, move,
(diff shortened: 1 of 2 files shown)


DenisBiryukov91:
@hyunuk-zz Could you please fix formatting ?

hyunuk-zz:
> @hyunuk-zz Could you please fix formatting ?

Fixed the formatting. Thanks!

PR #1313, opened 2026-09-08
fix(transport): report peer send errors without skipping healthy peers

### LINKED ISSUE #1312: [Bug] Error Code is lost in Peer Mode
### Describe the bug

While debugging with Zenoh-Pico in Peer mode, I found that some error messages are missing because they are lost. 
Looking into the source code, for example at https://github.com/eclipse-zenoh/zenoh-pico/blob/main/src/transport/common/tx.c#L139, I found that this is because of a missing error message.

If my Zenoh node is not a peer, no error message is written out.

### To reproduce

I discovered this when I tried to limit the network connection.

### System info

- Windows 11
- zenoh-pico 1.9.0

vpfkfl753:
### What does this PR do?

Return peer write failures from the transport send path while continuing delivery to the other peers. Frames and fragments use the same helper, and healthy peers still receive all fragments when another peer fails.

Only mark the transport as having transmitted data when a write succeeds. Clear a peer batch after attempting delivery, including partial failure, so a subsequent flush does not duplicate data at the peers that already received it.

The peer loops ignore `_z_link_send_wbuf()` results in both the frame and fragment paths. They return success even when every peer write fails, hiding errors from callers and the existing keep-alive error handling.

The regression injects failures at the first, middle, last and all peers, alongside a successful control. It checks transport messages, frames, fragments and batches, compares the healthy peers' bytes with the control, and checks subsequent sends and batch flushes. All four failing paths reproduce on the original code.

On Linux, the regression passes with the default features, fragmentation disabled, and multithreading/batching disabled. The iobuf, message-codec and fragment-reliability tests also pass. A separate socket-pair check through the POSIX write path reproduces the error on the original code and passes with the fix.

### Related Issues

Fixes #1312

---
## 🏷️ Label-Based Checklist

Current labels: _No labels_


vpfkfl753:
Could someone add the `bug` label? The checklist job requires it, and I don't have permission to apply labels.

PR #1314, opened 2026-09-09
fix(emscripten): usleep() instead of emscripten_sleep() for pthreads builds

Tobias-Fischer:
## Update: superseded by a more thorough non-blocking rewrite

The approach below (a real, blocking `usleep()` for pthreads builds) was
itself superseded by a project pivot away from pthreads entirely — real
pthreads requires WebAssembly `--shared-memory`, which deadlocks
completely once loaded into a non-pthreads host (confirmed: a
pthreads-enabled `rclpy` build loaded into JupyterLite's `xeus-python`
kernel never responds again once it blocks).

What actually shipped needs neither pthreads nor Asyncify: every
genuinely blocking call in zenoh-pico's own Emscripten platform layer
(`z_open_ws`/`_z_read_ws`/`_z_send_ws` in `network.c`, and the
client-side Init/Open handshake in `transport/unicast/transport.c`) was
rewritten to make exactly one non-blocking attempt per call and return
immediately — no `usleep()`, no `emscripten_sleep()`, nothing that
blocks or sleeps at all — with the *caller* responsible for retrying
across separate top-level calls, each with a real yield back to the
browser's event loop in between (a JS-driven tick loop, or Python's own
`await asyncio.sleep()`, driven through
esol-community/rmw_zenoh_pico#7's own non-blocking `session_connect()`
rewrite).

This PR's branch (based on an older zenoh-pico tag than the fork
currently pins) isn't updated to that approach here, since I haven't
verified the newer patches still apply cleanly against this base without
testing them against it directly — happy to open a proper follow-up PR
with the full non-blocking rewrite (six small, focused patches: a
`ws/` locator validity-check fix, the non-blocking I/O rewrite, a
resumable client handshake state machine, single-threaded-mode
mutex/condvar stubs, and a missing `<stddef.h>` include) if there's
interest, rather than leaving this one open against a now-superseded
approach. Full current source for all of it:
https://github.com/Tobias-Fischer/ros-rolling/tree/pico-update-tmp/extra_recipes/zenoh-pico

## Summary
- `z_sleep_ms`/`z_sleep_us` in the Emscripten platform layer call `emscripten_sleep()`, which requires the module to be built with Asyncify (or JSPI).
- A `pthreads`-enabled build (`-pthread -sUSE_PTHREADS=1`) has a real OS thread available, where a plain blocking `usleep()` works correctly and is both simpler and cheaper than pulling in Asyncify.
- Without this fix, any pthreads build calling `z_sleep_ms` (e.g. `_z_open_ws()`'s post-connect wait for the WebSocket to actually open, in `ws_emscripten.c`) crashes at runtime with:
  ```
  Please compile your program with async support in order to use asynchronous operations like emscripten_sleep
  ```
- Falls back to the existing `emscripten_sleep()` behavior when `__EMSCRIPTEN_PTHREADS__` isn't defined, so non-pthreads (Asyncify) builds are unaffected.

Found and fixed while getting a `rclc`/`rmw_zenoh_pico`-based ROS 2 node running as a real pthreads WebAssembly build in the browser, talking to a native `zenohd` router over WebSocket.

## Test plan
- [x] Confirmed the crash reproduces without this fix in a real browser (Emscripten 4.0.9, `-pthread -sUSE_PTHREADS=1 -sPROXY_TO_PTHREAD=1`), and disappears with it — a `zenoh-pico` WebSocket session now connects to a native router end-to-end.
- [ ] Not run through zenoh-pico's own CI/test suite locally (no CMake configure environment set up for this fork) — please let CI confirm the standard emscripten build targets still pass.

## Full write-up

All the changes this required, across every repo, are documented together in [Tobias-Fischer/ros2-emscripten-zenoh-demo](https://github.com/Tobias-Fischer/ros2-emscripten-zenoh-demo) — including a working `rclc` *and* `rclpy` browser demo verified end-to-end against a native `zenohd` router.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
## 🏷️ Label-Based Checklist

Current labels: _No labels_

### DIFF
--- a/src/system/emscripten/system.c
+++ b/src/system/emscripten/system.c
@@ -18,6 +18,7 @@
 #include <stddef.h>
 #include <stdio.h>
 #include <stdlib.h>
+#include <unistd.h>
 
 #include "zenoh-pico/config.h"
 #include "zenoh-pico/system/common/system_error.h"
@@ -128,13 +129,25 @@ z_result_t _z_condvar_wait_until(_z_condvar_t *cv, _z_mutex_t *m, const z_clock_
 #endif  // Z_FEATURE_MULTI_THREAD == 1
 
 /*------------------ Sleep ------------------*/
+// emscripten_sleep() requires the module to be built with Asyncify (or JSPI),
+// which is incompatible with a pthreads build. A pthreads build has a real OS
+// thread, so a plain blocking usleep() works and is preferred there; keep
+// emscripten_sleep() as the fallback for non-pthreads (Asyncify) builds.
 z_result_t z_sleep_us(size_t time) {
+#if defined(__EMSCRIPTEN_PTHREADS__)
+    usleep((useconds_t)time);
+#else
     emscripten_sleep((time / 1000) + (time % 1000 == 0 ? 0 : 1));
+#endif
     return 0;
 }
 
 z_result_t z_sleep_ms(size_t time) {
+#if defined(__EMSCRIPTEN_PTHREADS__)
+    usleep((useconds_t)time * 1000);
+#else
     emscripten_sleep(time);
+#endif
     return 0;
 }
 


hachi-kun:
This is an attempt at using it in a way that we hadn't considered before.

Can I see a sample code for your suggestion for rmw_zenoh_pico ?

rmw_zenoh_pico is a prototype implementation planned as a ZeonH implementation that can replace microoxrdds for microcontrollers.
For this reason, the most important thing is stable operation on the MCU.
Therefore, I understand that it has a different approach than the original ROS2.

If your proposal doesn't affect the MCU environment (zepher/freertos/baremetal,,), I think it will be merged.

> CMakeLists.txt / package.xml: depend on ament_cmake + ament_cmake_ros_core instead of the full
> CMakeLists.txt: drop-in ament_target_dependencies() shim, guarded by if(NOT COMMAND

Will there be any problems if I don't change this ament_xxx part?

ex.
> if(rosidl_typesupport_microxrcedds_cpp_FOUND)
>  ament_target_dependencies(${PROJECT_NAME}
>    "rosidl_typesupport_microxrcedds_cpp"
>)
>  target_compile_definitions(${PROJECT_NAME}
>    PUBLIC
>      HAVE_CPP_TYPESUPPORT
>)
>endif()

Will there be any problems if I don't change this ament_xxx part?

The "HAVE_CPP_TYPESUPPORT" setting in this section is not actually used anywhere. Therefore, it's better to remove it rather than modify it.
Also, it seems that the current RCL package configuration for microros does not support rosidl_typesupport in rclcpp.
While this functionality might be necessary in the future if a microros-compatible rclcpp is released, it's not needed at this time.
Perhaps the original author didn't have sufficient knowledge of microros at the time of creating this package.
I've heard that this issue is expected to be fixed in the next release.

> rmw_init.c: use a "ws/" locator instead of "tcp/" for the connect address when built for Emscripten

I think it's possible if the zenoh-pico side you're using supports it.
Which version of zenoh are you expecting?

I've heard that for the time being, rmw_zenoh_pico will use zenohpico/1.8 to match the version of zenoh used by the rmw_zenoh it connects to.

> zenoh_pico_session.c (session_connect()): z_open()'s own session/link-open handshake can genuinely need several attempts to complete

I agree with adding a retry function to z_open().

However, since rmw_zenoh_pico is based on the assumption that it operates in client mode, we need to compare it with the rmw_zenoh implementation to see if it's okay if z_open() fails. The connection with zenohd is a secondary connection.

> rmw_wait.c: under Z_FEATURE_MULTI_THREAD==0

Your point that support for single-tasking environments is insufficient is correct.
However, proving the necessity of directly porting microro to a bare-metal environment where multitasking is not possible is difficult.
In that case, it might be better to consider alternatives such as a subset of the ros-API that is compatible with rosAPI instead of microros.

Therefore, I think rmw_zenoh_pico will only support MutilThread for the time being. I feel memory efficiency will probably be the top priority.

> including working rclc, rclcpp, and rclpy browser demos, and an editable JupyterLite notebook, all verified end-to-end against a native zenohd router.

Did you verify this using rcl + rmw_zenoh_pico + (rcl/rclcpp/rclpy)?
At that time, were you using the rcl source code with the microros fixes added?
If you include the microros fixes in rcl, wouldn't some of the functions that directly manipulate topic names, like in rosbug, stop working correctly?

